PR #1, opened 2021-01-08
Bootstrap operator with kubebuilder

Danil-Grigorev:
Working on initial project structure.

Creates a structure for the future work on `ClusterOperator` resource, and some of the manifests to apply from CVO.


Fedosin (commented):
Add OWNERS please

Makefile
```diff
@@ -0,0 +1,100 @@
...
+	$(KUSTOMIZE) build config/crd | kubectl delete -f -
+
+# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: manifests kustomize
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default | kubectl apply -f -
+
+# UnDeploy controller from the configured Kubernetes cluster in ~/.kube/config
+undeploy:
+	$(KUSTOMIZE) build config/default | kubectl delete -f -
```
JoelSpeed, line 46: We won't need any of this

Makefile
```diff
@@ -0,0 +1,100 @@
...
+# Download controller-gen locally if necessary
+CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
+controller-gen:
+	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1)
+
+# Download kustomize locally if necessary
+KUSTOMIZE = $(shell pwd)/bin/kustomize
+kustomize:
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
+
```
JoelSpeed, line 87: Don't need this either

cmd/cluster-cloud-controller-manager-operator/main.go
```diff
@@ -0,0 +1,104 @@
...
+
+func main() {
+	var metricsAddr string
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
```
JoelSpeed, line 59: There are a few standard flags we normally add here, don't need to add them now but make a note to set them up later. For example, Namespace, Leader election configuration

cmd/cluster-cloud-controller-manager-operator/main.go
```diff
@@ -0,0 +1,104 @@
...
+	opts := zap.Options{
+		Development: true,
+	}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
+
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
```
JoelSpeed, line 66: I'd suggest swtiching to klogr instead of zap, it's what we use in our other projects

controllers/clusteroperator_controller.go
```diff
@@ -0,0 +1,69 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	// Lock in some nessesary libraries for operator functionality
+	_ "github.com/openshift/api/operator/v1"
+	_ "github.com/openshift/library-go/pkg/operator/events"
+	_ "github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	_ "github.com/openshift/library-go/pkg/operator/resource/resourcehash"
+	_ "github.com/openshift/library-go/pkg/operator/resource/resourcemerge"
```
JoelSpeed, line 33: These aren't being used? Should we drop them until they are being used? What's the benefit of doing this? What if they never get used?

controllers/clusteroperator_controller.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the ClusterOperator object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0/pkg/reconcile
```
JoelSpeed, line 55: I'd suggest replacing this with a more appropriate comment

controllers/suite_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Controller Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = BeforeSuite(func() {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
```
JoelSpeed, line 51: If we switch to klogr, do the same here too

controllers/suite_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	By("bootstrapping test environment")
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
```
JoelSpeed, line 55: We are going to need to import the cluster operator CRD, can we load that from the vendor directory? How do we do this elsewhere?

manifests/0000_26_cloud-controller-manager-operator_00_namespace.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    openshift.io/node-selector: ""
```
JoelSpeed, line 7: What's this annotation for? 🤔

JoelSpeed (commented):
This is a great start, wouldn't really have much issue merging as is apart from the one comment below. PTAL at my other comments throughout, not sure if they want to be fixed here or later

You need to remove the bin folder from this commit and add a gitignore to it, currently we have a binary controller-gen committed

controllers/clusteroperator_controller.go
Danil-Grigorev: I was trying to lock up `library-go` for now. But putting a clusterOpreator template logic already did it, so it is no longer needed.

manifests/0000_26_cloud-controller-manager-operator_00_namespace.yaml
Danil-Grigorev: This is a plain merge/copy from https://github.com/openshift/cluster-kube-controller-manager-operator/blob/master/manifests/0000_25_kube-controller-manager-operator_00_namespace.yaml

Danil-Grigorev:
Just tested it manually in a cluster, builds, runs and serves the CO and updates it's status. Let's add `Deployment` for operator to consume by CVO later.
```bash
dgrigore cluster-cloud-controller-manager-operator  $ oc get clusteroperator cloud-controller-manager 
NAME                       VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE
cloud-controller-manager   0.0.1     True        False         False      61s
```

PTAL: @JoelSpeed @Fedosin @mdbooth

controllers/suite_test.go
Danil-Grigorev: Let's discuss it and change when we will have actual tests. Suggestion makes sense to me.

controllers/clusteroperator_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+)
+
+const (
+	clusterOperatorName = "cloud-controller-manager"
```
JoelSpeed, line 39: Does this follow convention? This is the name of the operand rather than the operator?

Danil-Grigorev: You are right. Just looking on the list there is not much convention, but the name I picked is definitely for the operand. Changing it to `openshift-cloud-controller-manager` will probably be best.
```
NAME                                       VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE
authentication                             4.6.9     True        False         False      146m
cloud-controller-manager                   0.0.1     True        False         False      127m
cloud-credential                           4.6.9     True        False         False      177m
cluster-autoscaler                         4.6.9     True        False         False      171m
config-operator                            4.6.9     True        False         False      172m
console                                    4.6.9     True        False         False      157m
csi-snapshot-controller                    4.6.9     True        False         False      172m
dns                                        4.6.9     True        False         False      170m
etcd                                       4.6.9     True        False         False      170m
image-registry                             4.6.9     True        False         False      162m
ingress                                    4.6.9     True        False         False      162m
insights                                   4.6.9     True        False         False      172m
kube-apiserver                             4.6.9     True        False         False      168m
kube-controller-manager                    4.6.9     True        False         False      169m
kube-scheduler                             4.6.9     True        False         False      169m
kube-storage-version-migrator              4.6.9     True        False         False      162m
machine-api                                4.6.9     True        False         False      165m
machine-approver                           4.6.9     True        False         False      171m
machine-config                             4.6.9     True        False         False      170m
marketplace                                4.6.9     True        False         False      170m
monitoring                                 4.6.9     True        False         False      161m
network                                    4.6.9     True        False         False      172m
node-tuning                                4.6.9     True        False         False      172m
openshift-apiserver                        4.6.9     True        False         False      165m
openshift-controller-manager               4.6.9     True        False         False      169m
openshift-samples                          4.6.9     True        False         False      164m
operator-lifecycle-manager                 4.6.9     True        False         False      171m
operator-lifecycle-manager-catalog         4.6.9     True        False         False      171m
operator-lifecycle-manager-packageserver   4.6.9     True        False         False      165m
service-ca                                 4.6.9     True        False         False      172m
storage                                    4.6.9     True        False         False      171m
```

JoelSpeed: Looks like `openshift-cloud-controller-manager` or just `cloud-controller-manager` would be suitable, I think looking at the list I'd lean towards the latter. You updated it to include operator in the name which doesn't seem to fit with the list above

manifests/0000_26_cloud-controller-manager-operator_00_namespace.yaml
Fedosin: I believe it comes from OpenShift 3: https://docs.openshift.com/container-platform/3.11/admin_guide/managing_projects.html#setting-the-project-wide-node-selector

> If openshift.io/node-selector is set to an empty string (oc adm new-project --node-selector=""), the project will not have an administrator-set node selector, even if the cluster-wide default has been set. This means that, as a cluster administrator, you can set a default to restrict developer projects to a subset of nodes and still enable infrastructure or other projects to schedule the entire cluster.

So basically this annotation unsets the default cluster-wide node-selector for the namespace. I don't know if it's still valid for OpenShift 4 but I suggest to keep the annotation.

controllers/clusteroperator_controller.go
Danil-Grigorev: My bad, should be ok now :D

Fedosin: I prefer `cloud-controller-manager` too

JoelSpeed: Let's go with `cloud-controller-manager`, sorry for the confusion @Danil-Grigorev

Danil-Grigorev: NP, it is sealed.

JoelSpeed (approved):
/lgtm

Fedosin (commented):
/lgtm

Fedosin (approved):

PR #3, opened 2021-01-18
Operator deployment

Danil-Grigorev:
Manifest for image referencing and operator deployment, which will be later populated by CI.

To test image in production here is a temporary operator image location `quay.io/dgrigore/cluster-cloud-controller-manager-operator:latest`.

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
@@ -1,0 +1,48 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cluster-cloud-controller-manager
+  namespace: openshift-cloud-controller-manager-operator
+  labels:
+    k8s-app: cloud-manager-operator
+spec:
+  selector:
+    matchLabels:
+      k8s-app: cloud-manager-operator
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        k8s-app: cloud-manager-operator
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: cluster-cloud-controller-manager
+      containers:
+      - name: cluster-cloud-controller-manager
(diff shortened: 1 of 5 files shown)


manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
```diff
@@ -0,0 +1,48 @@
...
+  selector:
+    matchLabels:
+      k8s-app: cloud-manager-operator
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        k8s-app: cloud-manager-operator
+    spec:
+      priorityClassName: system-node-critical
```
JoelSpeed, line 18: What are the options here? Does this mean this component is critical if we want nodes to work correctly?

JoelSpeed (commented):
Should the directory be called manifests or install? It's install in MAO and CAO 🤔

manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
Danil-Grigorev: According to [doc](https://github.com/openshift/openshift-docs/blob/3f1927c4088f97385bd44b53ee0b62480958a67f/modules/nodes-pods-priority-about.adoc) there are a couple. I'm following the [practices](https://github.com/openshift/machine-api-operator/blob/331808946db5135185c82792b10ee430ec4b46c1/install/0000_30_machine-api-operator_11_deployment.yaml#L21) of both `MAO` and `KCM` here.

Danil-Grigorev:
I feel like manifests is a better variant from those two. It is used mostly anywhere: https://github.com/openshift/cluster-kube-controller-manager-operator/tree/master/manifests

JoelSpeed:
So long as the manifests can be picked up in the release payload I'm happy
/lgtm

Fedosin (approved):
/lgtm
/approve

PR #6, opened 2021-01-21
Add missing "operator" suffix to the image name

Fedosin:


JoelSpeed (commented):
/approve
This seems to be inline with existing images

PR #9, opened 2021-01-26
Initial implementation for resource provision with SSA

Danil-Grigorev:
Implemented for `Infrastructure` resource.

### DIFF
--- a/cmd/cluster-cloud-controller-manager-operator/main.go
+++ b/cmd/cluster-cloud-controller-manager-operator/main.go
@@ -48,6 +48,8 @@ var (
 	retryPeriod   = 90 * time.Second
 )
 
+const defaultManagedNamespace = "cluster-cloud-controller-manager"
+
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 	utilruntime.Must(configv1.AddToScheme(scheme))
@@ -89,12 +91,19 @@ func main() {
 		"The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled.",
 	)
 
+	managedNamespace := flag.String(
+		"namespace",
+		defaultManagedNamespace,
+		"The namespace for managed objects, where out-of-tree CCM binaries will run.",
+	)
+
 	flag.Parse()
 
 	ctrl.SetLogger(klogr.New().WithName("CCMOperator"))
 
 	syncPeriod := 10 * time.Minute
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Namespace:               *managedNamespace,
 		Scheme:                  scheme,
 		SyncPeriod:              &syncPeriod,
 		MetricsBindAddress:      *metricsAddr,
--- a/controllers/cache.go
+++ b/controllers/cache.go
@@ -1,0 +1,136 @@
+package controllers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+type WatcherOptions struct {
+	Cache  cache.Cache
+	Scheme *runtime.Scheme
+}
+
+type ObjectWatcher interface {
+	Watch(ctx context.Context, obj client.Object) error
+	EventStream() <-chan event.GenericEvent
+}
+
+func NewObjectWatcher(opts WatcherOptions) (ObjectWatcher, error) {
+	if opts.Cache == nil {
+		return nil, errors.New("Cache is required")
+	}
+
+	// Use the default Kubernetes Scheme if unset
+	if opts.Scheme == nil {
+		opts.Scheme = scheme.Scheme
+	}
+
+	return &objectWatcher{
+		objectCache:      opts.Cache,
+		scheme:           opts.Scheme,
+		eventChan:        make(chan event.GenericEvent),
+		watchedResources: make(map[string]struct{}),
+	}, nil
+}
+
+type objectWatcher struct {
+	objectCache      cache.Cache
+	scheme           *runtime.Scheme
+	eventChan        chan event.GenericEvent
+	watchedResources map[string]struct{}
+}
+
+func (n *objectWatcher) EventStream() <-chan event.GenericEvent {
+	return n.eventChan
+}
+
+func (n *objectWatcher) Watch(ctx context.Context, obj client.Object) error {
+	key, err := n.watchKey(obj)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := n.watchedResources[key]; !ok {
+		// watch not set up for this object yet
+		return n.watch(ctx, obj)
+	}
+
+	return nil
+}
+
+func (n *objectWatcher) watch(ctx context.Context, obj client.Object) error {
+	informer, err := n.objectCache.GetInformer(ctx, obj)
+	if err != nil {
+		return nil
+	}
+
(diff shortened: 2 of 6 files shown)


cmd/cluster-cloud-controller-manager-operator/main.go
```diff
@@ -112,9 +114,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	platform := &platform.InfrastrucutreOwner{}
+
+	setupLog.Info("Waiting for platform objects")
+	if err = wait.PollImmediate(
```
Danil-Grigorev, line 120: This should go away. We need to consider how to add and remove informers on managed resources dynamically during reconcile.

pkg/cloud/aws.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func GetAWSResources() OwnedResources {
+	return OwnedResources{
+		newTestResource(),
+	}
+}
+
+func newTestResource() *v1.Service {
+	svc := &v1.Service{
```
Fedosin, line 15: Tbh, I don't like the idea of creating these objects manually. I would prefer to automate it and generate them from yaml files. Here is an example https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/10

cmd/cluster-cloud-controller-manager-operator/main.go
```diff
@@ -89,12 +91,19 @@ func main() {
 		"The duration that non-leader candidates will wait after observing a leadership renewal until attempting to acquire leadership of a led but unrenewed leader slot. This is effectively the maximum duration that a leader can be stopped before it is replaced by another candidate. This is only applicable if leader election is enabled.",
 	)
 
+	managedNamespace := flag.String(
+		"namespace",
+		defaultManagedNamespace,
+		"The namespace for managed objects, where out-of-tree CCM binaries will be watched",
```
JoelSpeed, line 98: This might make more sense to a user
```suggestion
		"The namespace for managed objects, where out-of-tree CCM binaries will run",
```

controllers/cache.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+var defaultResyncTime = 10 * time.Hour
```
JoelSpeed, line 17: Don't think this is used

controllers/clusteroperator_controller.go
```diff
@@ -54,14 +60,52 @@ type CloudOperatorReconciler struct {
...
 }
 
+func (r *CloudOperatorReconciler) sync(ctx context.Context, key client.ObjectKey) error {
+	resources, err := r.GetObjects(ctx, r.Client, key)
+	if err != nil {
+		return err
+	}
+
+	for _, resource := range resources {
+		if err = ApplyServerSide(ctx, r.Client, clusterOperatorName, resource); err != nil {
```
JoelSpeed, line 90: May as well make this declare in a new scope as with the other if below
```suggestion
		if err := ApplyServerSide(ctx, r.Client, clusterOperatorName, resource); err != nil {
```

pkg/cloud/aws.go
Danil-Grigorev: Let's shelf it for now and return once we have a reference implementation with identified things we need to configure dynamically.

controllers/clusteroperator_controller.go
```diff
@@ -144,17 +188,48 @@ func (r *CloudOperatorReconciler) syncStatus(ctx context.Context, co *configv1.C
...
+	}
+	r.watcher = watcher
+
+	build := ctrl.NewControllerManagedBy(mgr).
+		For(
+			&configv1.ClusterOperator{},
+			builder.WithPredicates(clusterOperatorPredicates()),
+		).
+		Watches(
+			&source.Kind{Type: r.Object()},
```
JoelSpeed, line 206: `r.Object` is very not obvious here what this is, where does this come from, the platform owner? 🤔  Might be best to just have the type in here so it's obvious
```suggestion
			&source.Kind{Type: &configv1.Infrastructure{}},
```

pkg/platform/infrastructure.go
```diff
@@ -0,0 +1,57 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"k8s.io/klog/v2"
+)
+
+var _ PlatformOwner = &InfrastrucutreOwner{}
```
JoelSpeed, line 14: I'm not certain what the benefit of this abstraction is at the moment? What are we trying to achieve? Should we not just look up the `configv1.Infrastructure` at the beginning of the reconcile as we normally do in operators?

controllers/clusteroperator_controller.go
Danil-Grigorev: I'll show a POC for CAPI resources later. It is meant to do that.

Danil-Grigorev: @JoelSpeed Here is a POC how it will work for CAPI cluster with minimal set of changes: https://github.com/openshift/cluster-cloud-controller-manager-operator/commit/d844b87d308df3f24474f42193063895885e4c2f

pkg/platform/infrastructure.go
Danil-Grigorev: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/9#discussion_r565975485

controllers/clusteroperator_controller.go
JoelSpeed: Are we planning on integrating with CAPI? I had assumed this would be an OpenShift specific project and therefore we would always rely on the Infrastructure object. I think unless we have an immediate need for supporting CAPI we should keep this simple and perhaps only add this abstraction when we actually need it

Danil-Grigorev: I'm addressing the comment on the proposal. https://github.com/openshift/enhancements/pull/463#discussion_r561994942

Danil-Grigorev: Let's leave this implementation on it's own branch and we can return to the interface if it is needed. Ok with watching `Infrastructure` resources only.

Danil-Grigorev:
@JoelSpeed the PR is updated according to suggestions.

controllers/clusteroperator_controller.go
```diff
@@ -54,14 +59,66 @@ type CloudOperatorReconciler struct {
 
 // Reconcile will process the cloud-controller-manager clusterOperator
 func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	infraList := &configv1.InfrastructureList{}
+	if err := r.List(ctx, infraList); err != nil {
+		klog.Errorf("Unable to retrive list of Infrastructure objects: %v", err)
+		return ctrl.Result{}, err
+	}
```
JoelSpeed, line 66: Shouldn't there only be one infrastructure we care about, I would expect us to get just that one rather than listing all of them and iterating through? I think this is optimising for a case that won't happen any time soon in OCP and  might lead to issues

JoelSpeed (commented):
One little thing otherwise I think we are good to go

controllers/clusteroperator_controller.go
Fedosin:
```suggestion
	infra := configv1.Infrastructure{}
	
	if err := r.Get(ctx, types.NamespacedName{Name: "cluster", Namespace: ""}, &infra); err != nil {
		klog.Errorf("Unable to retrive the infrastructure object: %v", err)
		return ctrl.Result{}, err
	}

	resources := getResources(infra)
	if err := r.sync(ctx, resources); err != nil {
		klog.Errorf("Unable to sync operands: %s", err)
	}
```

Danil-Grigorev: All updated @Fedosin @JoelSpeed PTAL

Fedosin (commented):
/lgtm

JoelSpeed (commented):
/approve

Fedosin (commented):
/lgtm

PR #10, opened 2021-01-26
Add OpenStack resources

Fedosin:


assets/openstack/credentials_request.yaml
```diff
@@ -0,0 +1,16 @@
+
```
Danil-Grigorev, line 1: I think this should belong to a regular `manifests` folder. cloud-credentials-operator will filter which credentials to pass according to current cloud provider.

pkg/cloud/openstack.go
```diff
@@ -0,0 +1,138 @@
...
+var openstackClusterRoleBinding []byte
+
+func GetOpenStackResources() []client.Object {
+	config := newConfigResource()
+	if config == nil {
+		return nil
+	}
+
+	deployment := newDeploymentResource()
+	if config == nil {
```
mdbooth, line 40: s/config/deployment/

Cut/paste errors all the way down :)

Fedosin: yep :(

pkg/cloud/assets/openstack/deployment.yaml
```diff
@@ -0,0 +1,64 @@
...
+          operator: Exists
+        - key: node-role.kubernetes.io/master
+          operator: Exists
+          effect: "NoSchedule"
+        - key: node.cloudprovider.kubernetes.io/uninitialized
+          value: "true"
+          effect: NoSchedule
+      containers:
+        - name: openstack-cloud-controller-manager
+          image: docker.io/k8scloudprovider/openstack-cloud-controller-manager:latest
```
Danil-Grigorev, line 33: This should be managed by `CCCMO` and match the image version from in-cluster image registry., out-of-scope for this PR though.

mdbooth:
Not going to bother with a separate PR, but I posted a suggested tidy up in this branch: https://github.com/mdbooth/cluster-cloud-controller-manager-operator/tree/bindata

pkg/cloud/openstack/openstack.go
```diff
@@ -0,0 +1,48 @@
...
+		{&rbacv1.Role{}, "assets/rbac/role.yaml"},
+		{&rbacv1.RoleBinding{}, "assets/rbac/rolebinding.yaml"},
+		{&rbacv1.ClusterRoleBinding{}, "assets/rbac/clusterrolebinding.yaml"},
+	}
+	ret := make([]client.Object, 0, len(resources))
+
+	for i := 0; i < len(resources); i++ {
+		resource := &resources[i]
+		data, err := f.ReadFile(resource.asset)
+		if err != nil {
```
mdbooth, line 35: Not important, but I was wondering about it when reading embed.FS yesterday: do we need error handling here? Can this actually fail, except for an invalid path which would be picked up by the unit test.

pkg/cloud/openstack/openstack.go
```diff
@@ -0,0 +1,48 @@
...
+			return nil
+		}
+
+		dec := k8sYaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), 1000)
+		if err := dec.Decode(resource.obj); err != nil {
```
mdbooth, line 40: On the topic of error handling...

I wonder if we should move static asset parsing to an init block and panic on failure. @Danil-Grigorev?

Danil-Grigorev: I think we should properly pass cluster configuration in these methods and handle errors as usual. For now logging the error will be sufficient.

mdbooth: I was just thinking that it's a 'static' failure. That is, the only resolution is to build a new image. Does continuing to run serve any purpose if it's never going to work? Presumably we would never ship anything in this state, but it would also make debugging the source of a CI failure simpler.

JoelSpeed: IMO we should be panicking if we can't decode something that we have included in the binary by embedding. To me this signifies a programming error, hence a panic is suitable.

If the controller panics when it can't parse an embedded resource, this will be flagged up in tests and should prevent anyone checking in broken YAML. 

Additionally, a unit test that runs `GetResources` would be obviously broken if it panics and would flag any errors being added

Fedosin: I added logging to show proper error messages

pkg/cloud/openstack/openstack.go
```diff
@@ -0,0 +1,48 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	k8sYaml "k8s.io/apimachinery/pkg/util/yaml"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:embed assets/*
```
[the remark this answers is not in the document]
mdbooth: I believe the unit tests are currently covering this, so it should be ok

pkg/cloud/openstack/assets/deployment.yaml
```diff
@@ -0,0 +1,64 @@
...
+          operator: Exists
+        - key: node-role.kubernetes.io/master
+          operator: Exists
+          effect: "NoSchedule"
+        - key: node.cloudprovider.kubernetes.io/uninitialized
+          value: "true"
+          effect: NoSchedule
+      containers:
+        - name: openstack-cloud-controller-manager
+          image: docker.io/k8scloudprovider/openstack-cloud-controller-manager:latest
```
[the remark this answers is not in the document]
mdbooth: Is there some API for pulling references from current release? I suggest the 'driver' probably needs to do this directly, as depending on the cloud provider it could potentially use arbitrarily many images.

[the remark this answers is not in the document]
Danil-Grigorev: Yes, there is usually a combination of [image stream](https://github.com/openshift/machine-api-operator/blob/master/install/image-references), and preliminary [location](https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_01_images.configmap.yaml) for storing resolved image hashes, like a configmap for example. All operator is doing is making sure the operands image matches the one from configmap.

pkg/cloud/openstack/assets/deployment.yaml
```diff
@@ -0,0 +1,64 @@
...
+  selector:
+    matchLabels:
+      app: openstack-cloud-controller-manager
+  template:
+    metadata:
+      labels:
+        app: openstack-cloud-controller-manager
+    spec:
+      hostNetwork: true
+      serviceAccount: cluster-cloud-controller-manager
```
Danil-Grigorev, line 18: This SA apparently does not exist. @Fedosin

Danil-Grigorev: Also, it should probably follow or inherit or extend this SA with roles: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/15/files#diff-2b0c81c2d79e6551f107545166e6281fbb47a343b7080d29c87c0c2637cd76b4R1-R134 

As far as I know openstack CCM definitely needs access to manage nodes: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/15/files#diff-2b0c81c2d79e6551f107545166e6281fbb47a343b7080d29c87c0c2637cd76b4R30-R39

mdbooth:
There are still a bunch of issues here, but at this stage of development I think it's more important to get this merged. We can fix it up as we go.

Lets not even wait for the tests to pass: you can't have regressions if you have no users because it has never worked!

/lgtm

pkg/cloud/openstack/assets/rbac/clusterrolebinding.yaml
```diff
@@ -0,0 +1,14 @@
...
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+roleRef:
+  kind: ClusterRole
+  name: cloud-controller-manager
+subjects:
+- kind: ServiceAccount
+  namespace: openshift-cloud-controller-manager
+  name: openstack-cloud-controller-manager
```
Danil-Grigorev, line 14:
```suggestion
  name: cloud-controller-manager
```

This will be reused by all providers.

pkg/cloud/openstack/assets/rbac/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: openshift-cloud-controller-manager
+  name: openstack-cloud-controller-manager
```
Danil-Grigorev, line 5:
```suggestion
  name: cloud-controller-manager
```

Let's centralize the naming and move it to manifests for other providers to use.

pkg/cloud/openstack/assets/rbac/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ServiceAccount
```
Danil-Grigorev, line 2: Please move this file to manifests folder.

pkg/cloud/openstack/assets/rbac/rolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+roleRef:
+  kind: Role
+  name: openstack-cloud-controller-manager
+subjects:
+  - kind: ServiceAccount
+    namespace: openshift-cloud-controller-manager
+    name: openstack-cloud-controller-manager
```
Danil-Grigorev, line 15:
```suggestion
    name: cloud-controller-manager
```

Danil-Grigorev (commented):
Just some refactoring to move things in correct folders, and this PR is ready to be merged.

pkg/cloud/openstack/assets/deployment.yaml
Danil-Grigorev:
```suggestion
      serviceAccount: cloud-controller-manager
```

Danil-Grigorev (approved):
/approve

PR #13, opened 2021-02-02
Add release version reporting

Danil-Grigorev:
Operator is supposed to report release version to be considered healthy.


Fedosin (approved):
/lgtm
/approve

PR #15, opened 2021-02-11
Initial AWS CCM deployment using upstream image

Danil-Grigorev:
Requirements to work:
----
Needs to additionally be merged:
- [ ] https://github.com/openshift/machine-config-operator/pull/2401 - fix for AWS hostname for successful migration
- [ ] https://github.com/openshift/machine-config-operator/pull/2386 - Kubelet FG switch to external providers
- [ ] https://github.com/openshift/library-go/pull/895 (+ revendor in KCMO and KAPIO) - KCM and KAPI FG switch to external providers
- [x] https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/26 - Include in CCCMO in release

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_03_rbac_provider.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_03_rbac_provider.yaml
@@ -10,57 +10,6 @@ metadata:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
-metadata:
-  name: cloud-controller-manager
-  namespace: kube-system
-  annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
-rules:
-  - apiGroups:
-      - ""
-    resources:
-      - serviceaccounts
-    resourceNames:
-      - node-controller
-      - service-controller
-    verbs:
-      - get
-
-  # This grants CCMs permission to list and read all secrets in kube-system.
-  # This is required by client builder to fetch the ServiceAccount token
-  # secret. Ideally we would restrict this to the token secrets of specific
-  # ServiceAccounts.
-  - apiGroups:
-      - ""
-    resources:
-      - secrets
-    verbs:
-      - list
-      - get
-      - watch
-
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
-metadata:
-  name: cloud-controller-manager
-  namespace: kube-system
-  annotations:
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
-roleRef:
-  kind: Role
-  name: cloud-controller-manager
-  apiGroup: rbac.authorization.k8s.io
-subjects:
-  - kind: ServiceAccount
-    namespace: openshift-cloud-controller-manager
-    name: cloud-controller-manager
-
----
-apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: cloud-controller-manager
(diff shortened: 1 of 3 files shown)


Danil-Grigorev:
/hold cancel

This implementation works in my manual testing. Unit tests are already in place, the PR is ready for review.

manifests/0000_26_cloud-controller-manager-operator_03_rbac_provider.yaml
```diff
@@ -8,57 +8,6 @@ metadata:
...
-    include.release.openshift.io/single-node-developer: "true"
-roleRef:
-  kind: Role
-  name: cloud-controller-manager
-  apiGroup: rbac.authorization.k8s.io
-subjects:
-  - kind: ServiceAccount
-    namespace: openshift-cloud-controller-manager
-    name: cloud-controller-manager
-
```
Danil-Grigorev, line 61: AWS covers these RBAC changes with upaded cluster role [here](https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/15/files#diff-0856a88d16f3aa369e7f5d7fc7d1e356008fc796a5800ad4da0e304113f6af2cR60). I consider removing this for now, as for example Azure and vSphere both are using same clusterRole, which will work for OpenStack too. We will investigate and possible reduce the scope of it later.

Fedosin (changes requested):
What about adding unit tests similar to what we have with OpenStack: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/openstack/openstack_test.go

Danil-Grigorev:
@Fedosin It is already implemented and covered in https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/controllers/clusteroperator_controller_test.go#L271-L336

manifests/0000_26_cloud-controller-manager-operator_03_rbac_provider.yaml
mdbooth: Ok, sure. As I mentioned elsewhere I think we could potentially eliminate these entirely later.

pkg/cloud/aws.go
```diff
@@ -1,9 +1,44 @@
...
+//go:embed aws/assets/*
+var f embed.FS
+
 func GetAWSResources() []client.Object {
-	return []client.Object{}
+	resources := []struct {
+		obj   client.Object
+		asset string
+	}{
+		{&appsv1.Deployment{}, "aws/assets/deployment.yaml"},
```
elmiko, line 22: is the idea here that more resources will be added later?

pkg/cloud/aws.go
```diff
@@ -1,9 +1,44 @@
...
+	}
+	ret := make([]client.Object, 0, len(resources))
+
+	for i := 0; i < len(resources); i++ {
```
elmiko, line 26: why not use `for ... range` here?
```suggestion
	for _, resource := range resources {
```

elmiko (commented):
the code mostly makes sense, i have a few questions inline. also, is the main goal here for the code to dynamically load these manifests at run time?

pkg/cloud/aws.go
Danil-Grigorev: Correct, it is build with intention that it might need to deploy some SA or another Deployment potentially in the future.

elmiko (approved):
thanks Danil!
/approve

elmiko:
/hold
didn't realize it added both labels. just putting a hold so others can review

Danil-Grigorev:
/lgtm cancel

We need to fix the CI to prevent double label addition @Fedosin 
Let's wait for tests to pass again.

elmiko:
> Can we make this look more similar to the openstack one? Keeping it in its own package in case it expands later?

+1, i think that makes good sense

Danil-Grigorev:
@JoelSpeed OpenStack manifests are located incorrectly. They should belong to higher level, so cloud provider switch case selection won't have to import N packages per provider just to select correct collection of manifests: see https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/controllers/clusteroperator_controller.go#L25-L26

PR #16, opened 2021-02-15
[OCPCLOUD-1089] Simple e2e test suite for CCCMO clusterOperator 

Danil-Grigorev:
Make sure the `ClusterOperator` is provisioned, have necessary fields populated and marked available even when there is no infrastructure in place to provision.

### DIFF
--- a/controllers/clusteroperator_controller_test.go
+++ b/controllers/clusteroperator_controller_test.go
@@ -1,0 +1,187 @@
+package controllers
+
+import (
+	"context"
+	"os"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	timeout = time.Second * 10
+)
+
+var _ = Describe("Cluster Operator status controller", func() {
+	var operator *configv1.ClusterOperator
+	var operatorController *CloudOperatorReconciler
+
+	BeforeEach(func() {
+		operatorController = &CloudOperatorReconciler{
+			Client:           cl,
+			Scheme:           scheme.Scheme,
+			ManagedNamespace: defaultManagementNamespace,
+		}
+		operator = &configv1.ClusterOperator{}
+		operator.SetName(clusterOperatorName)
+	})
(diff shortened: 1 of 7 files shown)


controllers/clusteroperator_controller_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	type testCase struct {
+		releaseVersionEnvVariableValue string
+		namespace                      string
+		existingCO                     *configv1.ClusterOperator
+	}
+
+	DescribeTable("should ensure Cluster Operator status is present",
+		func(tc testCase) {
+			By("Running test case")
```
JoelSpeed, line 55: By statements are more useful when they describe parts of the test, so for example you can do:
```
By("Setting the release version", func() {
  expectedVersion := unknownVersionValue
  if tc.releaseVersionEnvVariableValue != "" {
    expectedVersion = tc.releaseVersionEnvVariableValue
    Expect(os.Setenv(releaseVersionEnvVariableName, tc.releaseVersionEnvVariableValue)).To(Succeed())
  }
})
```
Currently this By doesn't add much value

controllers/clusteroperator_controller_test.go
```diff
@@ -0,0 +1,182 @@
...
+				Expect(err).To(Succeed())
+			}
+
+			_, err := operatorController.Reconcile(context.Background(), reconcile.Request{})
+			Expect(err).To(Succeed())
+
+			getOp := &configv1.ClusterOperator{}
+			Eventually(func() (bool, error) {
+				var err error
+				err = cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, getOp)
```
JoelSpeed, line 77: Can inline this
```suggestion
				err := cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, getOp)
```

controllers/clusteroperator_controller_test.go
```diff
@@ -0,0 +1,183 @@
...
+		operator.SetName(clusterOperatorName)
+	})
+
+	AfterEach(func() {
+		os.Unsetenv(releaseVersionEnvVariableName)
+		err := cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, &configv1.ClusterOperator{})
+		if err == nil || !apierrors.IsNotFound(err) {
+			Expect(cl.Delete(context.Background(), operator)).To(Succeed())
+		}
+		Expect(apierrors.IsNotFound(cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, &configv1.ClusterOperator{}))).To(BeTrue())
```
JoelSpeed, line 44: I would suggest using async calls here to make sure these are reliable

```
Eventually(func() error {
  err := cl.Delete(context.Background(), operator)
  if err != nil && !apierrors.IsNotFound(err) {
    return err
  }
  return nil
}).Should(Succeed())
Eventually(func() error {
  err := cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, &configv1.ClusterOperator{})
  if apierrors.IsNotFound(err) {
    return nil
  }
  return errors.New("object still exists")
}).Should(Succeed())
```

Fedosin (approved):
/lgtm
/approve

PR #19, opened 2021-02-23
hack: Add operator rebuild-and-redeploy script for library-go changes

mdbooth:


hack/update-library-go.sh
```diff
@@ -0,0 +1,125 @@
...
+    [ ! -d "${sourcedir}" ] && \
+        git clone "https://github.com/openshift/${repo}.git" "${sourcedir}"
+    pushd "${sourcedir}"
+        go mod edit --replace "github.com/openshift/library-go=$librarygo"
+        go mod vendor
+
+        # This dance is so we get build output during execution while also
+        # capturing it
+        mkfifo /tmp/buildah.$$
+        buildah bud -t "${image}" Dockerfile.rhel7 | tee /tmp/buildah.$$ &
```
Danil-Grigorev, line 59: Would prefer having a way to use podman/docker for those tasks.

Danil-Grigorev (commented):
Would like to test the PR locally later this week, but visually lgtm with one comment.

hack/update-library-go.sh
mdbooth: buildah is from the same stable as podman. It's the the bit that does builds. `podman build` will use `buildah bud` under the covers.

We could add docker support, but given that AFAIK we don't support it I'm not sure it's worth the hassle.

Danil-Grigorev (approved):
/lgtm
/approve

PR #21, opened 2021-03-10
Mitigate hot-loop in SSA

Danil-Grigorev:
Compare objects on update, while stripping fields which change on no-op
resource update.

Avoiding https://github.com/kubernetes/kubernetes/issues/100024

### DIFF
--- a/controllers/cache.go
+++ b/controllers/cache.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
Code changes seem reasonable, how has this been tested? 
/approve

Danil-Grigorev:
@mdbooth @Fedosin PTAL.

Danil-Grigorev:
@JoelSpeed Yes, for now this PR was tested manually. This PR together with openstack [bits](https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/10) was deployed and didn't end up in endless updates as @mdbooth displayed in his manual testing.

controllers/cache.go
```diff
@@ -104,33 +105,54 @@ type eventToChannelHandler struct {
...
+		}
+
+		// We'd like to compare both objects ignoring changes in managedFields and resourceVersion
+		// Nither of those fields should be applied, so it is ok to strip
+		// Mitigating issue: https://github.com/kubernetes/kubernetes/issues/100024
+
+		old.SetManagedFields(nil)
+		new.SetManagedFields(nil)
+		old.SetResourceVersion("")
+		new.SetResourceVersion("")
```
mdbooth, line 146: These operations add a side-effect that wasn't previously there. We should copy old and new before modifying them.

mdbooth (changes requested):
Have I misunderstood the context here? I don't think this patch directly addresses the issue with SSA patch. Instead, it causes the controller not to call Reconcile for changes to only ManagedFields or ResourceVersion, right?

If so, that seems like a change with a potentially large blast radius. Either way, I think the comment could be more explicit about this (again, assuming I've correctly understood the context!).

controllers/cache.go
Danil-Grigorev: You are right. It is now updated to work with resource copy.

mdbooth (approved):
/lgtm

PR #22, opened 2021-03-17
Bindata

mdbooth:
See the commit message of the third commit for interesting detail.


mdbooth:
To discuss:

* Are any of these permission changes surprising?
* Can/should we use a common SA for all cloud providers?

Also, we could minimise these much further by changing the same the operator consumes the node-controller and service-controller service accounts.

PR #24, opened 2021-03-18
OpenStack CCM RBAC changes

mdbooth:
This change does 2 things:
* Move all RBAC permissions from assets to manifests
* Minimises the permissions of cloud-controller-manager

Specifically:

* Lease permissions are now confined to a single lease object. They were
  previously granted cluster-wide
* Events are now confined to the CCM namespace and create only. They
  were previous read/write and cluster wide
* Cluster-wide node and service permissions are now read-only
* PersistentVolume permissions are removed
* Endpoint permissions are removed
* ConfigMap permissions are removed
* Secret permissions are confined to kube-system

Note that CCMs using k8s.io/cloud-provider explicitly use the
node-controller and service-controller service accounts from kube-system
with the relevent controllers. However, the watches used by those
controllers use the operator's credentials. This means that:

* The operator requires read-only permissions for objects that the
  controllers manage
* The operator requires permissions sufficient to fetch node-operator
  and service-operator credentials from kube-system


JoelSpeed (approved):
/approve

PR #27, opened 2021-03-24
Unify pre-provisioning checks in a single location

Danil-Grigorev:
This implementation improves readability of the code by dedicating a specific domain for pre-provision cluster checks.

KCM-CCM migration will belong to one of such checks. Implementation TBD

### DIFF
--- a/controllers/clusteroperator_controller.go
+++ b/controllers/clusteroperator_controller.go
@@ -67,18 +67,12 @@ type CloudOperatorReconciler struct {
 
 // Reconcile will process the cloud-controller-manager clusterOperator
 func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
-	featureGate := &configv1.FeatureGate{}
-	if err := r.Get(ctx, client.ObjectKey{Name: externalFeatureGateName}, featureGate); errors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


controllers/clusteroperator_controller.go
```diff
@@ -67,18 +67,57 @@ type CloudOperatorReconciler struct {
...
 func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	allowedToProvision, err := r.provisioningAllowed(ctx)
+	if err != nil {
+		klog.Errorf("Unable to determine cluster state %v", err)
+		return ctrl.Result{}, err
+	} else if !allowedToProvision {
+		return ctrl.Result{}, nil
+	}
+
+	infra := &configv1.Infrastructure{}
```
JoelSpeed, line 78: Since we need the infra in `provisioningAllowed`, why not fetch that at this scope, and pass it into provisioning allowed? I think that would deduplicate the fetching logic here right?

elmiko: +1, that makes sense to me too

JoelSpeed:
This is quite out of date? Is it still relevant?

Fedosin:
/close

this change is implemented here: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/122

PR #28, opened 2021-03-24
Add images config map

Danil-Grigorev:
Mount config map into `CCCMO` deployment for correct image version
evaluation on operator startup. Images will be dynamically passed to
operand controllers.

To deploy operator locally you now need to specify a command line argument:
```bash
./bin/cluster-controller-manager-operator --images-json=hack/example-images.json
```

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cloud-controller-manager-images
+  namespace: openshift-cloud-controller-manager-operator
+data:
+  images.json: >
+    {
+      "cloudControllerManagerAWS": "registry.svc.ci.openshift.org/openshift:aws-cloud-controller-manager",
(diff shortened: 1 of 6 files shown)


manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
JoelSpeed, line 1: Do we also need an image-references as we do in MAO to get the images substituted correctly? https://github.com/openshift/machine-api-operator/blob/master/install/image-references

Danil-Grigorev: Yeah, right. I added these lines in the file :+1:

elmiko: there is an image-reference below, is that not sufficient?

elmiko (commented):
this makes sense to me, curious about Joel's comment.

elmiko (approved):
/lgtm

manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
JoelSpeed: It is sufficient, but it wasn't there when I reviewed ;)

PR #29, opened 2021-03-25
Provider image substitution

Danil-Grigorev:
This PR allows operator to dynamically configure image in provisioned resources such as CCM Deployment. 

This is done based on assumption that every CCM controller will be in a named container containing `controller` substring in it's name. All those containers images will be managed by operator, others will remain unchanged from initial manifest.

Operator is now populating Namespace for each resource it provisions. Cluster-scoped resources are also getting this namespace value, which is ok as it is not accounted and is not stored by API server on apply. This allows two operators to manage equivalent set of resources in different namespaces without clashes.

- [x] Testing

### DIFF
--- a/controllers/clusteroperator_controller.go
+++ b/controllers/clusteroperator_controller.go
@@ -97,8 +97,16 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	platform, err := getProviderFromInfrastructure(infra)
+	if err != nil {
+		klog.Errorf("Unable to determine platform from infrastructure: %s", err)
+		// Ignoring error here as infrastructure resource needs to be reconciled externally
+		// to provide correct platform
+		return ctrl.Result{}, nil
+	}
+
 	// Verify FeatureGate ExternalCloudProvider is enabled for operator to work in TP phase
-	external, err := cloudprovider.IsCloudProviderExternal(infra.Status.Platform, featureGate)
+	external, err := cloudprovider.IsCloudProviderExternal(platform, featureGate)
 	if err != nil {
 		klog.Errorf("Could not determine external cloud provider state: %v", err)
 		return ctrl.Result{}, err
@@ -113,8 +121,16 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
+	config, err := r.composeConfig(platform)
+	if err != nil {
+		klog.Errorf("Unable to build operator config %s", err)
+		return ctrl.Result{}, err
+	}
+
 	// Deploy resources for platform
-	resources := getResources(infra)
+	templates := getResources(platform)
+	resources := fillConfigValues(config, templates)
+
 	if err := r.sync(ctx, resources); err != nil {
 		klog.Errorf("Unable to sync operands: %s", err)
 		return ctrl.Result{}, err
@@ -162,13 +178,8 @@ func applyServerSide(ctx context.Context, c client.Client, owner client.FieldOwn
 	return c.Patch(ctx, obj, client.Apply, opts...)
 }
 
-func getResources(infra *configv1.Infrastructure) []client.Object {
-	if infra.Status == (configv1.InfrastructureStatus{}) {
-		klog.Warning("No platform value found in infrastructure")
-		return nil
-	}
-
-	switch infra.Status.Platform {
+func getResources(platform configv1.PlatformType) []client.Object {
+	switch platform {
 	case configv1.AWSPlatformType:
 		return cloud.GetAWSResources()
 	case configv1.OpenStackPlatformType:
@@ -178,6 +189,21 @@ func getResources(infra *configv1.Infrastructure) []client.Object {
 	}
 
 	return nil
+}
+
+func (r *CloudOperatorReconciler) composeConfig(platform configv1.PlatformType) (operatorConfig, error) {
+	config := operatorConfig{}
+
+	images, err := getImagesFromJSONFile(r.ImagesFile)
+	if err != nil {
+		klog.Errorf("Unable to decode images file from location %s", r.ImagesFile, err)
+		return config, err
+	}
+
+	config.ControllerImage = getProviderControllerFromImages(platform, images)
+	config.ManagedNamespace = r.ManagedNamespace
+
+	return config, nil
 }
 
 // statusAvailable sets the Available condition to True, with the given reason
--- a/controllers/clusteroperator_controller_test.go
+++ b/controllers/clusteroperator_controller_test.go
@@ -2,7 +2,9 @@ package controllers
 
 import (
 	"context"
+	"io/ioutil"
 	"os"
+	"testing"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -26,6 +28,7 @@ import (
 const (
 	timeout                 = time.Second * 10
 	testManagementNamespace = "openshift-cloud-controller-manager"
+	testImagesFilePath      = "../hack/example-images.json"
 )
 
 var _ = Describe("Cluster Operator status controller", func() {
@@ -236,6 +239,7 @@ var _ = Describe("Component sync controller", func() {
 			Scheme:           scheme.Scheme,
 			watcher:          w,
 			ManagedNamespace: testManagementNamespace,
+			ImagesFile:       testImagesFilePath,
 		}
 		originalWatcher, _ := w.(*objectWatcher)
 		watcher = mockedWatcher{watcher: originalWatcher}
@@ -265,6 +269,7 @@ var _ = Describe("Component sync controller", func() {
 	type testCase struct {
 		status          *configv1.InfrastructureStatus
 		featureGateSpec *configv1.FeatureGateSpec
+		config          operatorConfig
 		expected        []client.Object
 	}
 
@@ -285,8 +290,8 @@ var _ = Describe("Component sync controller", func() {
 
 			watchMap := watcher.getWatchedResources()
 
-			operands = tc.expected
-			for _, obj := range tc.expected {
+			operands = fillConfigValues(tc.config, tc.expected)
+			for _, obj := range operands {
 				Expect(watchMap[obj.GetName()]).ToNot(BeNil())
 
 				original, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj.DeepCopyObject())
@@ -307,13 +312,27 @@ var _ = Describe("Component sync controller", func() {
 		Entry("Should provision AWS resources", testCase{
 			status: &configv1.InfrastructureStatus{
 				Platform: configv1.AWSPlatformType,
+				PlatformStatus: &configv1.PlatformStatus{
+					Type: configv1.AWSPlatformType,
+				},
 			},
 			featureGateSpec: externalFeatureGateSpec,
-			expected:        cloud.GetAWSResources(),
+			config: operatorConfig{
+				ManagedNamespace: testManagementNamespace,
+				ControllerImage:  "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
+			},
+			expected: cloud.GetAWSResources(),
 		}),
 		Entry("Should provision OpenStack resources", testCase{
 			status: &configv1.InfrastructureStatus{
 				Platform: configv1.OpenStackPlatformType,
+				PlatformStatus: &configv1.PlatformStatus{
+					Type: configv1.OpenStackPlatformType,
+				},
+			},
+			config: operatorConfig{
+				ManagedNamespace: testManagementNamespace,
+				ControllerImage:  "registry.ci.openshift.org/openshift:openstack-cloud-controller-manager",
 			},
 			featureGateSpec: externalFeatureGateSpec,
 			expected:        openstack.GetResources(),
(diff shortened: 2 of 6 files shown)


controllers/clusteroperator_controller.go
```diff
@@ -180,6 +191,33 @@ func getResources(infra *configv1.Infrastructure) []client.Object {
 	return nil
 }
 
+func (r *CloudOperatorReconciler) composeConfig(ctx context.Context) (operatorConfig, error) {
```
JoelSpeed, line 194: We should pass the infra in here rather than getting in this function, as we have it in the parent scope anyway

controllers/clusteroperator_controller.go
```diff
@@ -180,6 +191,33 @@ func getResources(infra *configv1.Infrastructure) []client.Object {
...
+	if err != nil {
+		klog.Errorf("Unable to decode images file from location %s", r.ImagesFile, err)
+		return config, err
+	}
+
+	_, err = getProviderFromInfrastructure(infra)
+	if err != nil {
+		klog.Errorf("Unable to decode images file from location %s", r.ImagesFile, err)
+		return config, err
+	}
```
JoelSpeed, line 213: What's this? Do we not need the platform value? Also, if we already have this in the parent scope, why not pass it in?

controllers/clusteroperator_controller.go
```diff
@@ -180,6 +191,33 @@ func getResources(infra *configv1.Infrastructure) []client.Object {
...
+
+	config.ControllerImage = getProviderControllerFromImages(configv1.OpenStackPlatformType, *images)
```
JoelSpeed, line 215: Does the platform not need to go in here?

controllers/config.go
```diff
@@ -0,0 +1,57 @@
...
+		return "", fmt.Errorf("platform status is not pupulated on infrastructure")
+	}
+	if infra.Status.PlatformStatus.Type == "" {
+		return "", fmt.Errorf("no platform provider found on infrastructure")
+	}
+
+	return infra.Status.PlatformStatus.Type, nil
+}
+
+func getImagesFromJSONFile(filePath string) (*images, error) {
```
JoelSpeed, line 35: Since we don't need a point, maybe change the return type here?

controllers/config.go
```diff
@@ -0,0 +1,57 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// images allows build systems to inject images for CCCMO components
+type images struct {
```
JoelSpeed, line 13: You're shadowing the name of this struct in various places, perhaps call this `imageReference` and then it would be `images imgeReference` in other functions?

controllers/substitution.go
```diff
@@ -0,0 +1,44 @@
...
+	"strings"
+
+	v1 "k8s.io/api/apps/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	// commonControllerNamingSubstringPattern contains a substring which all CCM controller container names should have
+	// to be substituted with image from config
+	commonControllerNamingSubstringPattern = "controller"
```
JoelSpeed, line 13: I think the pattern here is good, might be better to use something explicit though, eg, if we have a list of predefined allowed values `cloud-controller-manager`, `node-controller-manager` etc that the manifests must obide by? WDYT?

controllers/substitution.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		d.Spec.Template.Spec.Containers[i].Image = config.ControllerImage
+	}
+}
+
+func fillConfigValues(config operatorConfig, templates []client.Object) []client.Object {
+	objects := make([]client.Object, len(templates))
+	for i, objectTemplate := range templates {
+		templateCopy := objectTemplate
```
JoelSpeed, line 30: Use the deepcopy method to get a real copy of the object
```suggestion
		templateCopy := objectTemplate.DeepCopy()
```

controllers/substitution.go
```diff
@@ -0,0 +1,44 @@
...
+
+		// Set namespaces for all object. Namespace on cluster-wide objects is stripped by API server and is not applied
```
JoelSpeed, line 32: Good comment

controllers/clusteroperator_controller.go
Danil-Grigorev: It will later use more then just *v1.Infrastructure. The one in parent scope will eventually go away, and we still operate on cache here. So two separate calls in my opinion is better option, as we won't change the method signature much with it.

controllers/clusteroperator_controller.go
Danil-Grigorev: Some leftovers from testing :sob: Very inconvenient when manifests from the platform I'm using are not merged: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/15

controllers/substitution.go
Danil-Grigorev: There are couple of reasons I can't do this. First DeepCopy on the client.Object returns runtime.Object which is not backwards compatible. We could probably use it on a typed object, but in this method there will be more then one type of objects. Also we probably don't need to have a copy, those templates are from beginning to an end meant to be changed in place and then applied.

controllers/substitution.go
Danil-Grigorev: We need some sort of CI task which will check that only those names are on containers and other subresources all the time.

controllers/config.go
Danil-Grigorev: I did the opposite :)

controllers/clusteroperator_controller.go
JoelSpeed: This feels like premature optimisation to me. We should implement things as they are today, because while you say things will change in the future, inevitably they will probably not or things will change. 

Changing the signature of a private function is a non-issue, I still stand by passing it in here as it makes this function more isolated and easier to understand. 

I'm going to revise my earlier suggestion though, we actually only need the `platform`, we should pass that in from the parent scope as we have it there readily available, makes this function much smaller

controllers/config.go
```diff
@@ -0,0 +1,57 @@
...
+	var i images
+	data, err := ioutil.ReadFile(filepath.Clean(filePath))
+	if err != nil {
+		return i, err
+	}
+
+	if err := json.Unmarshal(data, &i); err != nil {
+		return i, err
+	}
+	return i, nil
```
JoelSpeed, line 45: We should explicitly return an empty struct when there's an error, prevents side effects (eg partial unmarshal) from escaping this function
```suggestion
	data, err := ioutil.ReadFile(filepath.Clean(filePath))
	if err != nil {
		return images{}, err
	}

	var i images
	if err := json.Unmarshal(data, &i); err != nil {
		return images{}, err
	}
	return i, nil
```

controllers/substitution.go
```diff
@@ -0,0 +1,42 @@
...
+		objectTemplate.SetNamespace(config.ManagedNamespace)
+
+		dep, ok := objectTemplate.(*v1.Deployment)
+		if ok {
+			setDeploymentImages(config, dep)
+			// TODO: add cloud-config calculated hash to annotations to account for redeployment on content change
+		}
+
+		objects[i] = objectTemplate
+	}
```
JoelSpeed, line 40: This is ripe for having leaky side effects, we should have copies of the templates, not modify them in place directly
```suggestion
	for i, objectTemplate := range templates {
		obj, ok := objectTemplate.DeepCopy().(client.Object)
		if !ok {
		  // This should never happen. DeepCopy returns the same underlying
		  // type so it will always implement the interface.
		  panic("cloned object was not a client.Object")
		}
		// Set namespaces for all object. Namespace on cluster-wide objects is stripped by API server and is not applied
		obj.SetNamespace(config.ManagedNamespace)

		dep, ok := objectTemplate.(*v1.Deployment)
		if ok {
			setDeploymentImages(config, dep)
			// TODO: add cloud-config calculated hash to annotations to account for redeployment on content change
		}

		objects[i] = obj
	}
```

JoelSpeed (commented):
I think we need to clean this up a bit before we merge. I'm concerned about the side effects of some of the functions and also premature optimisations we are making. Let's keep the code clean now and worry about future changes later when we get to that point

controllers/clusteroperator_controller.go
```diff
@@ -180,6 +191,27 @@ func getResources(infra *configv1.Infrastructure) []client.Object {
...
+	if err != nil {
+		klog.Errorf("Unable to decode images file from location %s", r.ImagesFile, err)
+		return config, err
+	}
+
+	platform, err := getProviderFromInfrastructure(infra)
+	if err != nil {
+		klog.Errorf("Unable to decode images file from location %s", r.ImagesFile, err)
+		return config, err
+	}
```
JoelSpeed, line 207: We already have platform extracted in the parent scope, could just pass it in here rather than getting it again. Less error handling/complexity :)

JoelSpeed (commented):
In general, would be good to see some more unit testing for the methods we are adding here. There are lots of small parts (which is good!) but at the moment their functionality is only tested as a whole. If we introduce more scoped down unit testing this will help us to prove that individual functions are doing what we are intending them to do and in the long run will help identifying bugs (add test cases to specific functions and check the outputs)

Danil-Grigorev:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-cloud-controller-manager-operator/29/pull-ci-openshift-cluster-cloud-controller-manager-operator-master-e2e-aws-ccm/1386857755955957760 passed CI, will record it here as planing to address missed comment.

controllers/config.go
```diff
@@ -0,0 +1,57 @@
...
+// operatorConfig contains configuration values for templating resources
+type operatorConfig struct {
+	ManagedNamespace string
+	ControllerImage  string
+}
+
+func getProviderFromInfrastructure(infra *configv1.Infrastructure) (configv1.PlatformType, error) {
+	if infra == nil || infra.Status == (configv1.InfrastructureStatus{}) || infra.Status.PlatformStatus == nil {
+		return "", fmt.Errorf("platform status is not pupulated on infrastructure")
+	}
```
JoelSpeed, line 27: Nit: We don't need the middle part of this as it's not a pointer, deref will be fine
```suggestion
	if infra == nil || infra.Status.PlatformStatus == nil {
		return "", fmt.Errorf("platform status is not pupulated on infrastructure")
	}
```

controllers/config_test.go
```diff
@@ -0,0 +1,174 @@
...
+		name           string
+		infra          *configv1.Infrastructure
+		expectPlatform configv1.PlatformType
+		expectErr      bool
+	}{{
+		name:      "Passing empty infra causes error",
+		infra:     nil,
+		expectErr: true,
+	}, {
+		name: "No platorm type cauese error",
```
JoelSpeed, line 106: Typos
```suggestion
		name: "No platform type causes error",
```

JoelSpeed (commented):
Couple of nits but otherwise LGTM

controllers/clusteroperator_controller.go
Danil-Grigorev: Addressed this one. All should be fixed now, PTAL.

JoelSpeed (commented):
/approve

Thanks Danil, looks great!

controllers/config.go
```diff
@@ -0,0 +1,57 @@
...
+	"path/filepath"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// imagesReference allows build systems to inject imagesReference for CCCMO components
+type imagesReference struct {
+	CloudControllerManagerAWS       string `json:"cloudControllerManagerAWS"`
+	CloudControllerManagerOpenStack string `json:"cloudControllerManagerOpenStack"`
+}
```
elmiko, line 16: i'm just curious, will a cluster ever use multiple images from this reference?

controllers/config.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	if infra.Status.PlatformStatus.Type == "" {
+		return "", fmt.Errorf("no platform provider found on infrastructure")
+	}
+
+	return infra.Status.PlatformStatus.Type, nil
+}
+
+func getImagesFromJSONFile(filePath string) (imagesReference, error) {
+	data, err := ioutil.ReadFile(filepath.Clean(filePath))
```
elmiko, line 36: are these file paths loaded as volume mounts in the container?

controllers/substitution.go
```diff
@@ -0,0 +1,43 @@
+package controllers
+
+import (
+	v1 "k8s.io/api/apps/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// Names in that list are unique and will be substituted with an image from config
```
elmiko, line 9: is this an orphan comment?

controllers/substitution.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// cloudControlleManagerName is a name for default CCM controller container any provider may have
+	cloudControlleManagerName = "cloud-controller-manager"
```
elmiko, line 12: should this be `cloudControllerManagerName` ?

elmiko (commented):
this mostly makes sense to me, i have a couple questions though

controllers/substitution.go
Danil-Grigorev: No, I think this comment is in place. This is planned to be a list later. Things like `kube-rbac-proxy` and `node-cloud-controller-manager` should belong here as well.

controllers/config.go
Danil-Grigorev: Yes, they are loaded from a mounted config map introduced in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/31

controllers/config.go
Danil-Grigorev: Yes, this is possible, there is definitely going to be a `kube-rbac-proxy` image here once we will start investigating metrics support.

controllers/substitution.go
elmiko: ok, cool. i just wanted to make sure. i have a small suggestion here
```suggestion
	// Names in this list are unique and will be substituted with an image from config
```

controllers/clusteroperator_controller_test.go
```diff
@@ -307,32 +312,137 @@ var _ = Describe("Component sync controller", func() {
...
+
+func TestComposeConfig(t *testing.T) {
+	tc := []struct {
+		name          string
+		namespace     string
+		platform      configv1.PlatformType
+		imagesContent string
+		expectConfig  operatorConfig
+		expectError   bool
+	}{{
```
elmiko, line 376: i know it shouldn't matter, but i wonder if we should have a test with a duplicate entry just to make sure it doesn't break things?

elmiko (commented):
looking mostly good Danil, i have one small suggestion and a question about the tests

controllers/clusteroperator_controller_test.go
Danil-Grigorev: Added a test case for that in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/29/files#diff-57ee0125add50ed49739258385c78a28c3d1a738e949eb9c1210653facdcb658R43-R52 Although we are testing json unmarshal here, let's still have it for clarity.

elmiko (approved):
thanks Danil!
/lgtm

PR #31, opened 2021-03-29
Revert "Revert "Add images config map""

Danil-Grigorev:
Reverts openshift/cluster-cloud-controller-manager-operator#30

### DIFF
--- a/hack/example-images.json
+++ b/hack/example-images.json
@@ -1,4 +1,4 @@
 {
-    "cloudControllerManagerAWS": "registry.svc.ci.openshift.org/openshift:aws-cloud-controller-manager",
-    "cloudControllerManagerOpenStack": "registry.svc.ci.openshift.org/openshift:openstack-cloud-controller-manager"
+    "cloudControllerManagerAWS": "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
+    "cloudControllerManagerOpenStack": "registry.ci.openshift.org/openshift:openstack-cloud-controller-manager"
 }
--- a/hack/example-images.json
+++ b/hack/example-images.json
@@ -1,0 +1,4 @@
+{
+    "cloudControllerManagerAWS": "registry.svc.ci.openshift.org/openshift:aws-cloud-controller-manager",
+    "cloudControllerManagerOpenStack": "registry.svc.ci.openshift.org/openshift:openstack-cloud-controller-manager"
+}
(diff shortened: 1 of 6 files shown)


manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
```diff
@@ -49,3 +53,8 @@ spec:
         operator: "Exists"
         effect: "NoExecute"
         tolerationSeconds: 120
+      volumes:
+      - name: images
+        configMap:
+          defaultMode: 420
```
elmiko, line 59: is this the proper mode?

`chmod 420` produces `-r---w----` which doesn't look right to me.

hack/example-images.json
```diff
@@ -0,0 +1,4 @@
+{
+    "cloudControllerManagerAWS": "registry.svc.ci.openshift.org/openshift:aws-cloud-controller-manager",
+    "cloudControllerManagerOpenStack": "registry.svc.ci.openshift.org/openshift:openstack-cloud-controller-manager"
```
elmiko, line 3: we should probably fix these up before committing. no reason to use the old url

manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: cloud-controller-manager-images
+  namespace: openshift-cloud-controller-manager-operator
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+data:
+  images.json: >
+    {
+      "cloudControllerManagerAWS": "registry.svc.ci.openshift.org/openshift:aws-cloud-controller-manager",
+      "cloudControllerManagerOpenStack": "registry.svc.ci.openshift.org/openshift:openstack-cloud-controller-manager"
```
elmiko, line 13: same here

manifests/image-references
```diff
@@ -6,3 +6,11 @@ spec:
     from:
       kind: DockerImage
       name: registry.svc.ci.openshift.org/openshift:cluster-cloud-controller-manager-operator
+  - name: aws-cloud-controller-manager
+    from:
+      kind: DockerImage
+      name: registry.svc.ci.openshift.org/openshift:aws-cloud-controller-manager
```
elmiko, line 12: here too

elmiko (changes requested):
@Danil-Grigorev this looks fine to me, but i think we should update the image urls now since we are adding them

hack/example-images.json
Danil-Grigorev: Right, let me adjust that. CI is now passing

manifests/image-references
Danil-Grigorev: Fixed

elmiko (approved):
thanks Danil!
/approve

PR #32, opened 2021-04-07
Add management workload annotations

cynepco3hahue:
In support of the workload partitioning feature
(openshift/enhancements#703), we need to add
annotations to all management pods and namespaces so they can be
properly identified and assigned to segregated management cores on
clusters configured to do so.

Signed-off-by: Artyom Lukianov <alukiano@redhat.com>

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
@@ -15,6 +15,8 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         k8s-app: cloud-manager-operator
     spec:
--- a/manifests/0000_26_cloud-controller-manager-operator_00_namespace.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_00_namespace.yaml
@@ -5,6 +5,7 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
   labels:
     openshift.io/run-level: "0"
     openshift.io/cluster-monitoring: "true"
@@ -17,6 +18,7 @@ metadata:
     include.release.openshift.io/single-node-developer: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     openshift.io/node-selector: ""
+    workload.openshift.io/allowed: "management"
   labels:
     openshift.io/run-level: "0"
     openshift.io/cluster-monitoring: "true"
--- a/pkg/cloud/aws/assets/deployment.yaml
+++ b/pkg/cloud/aws/assets/deployment.yaml
@@ -12,6 +12,8 @@ spec:
       k8s-app: aws-cloud-controller-manager
   template:
     metadata:
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         k8s-app: aws-cloud-controller-manager
     spec:
--- a/pkg/cloud/openstack/assets/deployment.yaml
+++ b/pkg/cloud/openstack/assets/deployment.yaml
@@ -11,6 +11,8 @@ spec:
       app: openstack-cloud-controller-manager
   template:
     metadata:
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
         app: openstack-cloud-controller-manager
     spec:


JoelSpeed:
This operator creates more pods, cloud controller managers, I suspect we will need to annotate those as well right?

dhellmann:
> This operator creates more pods, cloud controller managers, I suspect we will need to annotate those as well right?

Yes, good point. We have a *bunch* of these PRs across the whole project, can you help out by pointing out in the code where the pod(s) are being created?

elmiko:
@dhellmann i _think_ this is the only one currently (openstack provider), https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/openstack/assets/deployment.yaml

JoelSpeed:
I'm wondering if, rather than adding to the deployment YAML, we should enforce this as code as part of the operator, that way we can guarantee they don't get missed on future providers

elmiko:
i think that makes some sense Joel, but i am concerned about creating some sort of bifurcated path for updating these manifests. if we are going to continue separating the kube-specific parts into yaml files that then get embedded into the binary, i think we should focus on keeping all the non-dynamic parts in the actual manifests.

JoelSpeed:
I'm happy with either, but if we are going down the route of keeping this in YAML, we need to have some sort of validation for our YAML so that we don't forget to add these when we add new platforms

I know we are going to have to integrate some logic to template the yamls in a way (to set build images), so we could add logic there for injecting annotations that are common across all platforms, either works

elmiko:
> I'm happy with either, but if we are going down the route of keeping this in YAML, we need to have some sort of validation for our YAML so that we don't forget to add these when we add new platforms
> 

big +1

> I know we are going to have to integrate some logic to template the yamls in a way (to set build images), so we could add logic there for injecting annotations that are common across all platforms, either works

honestly, my preference would be to see us move these templates entirely into code. but i realize this is a stylistic choice and my bias, still i think it's easier to manage all in one place or the other. so, i would like to see us keep the data centralized in one place, even if we add logic in another place.

dhellmann:
> > I'm happy with either, but if we are going down the route of keeping this in YAML, we need to have some sort of validation for our YAML so that we don't forget to add these when we add new platforms
> 
> big +1

We plan to add a CI test to ensure that all workloads created by core components have the annotation as part of https://issues.redhat.com/browse/CNF-1990 We hoped to put most of the annotations in place so we can minimize the amount of whack-a-mole for any we missed as part of that work.

> > I know we are going to have to integrate some logic to template the yamls in a way (to set build images), so we could add logic there for injecting annotations that are common across all platforms, either works
> 
> honestly, my preference would be to see us move these templates entirely into code. but i realize this is a stylistic choice and my bias, still i think it's easier to manage all in one place or the other. so, i would like to see us keep the data centralized in one place, even if we add logic in another place.
> 
> edit:
> maybe for now we just update the deployment manifest, and create an issue to revisit how we are using these manifests?

That would be ideal. We have a limited amount of time to land this stuff under our feature freeze extension. I would be happy to help brainstorm/review other solutions, if you want another pair of eyes when the time comes.

JoelSpeed:
Let's start off with it in the deployment yamls and we can review later, the file is https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/openstack/assets/deployment.yaml

Danil-Grigorev:
@cynepco3hahue AWS manifests were just merged, I think this PR could add annotations to https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/openstack/assets/deployment.yaml and https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/aws/assets/deployment.yaml

cynepco3hahue:
@Danil-Grigorev Thanks, done for both OpenStack and AWS.

manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
```diff
@@ -15,6 +15,8 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
```
elmiko, line 19: i was looking at the enhancement, and i did not see this annotation mentioned. is there anywhere i could read about it?

elmiko (commented):
this looks good to me, i just had a question about one of the annotations.

manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
cynepco3hahue: Sorry, we had some additional renaming, you can find the relevant information under the PR - https://github.com/openshift/enhancements/pull/739

elmiko: awesome, thank you!

elmiko (approved):
/lgtm

PR #33, opened 2021-04-16
Add anti-affinity settings to preserve HA for CCM pods

Danil-Grigorev:
Following proposal we need to run at least 2 replicas of CCM
controllers scheduled on different master nodes for HA: 
https://github.com/openshift/enhancements/pull/463/files#diff-3e0e2c48e70215076dfe36c13768a823ab7080d929d80292f37db2ef5a2121e8R123

Requires pieces from https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/15 to be merged first.

### DIFF
--- a/pkg/cloud/aws/assets/deployment.yaml
+++ b/pkg/cloud/aws/assets/deployment.yaml
@@ -48,3 +48,5 @@ spec:
         key: node.kubernetes.io/not-ready
         operator: Exists
         tolerationSeconds: 120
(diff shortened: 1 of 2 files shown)


JoelSpeed:
Changes in this look ok to me, will need to see the base PR merge first though

elmiko (approved):
i am not super familiar with the antiaffinity settings, but this generally looks good to me.
/lgtm

PR #35, opened 2021-04-26
WIP: put MCO overrides in manifests folder for testing

Danil-Grigorev:
This PR is for testing. Actual result (if it works in CI) will be applied from different Dockerfile and copied into manifests folder only on specific CI runs. This payload should never get into release.


Danil-Grigorev:
/close 

This PR didn't achieve the goal and a workaround was implemented in https://github.com/openshift/release/pull/18138

PR #38, opened 2021-05-07
Ensure priorityClass is set on AWS CCM

Danil-Grigorev:
Satisfy requirements of "[sig-arch] Managed cluster should ensure platform components have system-* priority class associated"

Full log: https://console.build01.ci.openshift.org/api/kubernetes/api/v1/namespaces/ci-op-6t1wq5k4/pods/e2e-aws-ccm-openshift-e2e-test/log?container=test

Increase weight on affinity settings to ensure the two pods are not scheduled on the same Node in edge cases.

Reasoning: 
https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical

OpenStack manifests have that same field. Cluster will not become healthy during install if those pods are not healthy. Not `system-node-critical` as provisioned Nodes could operate without these pods running.

### DIFF
--- a/pkg/cloud/aws/assets/deployment.yaml
+++ b/pkg/cloud/aws/assets/deployment.yaml
@@ -17,6 +17,7 @@ spec:
       labels:
         k8s-app: aws-cloud-controller-manager
     spec:
+      priorityClassName: system-cluster-critical
       containers:
       - args:
         - --cloud-provider=aws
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
Could we add to the description why this is `system-cluster-critical` as opposed to the other priority classes please?

Danil-Grigorev:
Openstack manifests are `system-cluster-critical`. It could be `system-node-critical` but it shouldn't be. Cluster during initialization is not healthy if these pods were not healthy. @JoelSpeed

Danil-Grigorev:
https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical

PR #40, opened 2021-05-11
Improve operator status management

Danil-Grigorev:
Operator correctly reports degraded and available conditions during reconcile.

- Distribute methods into packages where they belong.
- Add events on resource updates, cluster operator changes.
- Report progressing condition on cluster upgrade or resources change.

### DIFF
--- a/controllers/clusteroperator_controller.go
+++ b/controllers/clusteroperator_controller.go
@@ -23,6 +23,8 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/openstack"
+	"github.com/openshift/library-go/pkg/cloudprovider"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
@@ -35,8 +37,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
-
-	"github.com/openshift/library-go/pkg/cloudprovider"
 )
 
 const (
@@ -143,13 +143,6 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
-	// TODO: Set progressing condition only once we've evaluated there is a change in managed resources
-	// // Sync provider operands
-	// if err := r.statusProgressing(ctx); err != nil {
-	// 	klog.Errorf("Error syncing ClusterOperatorStatus: %v", err)
-	// 	return ctrl.Result{}, fmt.Errorf("error syncing ClusterOperatorStatus: %v", err)
-	// }
-
 	if err := r.sync(ctx, config); err != nil {
 		klog.Errorf("Unable to sync operands: %s", err)
 		if err := r.setStatusDegraded(ctx, err); err != nil {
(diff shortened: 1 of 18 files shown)


controllers/status.go
```diff
@@ -0,0 +1,199 @@
...
+
+	desiredVersions := []configv1.OperandVersion{{Name: operatorVersionKey, Version: getReleaseVersion()}}
+	currentVersions := co.Status.Versions
+	var isProgressing configv1.ConditionStatus
+
+	var message, reason string
+	if !reflect.DeepEqual(desiredVersions, currentVersions) {
+		klog.V(2).Info("Syncing status: progressing")
+		message = fmt.Sprintf("Progressing towards %s", printOperandVersions(desiredVersions))
+		// r.eventRecorder.Eventf(co, v1.EventTypeNormal, "Status upgrade", message)
```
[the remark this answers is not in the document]
Danil-Grigorev: Was not implemented yet, now it is integrated.

controllers/clusteroperator_controller.go
```diff
@@ -144,38 +160,64 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
...
+		resourceUpdated := resource.DeepCopyObject().(client.Object)
+		if err := r.Patch(ctx, resourceUpdated, client.Apply, client.ForceOwnership, client.FieldOwner(clusterOperatorName)); err != nil {
+			klog.Errorf("Unable to apply object %s '%s': %+v", resource.GetObjectKind().GroupVersionKind(), resource.GetName(), err)
+			r.Recorder.Event(resourceExisting, corev1.EventTypeWarning, "Update failed", err.Error())
+			return false, err
+		}
+		klog.V(2).Infof("Applied %s %q successfully", resource.GetObjectKind().GroupVersionKind(), client.ObjectKeyFromObject(resource))
+
+		if resourceExisting.GetGeneration() != resourceUpdated.GetGeneration() {
+			klog.Infof("Resource %s %q generation increased, updating...", resource.GetObjectKind().GroupVersionKind(), client.ObjectKeyFromObject(resource))
```
lobziik, line 204: Maybe change this message to smth like ""Resource %s %q updated.""? As far as I understand resource already updated there.

lobziik: Or specify somehow what is gonna update here.

controllers/config_test.go
```diff
@@ -181,3 +182,89 @@ func TestGetProviderControllerFromImages(t *testing.T) {
 		})
 	}
 }
+
+func TestComposeConfig(t *testing.T) {
```
lobziik, line 186: Can we use ginkgo for this test?

controllers/status.go
```diff
@@ -0,0 +1,185 @@
...
+const (
+	clusterOperatorName        = "cloud-controller-manager"
+	operatorVersionKey         = "operator"
+	defaultManagementNamespace = "openshift-cloud-controller-manager-operator"
+)
+
+// statusDegraded sets the Degraded condition to True, with the given reason and
+// message, and sets the upgradeable condition.  It does not modify any existing
+// Available or Progressing conditions.
+func (r *CloudOperatorReconciler) statusDegraded(ctx context.Context, reconcileErr error) error {
```
lobziik, line 36: suggest to add verb in this functions names, e.g `setStatusDegraded`. WDYT?

controllers/config_test.go
Danil-Grigorev: I don't think it is needed. We use ginkgo because of the comfort working with real API server in C-R, and this is just a unit test for method which does no API requests.

controllers/status.go
Danil-Grigorev: We call it `statusDegraded` in MAO and in cluster-machine-approver. cc @JoelSpeed, do you think it is ok to rename these methods?

controllers/clusteroperator_controller.go
```diff
@@ -144,38 +160,64 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
...
+
+		resourceUpdated := resource.DeepCopyObject().(client.Object)
+		if err := r.Patch(ctx, resourceUpdated, client.Apply, client.ForceOwnership, client.FieldOwner(clusterOperatorName)); err != nil {
+			klog.Errorf("Unable to apply object %s '%s': %+v", resource.GetObjectKind().GroupVersionKind(), resource.GetName(), err)
+			r.Recorder.Event(resourceExisting, corev1.EventTypeWarning, "Update failed", err.Error())
+			return false, err
+		}
+		klog.V(2).Infof("Applied %s %q successfully", resource.GetObjectKind().GroupVersionKind(), client.ObjectKeyFromObject(resource))
+
+		if resourceExisting.GetGeneration() != resourceUpdated.GetGeneration() {
```
JoelSpeed, line 203: How have you tested that this works? Does it stabilise?

controllers/clusteroperator_controller_test.go
```diff
@@ -365,84 +365,104 @@ var _ = Describe("Component sync controller", func() {
...
+		resources = append(resources, cloud.GetAWSResources()...)
+
+		updated, err := reconciler.applyResources(context.TODO(), resources)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(updated).To(BeTrue())
+		Eventually(recorder.Events).Should(Receive(ContainSubstring("Resource was successfully updated")))
+	})
+
+	It("Expect update when deployment generation have changed", func() {
+		dep := &appsv1.Deployment{}
```
JoelSpeed, line 407: Nit: To make it clearer that we are creating a reference to the deployment resource, I would init this in a different way
```suggestion
		var dep *appsv1.Deployment
```

With the current way it's like we are creating an empty one then starting to fill it out

controllers/status.go
JoelSpeed: I'm happy to rename so long as they are consistent within a single operator. IMO `setStatusFoo` is better than `statusFoo` anyway 😄

controllers/status.go
```diff
@@ -0,0 +1,185 @@
...
+		klog.Errorf("Failed to get or create Cluster Operator: %v", err)
+		return err
+	}
+
+	desiredVersions := []configv1.OperandVersion{{Name: operatorVersionKey, Version: r.ReleaseVersion}}
+	currentVersions := co.Status.Versions
+	if err != nil {
+		klog.Errorf("Error getting current versions: %v", err)
+		return err
+	}
```
JoelSpeed, line 48: This seems... random?

controllers/status.go
```diff
@@ -0,0 +1,185 @@
...
+
+	conds := []configv1.ClusterOperatorStatusCondition{
+		newClusterOperatorStatusCondition(configv1.OperatorDegraded, configv1.ConditionTrue,
+			ReasonSyncFailed, message),
+		newClusterOperatorStatusCondition(configv1.OperatorUpgradeable, configv1.ConditionFalse, ReasonAsExpected, ""),
+	}
+
+	r.Recorder.Eventf(co, corev1.EventTypeWarning, "Status degraded", reconcileErr.Error())
+	klog.V(2).Info("Syncing status: degraded")
```
JoelSpeed, line 64: Maybe include the message? Or error?

controllers/status.go
```diff
@@ -0,0 +1,185 @@
...
+		return err
+	}
+
+	desiredVersions := []configv1.OperandVersion{{Name: operatorVersionKey, Version: r.ReleaseVersion}}
+	currentVersions := co.Status.Versions
+
+	var message, reason string
+	if !reflect.DeepEqual(desiredVersions, currentVersions) {
+		klog.V(2).Info("Syncing status: progressing")
+		message = fmt.Sprintf("Progressing towards %s", printOperandVersions(desiredVersions))
```
JoelSpeed, line 84: Maybe include the message in the log?

controllers/config_test.go
JoelSpeed: Could use gomega still, definitely prefer that personally, rather than `t.Fatal`, but up to you :)

Danil-Grigorev: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/41 is refactoring this in preparation for bootstrap support in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/42. It is using `assert` which is very convenient. I still think gomega should be used in cases where there is a need to operator with API server, and this is not the case. `DescribeTable` is very unintuitive, and with use of assert simplified this tests enough that AfterEach and BeforeEach does not add value.

controllers/status.go
Danil-Grigorev: Changed.

controllers/clusteroperator_controller.go
Danil-Grigorev: Yes, this does stabilise. We are running real api server in testenv, so our e2e is covering it in these cases:
- https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/40/files#diff-c896f20f6b6cdd8a7745c1a4b0dbc4ce7c0ff9198d3e83929bcef8ed9eec8092R428
- https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/40/files#diff-c896f20f6b6cdd8a7745c1a4b0dbc4ce7c0ff9198d3e83929bcef8ed9eec8092R444-R455

PR #41, opened 2021-05-17
Refactor config and substitution

Danil-Grigorev:
This change separates substitution, config and cloud methods in their own packages in 
preparation to support bootstrap implementation folowing in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/42.

It refactors testing and adding full test coverage for moved methods. This is required for later bootstrap manifests implementation.

Commits with actual changes https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/41/files/600422dafca2e1c6cc63068a0061fc52f034366f..HEAD

Follow up on https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/40


pkg/cloud/openstack/openstack.go
```diff
@@ -1,46 +0,0 @@
-package openstack
```
lobziik, line 1: Honestly I like this approach with separate directory more then proposed one. Was there some discussion and agreement around this topic?

pkg/config/config_test.go
```diff
@@ -107,19 +102,19 @@ func TestGetProviderFromInfrastructure(t *testing.T) {
 		name           string
 		infra          *configv1.Infrastructure
 		expectPlatform configv1.PlatformType
-		expectErr      bool
+		expectErr      string
 	}{{
 		name:      "Passing empty infra causes error",
 		infra:     nil,
-		expectErr: true,
+		expectErr: "platform status is not pupulated on infrastructure",
```
lobziik, line 109: nit: typo

pkg/cloud/aws/aws.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	return resources
+}
+
+//go:embed assets/*
+var awsFS embed.FS
+var awsSources = []common.ObjectSource{
+	{Object: &appsv1.Deployment{}, Path: "assets/deployment.yaml"},
+}
```
elmiko, line 33: i think it would be nice to group the global variables at the top of the file. imo, it makes it easier to find when looking at the source. not a blocker, just a suggestion

pkg/cloud/openstack/openstack.go
```diff
@@ -1,46 +1,35 @@
...
+	return resources
+}
 
-	return ret
+//go:embed assets/*
+var OpenStackFS embed.FS
+var OpenStackSources = []common.ObjectSource{
+	{Object: &v1.ConfigMap{}, Path: "assets/config.yaml"},
+	{Object: &appsv1.Deployment{}, Path: "assets/deployment.yaml"},
 }
```
elmiko, line 35: similar suggestion here about the global variables

pkg/cloud/aws/aws.go
elmiko: also, i notice that the openstack variables are exported, should these be `AwsFS` and `AwsSources`?

pkg/config/config.go
```diff
@@ -16,16 +16,16 @@ type imagesReference struct {
 	CloudControllerManagerOpenStack string `json:"cloudControllerManagerOpenStack"`
 }
 
-// operatorConfig contains configuration values for templating resources
-type operatorConfig struct {
+// Config contains configuration values for templating resources
+type Config struct {
```
elmiko, line 20: just curious, will we have any other config structs later? should we keep this named `OperatorConfig` ?

elmiko (commented):
nice work Danil, looks generally good to me but i have a couple style suggestions and a few questions.

pkg/config/config.go
Danil-Grigorev: Let's name it like this. Render will also create OperatorConfig, sounds ok.

pkg/cloud/aws/aws.go
Danil-Grigorev: Rather the opposite, we only need to expose the `GetResources()` method.

pkg/cloud/cloud.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func GetResources(platform configv1.PlatformType) []client.Object {
+	switch platform {
+	case configv1.AWSPlatformType:
+		return aws.GetResources()
+	case configv1.OpenStackPlatformType:
+		return openstack.GetResources()
+	default:
+		klog.Warning("No recognized cloud provider platform found in infrastructure")
```
JoelSpeed, line 18: Might be good to include the platform in the warning too? So we can see what it thinks it's seeing. (Also I reworded a little to improve readability IMO, open for discussion there)
```suggestion
		klog.Warningf("Unrecognized platform type %q found in infrastructure", platform)
```

pkg/cloud/common/sources.go
```diff
@@ -0,0 +1,36 @@
...
+	ret := []client.Object{}
+	for _, source := range sources {
+		data, err := f.ReadFile(source.Path)
+		if err != nil {
+			klog.Errorf("Cannot read embedded resource %v: %v", source.Path, err)
+			return nil, err
+		}
+
+		dec := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), 1000)
+		if err := dec.Decode(source.Object); err != nil {
```
JoelSpeed, line 27: Should we be copying the source object, otherwise this modifies the input to the function which could be a nasty future side effect. The alternative is to make this a `ReadResourcesInto` style method so it's expected that it reads into the `sources` input

JoelSpeed (commented):
This looks good, good separation of concerns into the different areas.

In general it would be good to see comments on public structs and public methods so that we can have a godoc style reference to what is going on here.

Also, I would like to see unit tests for the provider packages so that we know from the unit tests that they don't panic on init.

elmiko (approved):
thanks Danil!
/lgtm

PR #42, opened 2021-05-17
Implement bootstrap aws

Danil-Grigorev:
Following Pod will serve Node initialization during cluster bootstrap in AWS.
- Running only ` --controllers=cloud-node` as `cloud-node` controller contains node initialization parts we require in bootstrap.
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: aws-cloud-controller-manager
  namespace: kube-system
spec:
  priorityClassName: system-cluster-critical
  containers:
  - args:
    - --cloud-provider=aws
    - --use-service-account-credentials=false
    - --controllers=cloud-node # run only cloud-node controller required to bootstrap master nodes
    - --kubeconfig=/etc/kubernetes/secrets/kubeconfig
    - --leader-elect=false
    - -v=2
    image: gcr.io/k8s-staging-provider-aws/cloud-controller-manager:v1.19.0-alpha.1
    imagePullPolicy: IfNotPresent
    volumeMounts:
    - mountPath: /etc/kubernetes/secrets
      name: secrets
      readOnly: true
  hostNetwork: true
  volumes:
  - hostPath:
      path: /etc/kubernetes/bootstrap-secrets
    name: secrets
```

### DIFF
--- a/pkg/cloud/aws/bootstrap/pod.yaml
+++ b/pkg/cloud/aws/bootstrap/pod.yaml
@@ -1,0 +1,27 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: aws-cloud-controller-manager
+  namespace: kube-system
+spec:
+  priorityClassName: system-cluster-critical
+  containers:
+  - args:
+    - --cloud-provider=aws
+    - --use-service-account-credentials=false
+    - --controllers=cloud-node # run only cloud-node controller required to bootstrap master nodes
+    - --kubeconfig=/etc/kubernetes/secrets/kubeconfig
+    - --leader-elect=false
+    - -v=2
+    image: gcr.io/k8s-staging-provider-aws/cloud-controller-manager:v1.19.0-alpha.1
+    imagePullPolicy: IfNotPresent
+    name: cloud-controller-manager
+    volumeMounts:
+    - mountPath: /etc/kubernetes/secrets
+      name: secrets
+      readOnly: true
+  hostNetwork: true
+  volumes:
+  - hostPath:
+      path: /etc/kubernetes/bootstrap-secrets
+    name: secrets
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -71,3 +71,66 @@ func TestGetResources(t *testing.T) {
 		})
 	}
 }
+
+func TestGetBootstrapResources(t *testing.T) {
+	tc := []struct {
+		name     string
+		platform configv1.PlatformType
+		expected []client.Object
+	}{{
+		name:     "AWS resources returned as expected",
+		platform: configv1.AWSPlatformType,
+		expected: aws.GetBootstrapResources(),
+	}, {
+		name:     "OpenStack resources are empty, as the platform is not yet supported",
+		platform: configv1.OpenStackPlatformType,
+	}, {
+		name:     "GCP resources are empty, as the platform is not yet supported",
+		platform: configv1.GCPPlatformType,
+	}, {
+		name:     "Azure resources are empty, as the platform is not yet supported",
+		platform: configv1.AzurePlatformType,
+	}, {
+		name:     "VSphere resources are empty, as the platform is not yet supported",
+		platform: configv1.VSpherePlatformType,
+	}, {
+		name:     "OVirt resources are empty, as the platform is not yet supported",
+		platform: configv1.OvirtPlatformType,
+	}, {
+		name:     "IBMCloud resources are empty, as the platform is not yet supported",
+		platform: configv1.IBMCloudPlatformType,
+	}, {
+		name:     "Libvirt resources are empty",
+		platform: configv1.LibvirtPlatformType,
+	}, {
+		name:     "Kubevirt resources are empty",
+		platform: configv1.KubevirtPlatformType,
+	}, {
+		name:     "BareMetal resources are empty",
+		platform: configv1.BareMetalPlatformType,
+	}, {
+		name:     "None platform resources are empty",
+		platform: configv1.NonePlatformType,
+	}}
+
+	for _, tc := range tc {
+		t.Run(tc.name, func(t *testing.T) {
+			resources := GetBootstrapResources(tc.platform)
+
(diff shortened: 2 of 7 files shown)


pkg/cloud/aws/bootstrap/pod.yaml
```diff
@@ -0,0 +1,33 @@
...
+  namespace: kube-system
+spec:
+  priorityClassName: system-cluster-critical
+  containers:
+  - args:
+    - --cloud-provider=aws
+    - --use-service-account-credentials=false
+    - --controllers=*,-service,-route # disable service and route controllers in bootstrap
+    - --kubeconfig=/etc/kubernetes/secrets/kubeconfig
+    - --leader-elect=false
```
lobziik, line 14: Might it cause conflicts when ccm from actual masters comes in?

pkg/cloud/aws/bootstrap/pod.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - -v=2
+    - "2> /var/log/bootstrap-control-plane/cloud-controller-manager.log"
+    image: gcr.io/k8s-staging-provider-aws/cloud-controller-manager:v1.19.0-alpha.1
```
lobziik, line 17: Guess we should use ocp image there. Don't see substitution call in aws.GetBootstrapResources...

lobziik:
Overall approach make sense to me, but Pod resource is quite questionable ATM...

pkg/cloud/aws/bootstrap/pod.yaml
Danil-Grigorev: Saw some failures with custom resource used for leader election here, so this feature is disabled. No it would not, additionally CCM could run in parallel with little or no issues. Once all masters are ready this bootstrap node would go away, and so additional CCM replica.

pkg/cloud/aws/bootstrap/pod.yaml
Danil-Grigorev: It would be there, if we add pod substitution pattern. We currently substitute only Deployments. The resource goes through substitution cycle in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/45/files#diff-0f4e2af2ebf3a82b1841788f7aad091a1e46fdee85a28a07c9747eb514dbd83fR68

pkg/cloud/aws/bootstrap/pod.yaml
```diff
@@ -0,0 +1,33 @@
...
+  priorityClassName: system-cluster-critical
+  containers:
+  - args:
+    - --cloud-provider=aws
+    - --use-service-account-credentials=false
+    - --controllers=*,-service,-route # disable service and route controllers in bootstrap
+    - --kubeconfig=/etc/kubernetes/secrets/kubeconfig
+    - --leader-elect=false
+    - -v=2
+    - "2> /var/log/bootstrap-control-plane/cloud-controller-manager.log"
```
Danil-Grigorev, line 16: Unfortunately the approach to store logs on bootstrap machine does not work. Neither this, nor `--log-file` option works in the CCM. I assume `--log-file` is an issue in cloud-provider. I would leave this line in, but we would need to address this issue to simplify CCM debug in bootstrap node. cc @JoelSpeed @lobziik

pkg/cloud/cloud_test.go
```diff
@@ -71,3 +71,66 @@ func TestGetResources(t *testing.T) {
...
+	}, {
+		name:     "OpenStack resources are empty, as the platform is not yet supported",
+		platform: configv1.OpenStackPlatformType,
+	}, {
+		name:     "GCP resources are empty, as the platform is not yet supported",
+		platform: configv1.GCPPlatformType,
+	}, {
+		name:     "Azure resources are empty, as the platform is not yet supported",
+		platform: configv1.AzurePlatformType,
+	}, {
```
elmiko, line 93: should we have an entry for azure stack hub as well?

pkg/cloud/common/sources.go
```diff
@@ -8,11 +8,17 @@ import (
...
 )
 
+// ObjectSource is a holder for the path of the object to read from
+// and typed Object, where the resource content would be unserialized
 type ObjectSource struct {
 	Object client.Object
 	Path   string
 }
 
+// ReadResources is responsible for reading resources from ebmed path in the binary.
```
elmiko, line 18:
```suggestion
// ReadResources is responsible for reading resources from embed path in the binary.
```

pkg/cloud/common/sources.go
```diff
@@ -8,11 +8,17 @@ import (
...
+// This method is for internal use only, and a <provider> should use this method
+// only in init() and later provide a copy of aquired objects. See /pkg/cloud for
```
elmiko, line 20: i'm curious why this needs to be run in an init() method?

elmiko (commented):
this generally makes sense to me, i just have a couple questions

pkg/cloud/common/sources.go
Danil-Grigorev: `init()` will run only once, and a failure there will prevent the binary from starting, which will be immediately displayed in CI tests and in attempts to run CCCMO locally. As we don't require any external dependency, and all our manifests are baked in the binary with embed, an ideal scenario would be that this read would not create any errors. Unfortunately we can't rely on that, so we need to make sure that the implementation would return the error if there is an issue with this static manifest. There could be no changes to this manifest after it, so the moment it being added to the binary is the only one we need to watch. Ideal scenario for init.

elmiko: perfect, thank you for the explanation =)

pkg/cloud/cloud_test.go
Danil-Grigorev: There is no platform for Azure Stack Hub :( We will have to implement additional logic to separate Azure and Azure Stack Hub similar to https://issues.redhat.com/browse/OCPCLOUD-1187 (well, starting from that, and then read the full infra status in operator as well to get it right).

elmiko: ack, thanks!

elmiko (approved):
/approve

PR #45, opened 2021-05-18
Implement render binary for use in installer

Danil-Grigorev:
For review: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/45/files/7ac280f0c47a919d6a0b4ef914bde93ec671889f..0ee19ba06b9d7914f1fa74d96eb9c5bfc9d6eb32 

Based on https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/42
Used in https://github.com/openshift/installer/pull/4947
```
Starts Cluster Cloud Controller Manager in render mode

Usage:
  run [flags]

Flags:
      --cluster-infrastructure-file string   Input path for the cluster infrastructure file.
      --dest-dir string                      The destination dir where CCCMO writes the generated static pods for CCM.
  -h, --help                                 help for run
      --images-file string                   Input path for the images config map file.

```

To run locally:
```
make build
./bin/render run --images-file=manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml --dest-dir=cccmo-render --cluster-infrastructure-file=infra.yaml
```


pkg/render/render_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+			// Assert all files were written to bootstrap dir
+			files, err := ioutil.ReadDir(path.Join(destination, bootstrapPrefix))
+			assert.NoError(t, err)
+			assert.Len(t, files, len(tc.expectObjects))
+		})
+	}
+}
+
+func TestWriteAssets(t *testing.T) {
```
lobziik, line 234: Would be nice to check written file content as well.

cmd/render/main.go
```diff
@@ -0,0 +1,94 @@
...
+	if st.Size() <= 0 {
+		return fmt.Errorf("%q is empty", path)
+	}
+
+	return nil
+}
+
+func main() {
+	if err := renderCmd.Execute(); err != nil {
+		klog.Exitf("Error executing render: %v", err)
```
elmiko, line 92: i think it would be nice to print the individual errors here, when i run the render command i see them grouped together like this:
```
Error: [error reading --cluster-infrastructure-file: stat : no such file or directory, error reading --images-file: stat : no such file or directory]```

**Review by elmiko — COMMENTED** [review]
this generally makes sense to me, i tried it out locally and everything worked as expected. i have one small suggestion and i agree with @lobziik about adding a test on the output file contents.

File: pkg/render/render_test.go (hunk 1 above)
Comment (Danil-Grigorev) [reply pkg/render/render_test.go:235]: Done.

File: pkg/render/render_test.go [hunk 3]
```diff
@@ -0,0 +1,309 @@
...
+
+			names := []string{}
+			for _, file := range files {
+				names = append(names, file.Name())
+			}
+			for _, res := range tc.objects {
+				filename := fmt.Sprintf("%s-%s.yaml", res.GetName(), strings.ToLower(res.GetObjectKind().GroupVersionKind().Kind))
+				found := assert.Contains(t, names, filename)
+
+				if found {
```
lobziik, line 292: nit: I think it might make sense to split this case into two. This if condition block looks bit strange for me.
Up to you.

Fedosin (approved):
It works well for me, so I don't want to hold it. If there are any issues we can fix it later.

/approve

PR #46, opened 2021-05-20
Add azure images

lobziik:
Add azure images to images configmap and image-reference file

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
@@ -10,5 +10,7 @@ data:
   images.json: >
     {
       "cloudControllerManagerAWS": "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
+      "cloudControllerManagerAzure": "registry.ci.openshift.org/openshift:azure-cloud-controller-manager",
+      "cloudNodeManagerAzure": "registry.ci.openshift.org/openshift:azure-cloud-node-manager",
       "cloudControllerManagerOpenStack": "registry.ci.openshift.org/openshift:openstack-cloud-controller-manager"
     }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -13,6 +13,8 @@ import (
 // imagesReference allows build systems to inject imagesReference for CCCMO components
 type imagesReference struct {
 	CloudControllerManagerAWS       string `json:"cloudControllerManagerAWS"`
+	CloudControllerManagerAzure     string `json:"cloudControllerManagerAzure"`
+	CloudNodeManagerAzure           string `json:"cloudNodeManagerAzure"`
 	CloudControllerManagerOpenStack string `json:"cloudControllerManagerOpenStack"`
 }
 
@@ -20,6 +22,7 @@ type imagesReference struct {
 type OperatorConfig struct {
 	ManagedNamespace string
 	ControllerImage  string
+	CloudNodeImage   string
 	Platform         configv1.PlatformType
 }
 
@@ -53,6 +56,17 @@ func getCloudControllerManagerFromImages(platform configv1.PlatformType, images 
 		return images.CloudControllerManagerAWS
 	case configv1.OpenStackPlatformType:
 		return images.CloudControllerManagerOpenStack
+	case configv1.AzurePlatformType:
+		return images.CloudControllerManagerAzure
+	default:
+		return ""
+	}
+}
+
+func getCloudNodeManagerFromImages(platform configv1.PlatformType, images imagesReference) string {
+	switch platform {
+	case configv1.AzurePlatformType:
+		return images.CloudNodeManagerAzure
 	default:
 		return ""
 	}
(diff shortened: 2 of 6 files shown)


JoelSpeed:
I spoke with @Danil-Grigorev yesterday, would it be better to have a single image for Azure but with the two binaries in it? This would make the image substitution easier IIUC. This is something we have done with other things like spot termination handlers as part of MAO

lobziik:
Agree, it would be easier. I'm mostly doubting about delivering ccm binary to worker nodes... Don't think that it might cause big issues but would prefer to keep it separate. Also building windows binary might require special treatment, but I'm not sure there.

For additional context: 
upstream manifests for node-manager: https://github.com/openshift/cloud-provider-azure/blob/master/examples/out-of-tree/cloud-node-manager.yaml

lobziik:
This might be another thing https://github.com/kubernetes-sigs/cloud-provider-azure/pull/580 - we might want to use different images for different types of nodes

manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
```diff
@@ -10,5 +10,7 @@ data:
   images.json: >
     {
       "cloudControllerManagerAWS": "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
+      "cloudControllerManagerAzure": "registry.ci.openshift.org/openshift:azure-cloud-controller-manager",
+      "cloudNodeManagerAzure": "registry.ci.openshift.org/openshift:azure-cloud-node-manager",
```
Danil-Grigorev, line 14: These fields should be supported in `config.go`

Danil-Grigorev:
> Agree, it would be easier. I'm mostly doubting about delivering ccm binary to worker nodes... Don't think that it might cause big issues but would prefer to keep it separate. Also building windows binary in future might require special treatment, but I'm not sure there.
> 
> For additional context:
> upstream manifests for node-manager: https://github.com/openshift/cloud-provider-azure/blob/master/examples/out-of-tree/cloud-node-manager.yaml

https://coreos.slack.com/archives/CM4ERHBJS/p1620994660139000 сс @lobziik We are not required to run azure for windows nodes yet.

pkg/config/config.go
```diff
@@ -53,6 +56,17 @@ func getCloudControllerManagerFromImages(platform configv1.PlatformType, images
...
 	case configv1.OpenStackPlatformType:
 		return images.CloudControllerManagerOpenStack
+	case configv1.AzurePlatformType:
+		return images.CloudControllerManagerAzure
+	default:
+		return ""
+	}
+}
+
+func getCloudNodeManagerFromImages(platform configv1.PlatformType, images imagesReference) string {
```
Danil-Grigorev, line 66: This method should be tested, in case we would add new providers following similar pattern.

pkg/substitution/substitution.go
```diff
@@ -20,11 +21,22 @@ func setDeploymentImages(config config.OperatorConfig, d *v1.Deployment) {
...
-		klog.Infof("Substituting %q: %s", container.Name, config.ControllerImage)
+		klog.Infof("Substituting %q in %q with %s", container.Name, d.Kind, config.ControllerImage)
 		d.Spec.Template.Spec.Containers[i].Image = config.ControllerImage
 	}
 }
 
+func setDaemonSetImage(config config.OperatorConfig, d *v1.DaemonSet) {
+	for i, container := range d.Spec.Template.Spec.Containers {
+		if container.Name != cloudNodeManagerName {
+			continue
```
Danil-Grigorev, line 32: This branch is not covered with tests.

pkg/substitution/substitution.go
```diff
@@ -33,12 +45,12 @@ func FillConfigValues(config config.OperatorConfig, templates []client.Object) [
 		// Set namespaces for all object. Namespace on cluster-wide objects is stripped by API server and is not applied
 		templateCopy.SetNamespace(config.ManagedNamespace)
 
-		dep, ok := templateCopy.(*v1.Deployment)
-		if ok {
+		switch dep := templateCopy.(type) {
```
Danil-Grigorev, line 48: Please pick another name, as we are already working with Deployments and DaemonSets.

pkg/config/config.go
```diff
@@ -13,13 +13,16 @@ import (
...
+	CloudControllerManagerAzure     string `json:"cloudControllerManagerAzure"`
+	CloudNodeManagerAzure           string `json:"cloudNodeManagerAzure"`
 	CloudControllerManagerOpenStack string `json:"cloudControllerManagerOpenStack"`
 }
 
 // OperatorConfig contains configuration values for templating resources
 type OperatorConfig struct {
 	ManagedNamespace string
 	ControllerImage  string
+	CloudNodeImage   string
```
Danil-Grigorev, line 25: I think this one will be better named `CloudNodeControllerImage`

lobziik:
Suggest to rewiew and merge it for now as well as https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/45 

Gonna introduce finished refactoring with interface (https://github.com/lobziik/cluster-cloud-controller-manager-operator/pull/1) as follow up, would be way more convenient for development and review.

Danil-Grigorev (approved):
/approve0

elmiko (approved):
/lgtm

PR #49, opened 2021-05-21
Tolerate uninitialized taint in operator pods

Danil-Grigorev:
Operator has to be able to start on master Nodes, when all of them are
marked as uninitialized. Additionally not-ready taint has NoSchedule
effect on Nodes when the deadlock happens.

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
@@ -44,6 +44,9 @@ spec:
         node-role.kubernetes.io/master: ""
       restartPolicy: Always
       tolerations:
+      - key: node.cloudprovider.kubernetes.io/uninitialized
+        effect: NoSchedule
+        value: "true"
       - key: "node-role.kubernetes.io/master"
         operator: "Exists"
         effect: "NoSchedule"
--- a/pkg/cloud/aws/assets/deployment.yaml
+++ b/pkg/cloud/aws/assets/deployment.yaml
@@ -57,3 +57,4 @@ spec:
         tolerationSeconds: 120
       - effect: NoSchedule
         key: node.cloudprovider.kubernetes.io/uninitialized
+        value: "true"


manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
```diff
@@ -44,6 +44,11 @@ spec:
         node-role.kubernetes.io/master: ""
       restartPolicy: Always
       tolerations:
+      - key: node.cloudprovider.kubernetes.io/uninitialized
+        effect: NoSchedule
+        value: "true"
+      - key: node.kubernetes.io/not-ready
+        effect: NoSchedule
```
JoelSpeed, line 51: Is this documented somewhere? I wouldn't have expected this to be required from the enhancement. Would be good to find official word on this to make sure this is the effect of the external provider and not some pot luck or other change we aren't aware of

JoelSpeed:
The operator and the operands will be required to tolerate the uninitialised taint, what is the plan for covering the operands?

elmiko (commented):
this looks good to me, @Danil-Grigorev anything we can do about this https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/49/files#r637904935 ?

manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
JoelSpeed: The node is being marked as not ready because the CNI driver is not yet running. We need to make sure the CNI driver tolerates the unitialized taint. Then networking will come up as expected and we won't need to tolerate this not-ready taint on any of our pods.

This should be considered a blocker for this PR IMO

elmiko: i think the openshift-sdn component is already set to tolerate this, see https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/openshift-sdn/controller.yaml#L60

Danil-Grigorev: I removed the not-ready taint, should be ready to go.

JoelSpeed (commented):
/approve

elmiko (approved):
/lgtm

PR #51, opened 2021-05-25
WIP: Test with wrong type in object source

lobziik:
Seems we need figure out smth with unmarshalling procedure...


elmiko:
> Seems we need figure out smth with unmarshalling procedure...

++, we should definitely be using some sort of structured schema validation. we know what these files will look like.

Danil-Grigorev:
This is fixed in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/52 @elmiko @lobziik

lobziik:
/close 
Due to #52

PR #52, opened 2021-05-26
Use strict yaml unmarshal to ensure partial type safety

Danil-Grigorev:
Is seems a deployment without replicas is perfectly marshable to
daemonSet, this is a nature of parsing some byte stream to a compatible
resource holder.

cc @lobziik


lobziik:
Does not completely resolve the issue, but better than before

/approve
/lgtm

elmiko (approved):
this generally make sense to me, thanks Danil
/lgtm

PR #53, opened 2021-05-26
Validate object source types on startup

Danil-Grigorev:


pkg/cloud/common/validate.go
```diff
@@ -0,0 +1,25 @@
...
+func init() {
+	err := validator.SetValidationFunc("supportedType", supportedType)
+	utilruntime.Must(err)
+}
+
+// supportedType will validate the type of Object is supported
+func supportedType(object interface{}, _ string) error {
+	switch object.(type) {
+	case appsv1.Deployment, corev1.ConfigMap:
+		return nil
```
elmiko, line 21: would we eventually put more schema-based validations in here?

elmiko (commented):
i think this looks like a decent approach, just have one small question

JoelSpeed:
Is this PR something we want to be keeping/do we need it still?

pkg/cloud/common/validate.go
Danil-Grigorev: It is more of a manual restriction, not something we don't support because of the schema. Probably not, there is always a list

lobziik:
/close

If we want this, think its better to proceed in #110 or followups

PR #56, opened 2021-06-08
Ensure we use strict yaml unmarshalling in render

Danil-Grigorev:
This would guarantee that we won't accidentally run CCCMO render and accept a miss-configured deprecated platform configuration or we won't read the newer additions to Infrastructure resource API fields.

- Add missing tests for file write.
- Revendor with latest kapi/oapi changes.


elmiko (approved):
/lgtm

PR #57, opened 2021-06-08
WIP: Hide cloud provider specific things behind interface

lobziik:


pkg/config/config.go
```diff
@@ -42,103 +38,88 @@ func GetProviderFromInfrastructure(infra *configv1.Infrastructure) (configv1.Pla
...
-		return images.CloudControllerManagerOpenStack
-	case configv1.AzurePlatformType:
-		return images.CloudControllerManagerAzure
-	default:
-		return ""
-	}
+	config.ImagesFileContent = imagesFileContent
+
+	commonImages, err := getCommonImageReferences(imagesFileContent)
+	config.CommonImages = commonImages
```
Danil-Grigorev, line 76: This looks unreadable, OperatorConfig is partially initialized in the beginning, and partially here. Let's initialize in one place.

pkg/config/config.go
```diff
@@ -42,103 +38,88 @@ func GetProviderFromInfrastructure(infra *configv1.Infrastructure) (configv1.Pla
...
+		ManagedNamespace: managedNamespace,
 	}
 
-	i := imagesReference{}
-	if err := json.Unmarshal(data, &i); err != nil {
-		return imagesReference{}, err
+	imagesFileContent, err := readImagesJSONFile(imagesFile)
+	if err != nil {
+		klog.Errorf("Unable to decode images file from location %s", imagesFile, err)
+		return config, err
```
Danil-Grigorev, line 70: This is not idiomatic to return half baked config and an error. Return an error and nil as for config.

pkg/cloud/cloud.go
```diff
@@ -1,44 +1,24 @@
...
-// This pod is required for platforms that allow multiple Node initialization from
-// a single CCM instance which is not bound to link-local VM IP address and node name.
-// Allows to initalize master Nodes immediately after they are created by the installer.
-func GetBootstrapResources(platform configv1.PlatformType) []client.Object {
-	switch platform {
+	switch operatorConfig.Platform {
 	case configv1.AWSPlatformType:
-		return aws.GetBootstrapResources()
+		return aws.NewAssets(operatorConfig)
+	case configv1.OpenStackPlatformType:
```
Danil-Grigorev, line 19: These cases should use <cloud>.GetPlatformType methods for cases.

pkg/cloud/common/templates.go
```diff
@@ -0,0 +1,82 @@
...
+	err = yaml.UnmarshalStrict(buf.Bytes(), object)
+	if err != nil {
+		klog.Errorf("Cannot decode data from embedded resource %v: %v", tmpl.TemplateSource.Path, err)
+		return nil, err
+	}
+
+	object = substituteRenderedObject(object.(client.Object), providerContext)
+
+	return object.(client.Object), nil
+}
```
Danil-Grigorev, line 61: As this method is called on every reconcile, this is unnecessary. Only DeepCopy and substitute part could belong and be encapsulated there.

You read filesystem once and store the object in the client.Object of the TemplateSource. This will allow validation to work in the future, and make code much much simpler, as you would work not with templates, but with client.Objects only.

pkg/cloud/common/templates.go
```diff
@@ -0,0 +1,82 @@
...
+	for i, tmpl := range templates {
+		object, err := tmpl.render(providerContext)
+		if err != nil {
+			return nil, err
+		}
+
+		objects[i] = object
+	}
+	return objects, nil
+}
```
Danil-Grigorev, line 82: This method should be split into two. First will render templates and store data into client.Object, and another could continuously be called from every reconcile and do substitution of config values.

Danil-Grigorev: The first method will be called only once, while substitution could be called repeatedly.

pkg/cloud/common/templates.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type TemplateSource struct {
+	Object client.Object
+	Path   string
+}
+
+type ObjectTemplate struct {
+	TemplateSource
+	Template *template.Template
```
Danil-Grigorev, line 21: I do think that template should not be in the structure, and just be an internal way to read embed fs.

Danil-Grigorev (commented):
WIP, but looks good. We could make it work.

elmiko (commented):
i like the direction this is going. i think the interface is a nice addition and helps make it clear what needs to be added for each platform. i don't have a strong feeling about the template vs substitution, i think there are good points on both sides. i think we should probably discuss as a group and see if we can arrive at a decision.

lobziik:
/close 

proceed in #110

PR #58, opened 2021-06-15
[OCPCLOUD-1162] Implement image substitution for Pod and generalise the approach for other objects

Danil-Grigorev:
This will add support for image substitution in static pod, used in render implementation.

### DIFF
--- a/pkg/substitution/substitution.go
+++ b/pkg/substitution/substitution.go
@@ -2,7 +2,8 @@ package substitution
 
 import (
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/config"
-	v1 "k8s.io/api/apps/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


pkg/substitution/substitution.go
```diff
@@ -14,27 +15,27 @@ const (
...
 
-// setDeploymentImages substitutes controller containers in Deployment with correct image
-func setDeploymentImages(config config.OperatorConfig, d *v1.Deployment) {
-	for i, container := range d.Spec.Template.Spec.Containers {
-		if container.Name != cloudControllerManagerName {
+// setCloudControllerImage substitutes controller containers in provided pod specs with correct image
+func setCloudControllerImage(config config.OperatorConfig, p corev1.PodSpec) corev1.PodSpec {
+	updatedPod := *p.DeepCopy()
+	for i, container := range p.Containers {
+		substitute := ""
```
JoelSpeed, line 22: Bit of a nit, but, would it be better to call this `substituteName`, would be less ambiguous IMO because substitute could be a verb of noun but we definitely want a noun here

pkg/substitution/substitution.go
```diff
@@ -14,27 +15,27 @@ const (
...
-		d.Spec.Template.Spec.Containers[i].Image = config.ControllerImage
-	}
-}
-
-func setDaemonSetImage(config config.OperatorConfig, d *v1.DaemonSet) {
-	for i, container := range d.Spec.Template.Spec.Containers {
-		if container.Name != cloudNodeManagerName {
-			continue
+		if substitute != "" {
+			klog.Infof("Substituting %q in with %s", container.Name, substitute)
```
JoelSpeed, line 33: Perhaps a bit more explicit what's happening here
```suggestion
			klog.Infof("Substituting container image for container %q with %q", container.Name, substitute)
```

JoelSpeed (commented):
LGTM, got a couple of nits, LMK if you disagree

Danil-Grigorev:
@JoelSpeed The PR is ready, but we can't merge it just yet, or we risk destroying our CI once https://github.com/openshift/installer/pull/4947 is merged, just after operator is added to release. We need to add all supported provider forks into release first, then merge this PR. Currently it is Azure and AWS.

/hold

JoelSpeed (commented):
/approve

JoelSpeed:
> @JoelSpeed The PR is ready, but we can't merge it just yet, or we risk destroying our CI once openshift/installer#4947 is merged, just after operator is added to release. We need to add all supported provider forks into release first, then merge this PR. Currently it is Azure and AWS.

@Danil-Grigorev I don't see how this PR would affect that specifically? Does the operator not wait until there is a feature gate present before doing anything? I would have thought during bootstrap it would also only generate a pod if the FG is present or it knows that platform is specifically GA'd as out of tree? Did we change what we are doing here?

Danil-Grigorev:
/hold cancel

It is safe to merge, as the operator could not be added to release before each provider addition.

PR #59, opened 2021-06-17
Update component name in Owners

Fedosin:
There is no such registered component as "cloud-controller-manager".
It should be:

component: "Cloud Compute"
subcomponent: "Cloud Controller Manager"


elmiko (commented):
/lgtm

PR #61, opened 2021-06-18
[OCPCLOUD-1181] Cloud conf sync controller

lobziik:
Separate controller for react on changes in cloud-config configmaps in `openshift-config` and `openshift-config-managed` namespaces

### DIFF
--- a/controllers/cloud_config_sync_controller.go
+++ b/controllers/cloud_config_sync_controller.go
@@ -1,0 +1,158 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	openshiftConfigNamespace        = "openshift-config"
+	openshiftManagedConfigNamespace = "openshift-config-managed"
+	managedCloudConfigMapName       = "kube-cloud-config"
+
+	cloudConfigMapName = "cloud-conf"
+	defaultConfigKey   = "cloud.conf"
+)
+
+type CloudConfigReconciler struct {
+	client.Client
+	Scheme          *runtime.Scheme
+	Recorder        record.EventRecorder
+	TargetNamespace string
+}
+
+func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.Infof("%s emitted event, syncing cloud-conf ConfigMap", req)
+
+	// Use kube-cloud-config from openshift-config-managed namespace as default source.
+	// If it is not exists try to use cloud-config reference from infra resource.
+	// https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go#L82
+	defaultSourceCMObjectKey := client.ObjectKey{
+		Name: managedCloudConfigMapName, Namespace: openshiftManagedConfigNamespace,
+	}
+	sourceCM := &corev1.ConfigMap{}
+	if err := r.Get(ctx, defaultSourceCMObjectKey, sourceCM); errors.IsNotFound(err) {
+		klog.Warningf("managed cloud-config is not found, falling back to infrastructure config")
+		infra := &configv1.Infrastructure{}
+		if err := r.Get(ctx, client.ObjectKey{Name: infrastructureResourceName}, infra); err != nil {
+			klog.Errorf("infrastructure resource not found")
+			return ctrl.Result{}, err
+		}
+
+		openshiftUnmanagedCMKey := client.ObjectKey{Name: infra.Spec.CloudConfig.Name, Namespace: openshiftConfigNamespace}
+		if err := r.Get(ctx, openshiftUnmanagedCMKey, sourceCM); err != nil {
+			klog.Errorf("unable to get cloud-config for sync")
+			return ctrl.Result{}, err
+		}
+		sourceCM, err = r.prepareSourceConfigMap(sourceCM, infra)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	} else if err != nil {
+		klog.Errorf("unable to get managed cloud-config for sync")
+		return ctrl.Result{}, err
+	}
+
+	targetCM := &corev1.ConfigMap{}
+	targetConfigMapKey := client.ObjectKey{
+		Namespace: r.TargetNamespace,
+		Name:      cloudConfigMapName,
+	}
+
+	// If the config does not exist, it will be created later, so we can ignore a Not Found error
+	if err := r.Get(ctx, targetConfigMapKey, targetCM); err != nil && !errors.IsNotFound(err) {
+		klog.Errorf("unable to get target cloud-config for sync")
+		return ctrl.Result{}, err
+	}
+
+	if r.isCloudConfigEqual(sourceCM, targetCM) {
+		klog.Infof("source and target cloud-config content are equal, no sync needed")
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.syncCloudConfigData(ctx, sourceCM, targetCM); err != nil {
+		klog.Errorf("unable to sync cloud config")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *CloudConfigReconciler) prepareSourceConfigMap(source *corev1.ConfigMap, infra *configv1.Infrastructure) (*corev1.ConfigMap, error) {
+	// Keys might be different between openshift-config/cloud-config and openshift-config-managed/kube-cloud-config
+	// Always use "cloud.conf" which is default one across openshift
+	cloudConfCm := source.DeepCopy()
+	if _, ok := cloudConfCm.Data[defaultConfigKey]; ok {
+		return cloudConfCm, nil
+	}
+
+	infraConfigKey := infra.Spec.CloudConfig.Key
+	if val, ok := cloudConfCm.Data[infraConfigKey]; ok {
+		cloudConfCm.Data[defaultConfigKey] = val
+		delete(cloudConfCm.Data, infraConfigKey)
+		return cloudConfCm, nil
+	}
+	return nil, fmt.Errorf(
+		"key %s specified in infra resource does not found in source configmap %s",
+		infraConfigKey, client.ObjectKeyFromObject(source),
+	)
+}
+
+func (r *CloudConfigReconciler) isCloudConfigEqual(source *corev1.ConfigMap, target *corev1.ConfigMap) bool {
+	return source.Immutable == target.Immutable &&
+		reflect.DeepEqual(source.Data, target.Data) && reflect.DeepEqual(source.BinaryData, target.BinaryData)
+}
+
+func (r *CloudConfigReconciler) syncCloudConfigData(ctx context.Context, source *corev1.ConfigMap, target *corev1.ConfigMap) error {
+	target.SetName(cloudConfigMapName)
+	target.SetNamespace(r.TargetNamespace)
+	target.Data = source.Data
+	target.BinaryData = source.BinaryData
+	target.Immutable = source.Immutable
+
+	// check if target config exists, create if not
+	err := r.Get(ctx, client.ObjectKeyFromObject(target), &corev1.ConfigMap{})
+
+	if err != nil && errors.IsNotFound(err) {
+		return r.Create(ctx, target)
+	} else if err != nil {
+		return err
+	}
+
+	return r.Update(ctx, target)
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *CloudConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	build := ctrl.NewControllerManagedBy(mgr).
+		For(
+			&corev1.ConfigMap{},
+			builder.WithPredicates(
+				predicate.Or(
+					ownCloudConfigPredicate(r.TargetNamespace),
+					openshiftCloudConfigMapPredicates(),
+				),
+			),
+		).
+		Watches(
+			&source.Kind{Type: &configv1.Infrastructure{}},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(infrastructurePredicates()),
+		)
+
+	return build.Complete(r)
+}
--- a/controllers/cloud_config_sync_controller_test.go
+++ b/controllers/cloud_config_sync_controller_test.go
@@ -1,0 +1,279 @@
+package controllers
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	infraCloudConfName = "test-config"
+	infraCloudConfKey  = "foo"
+)
+
+func makeInfrastrucutreResource() *configv1.Infrastructure {
+	return &configv1.Infrastructure{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: infrastructureResourceName,
+		},
+		Spec: configv1.InfrastructureSpec{
+			CloudConfig: configv1.ConfigMapFileReference{
+				Name: infraCloudConfName,
+				Key:  infraCloudConfKey,
+			},
+			PlatformSpec: configv1.PlatformSpec{
+				Type: configv1.AzurePlatformType,
+			},
+		},
+	}
+}
+
+func makeInfraCloudConifg() *corev1.ConfigMap {
+	return &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
+		Name:      infraCloudConfName,
+		Namespace: openshiftConfigNamespace,
+	}, Data: map[string]string{infraCloudConfKey: "bar"}}
+}
+
+func makeManagedCloudConifg() *corev1.ConfigMap {
+	return &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
+		Name:      managedCloudConfigMapName,
+		Namespace: openshiftManagedConfigNamespace,
+	}, Data: map[string]string{"cloud.conf": "bar"}}
+}
+
+var _ = Describe("isCloudConfigEqual reconciler method", func() {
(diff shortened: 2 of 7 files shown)


cmd/cluster-cloud-controller-manager-operator/main.go
```diff
@@ -127,12 +138,19 @@ func main() {
...
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
 
+	infraResource, err := getInfrastructureResource(mgr.GetAPIReader())
+	if err != nil {
+		setupLog.Error(err, "unable to get platform type from infrastructure resource")
+		os.Exit(1)
+	}
```
JoelSpeed, line 152: Out of interest, why pull this here rather than during the reconcile? Is it because we aren't caching the cluster scoped resources? If so, might want to add a comment with the motivation

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+	Scheme                 *runtime.Scheme
+	Recorder               record.EventRecorder
+	watcher                ObjectWatcher
+	TargetNamespace        string
+	InfrastructureResource *configv1.Infrastructure
+}
+
+func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.Infof("%s emitted event, syncing cloud-conf ConfigMap", req)
+	var err error
```
JoelSpeed, line 42: No need to declare this here, decalre it inline with the `:=` operator where it's first declared

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+
+	sourceCM := &corev1.ConfigMap{}
+	sourceConfigMapKey := client.ObjectKey{Name: req.Name, Namespace: req.Namespace}
+	if req.Namespace == r.TargetNamespace {
+		klog.Infof("target cloud-config somewhat changed, syncing with cloud-config from openshift-config namespace")
```
JoelSpeed, line 47: Nit
```suggestion
		klog.Infof("target cloud-config has changed, syncing with cloud-config from openshift-config namespace")
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+
+import (
+	"context"
+	"fmt"
+	configv1 "github.com/openshift/api/config/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog"
```
JoelSpeed, line 11: Should use klog v2 now
```suggestion
	"k8s.io/klog/v2"
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+	"reflect"
```
JoelSpeed, line 12: Can we group the imports so that core imports are in their own group please? Shouldn't we have a lint rule for goimports somewhere?

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+		sourceConfigMapKey = client.ObjectKey{
+			Name: r.InfrastructureResource.Spec.CloudConfig.Name, Namespace: openshiftConfigNamespace,
+		}
+	}
+	err = r.Get(ctx, sourceConfigMapKey, sourceCM)
+	if err != nil {
+		klog.Errorf("unable to get source cloud-config: %v", err)
+		return ctrl.Result{}, err
+	}
+	sourceCM, err = r.prepareSourceConfigMap(sourceCM)
```
JoelSpeed, line 57: Might be less error prone/confusing if we use a new variable here instead?
```suggestion
	preparedCM, err := r.prepareSourceConfigMap(sourceCM)
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+	sourceCM := &corev1.ConfigMap{}
+	sourceConfigMapKey := client.ObjectKey{Name: req.Name, Namespace: req.Namespace}
+	if req.Namespace == r.TargetNamespace {
+		klog.Infof("target cloud-config somewhat changed, syncing with cloud-config from openshift-config namespace")
+		sourceConfigMapKey = client.ObjectKey{
+			Name: r.InfrastructureResource.Spec.CloudConfig.Name, Namespace: openshiftConfigNamespace,
+		}
+	}
+	err = r.Get(ctx, sourceConfigMapKey, sourceCM)
+	if err != nil {
```
JoelSpeed, line 53: Bit of a stylistic thing, but typically when a Go function only returns an error, people inline the if statement like
```suggestion
	if err := r.Get(ctx, sourceConfigMapKey, sourceCM); err != nil {
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+			return ctrl.Result{}, err
+		}
+	}
+	if err != nil {
+		klog.Errorf("unable to obtain target cloud-config: %v", err)
+		return ctrl.Result{}, err
+	}
+
+	if r.isCloudConfigEqual(sourceCM, targetCM) {
+		klog.Infof("source and target cloud-configs content are equal, no sync needed")
```
JoelSpeed, line 83: Nit
```suggestion
		klog.Infof("source and target cloud-config content are equal, no sync needed")
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+		return ctrl.Result{}, nil
+	} else {
```
JoelSpeed, line 85: This else isn't needed because the previous block always returns, you can un-indent this block by removing the else

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+	target.SetName(cloudConfigMapName)
+	target.SetNamespace(r.TargetNamespace)
+	target.Data = source.Data
+	target.BinaryData = source.BinaryData
+	target.Immutable = source.Immutable
+
+	// check if target config exists, crate if not
+	err := r.Get(ctx, client.ObjectKeyFromObject(target), &corev1.ConfigMap{})
+
+	if err != nil && errors.IsNotFound(err) {
```
JoelSpeed, line 130: Nit, spelling and best practice to "pair" the error declaration with the check
```suggestion
	// check if target config exists, create if not
	err := r.Get(ctx, client.ObjectKeyFromObject(target), &corev1.ConfigMap{})
	if err != nil && errors.IsNotFound(err) {
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+		err = r.Create(ctx, target)
```
JoelSpeed, line 131: This error is unchecked, did you mean to return here instead of declare the variable?
```suggestion
		return r.Create(ctx, target)
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *CloudConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	watcher, err := NewObjectWatcher(WatcherOptions{
+		Cache:  mgr.GetCache(),
+		Scheme: mgr.GetScheme(),
+	})
+	if err != nil {
+		return err
+	}
+	r.watcher = watcher
```
JoelSpeed, line 148: What's the watcher do? I can't see this being used anywhere?

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+		configMap, ok := obj.(*corev1.ConfigMap)
+		return ok && configMap.GetNamespace() == targetNamespace && configMap.GetName() == cloudConfigMapName
+	}
+
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool { return false },
+		UpdateFunc: func(e event.UpdateEvent) bool { return isOwnCloudConfigMap(e.ObjectNew) },
+		DeleteFunc: func(e event.DeleteEvent) bool { return isOwnCloudConfigMap(e.Object) },
+		// filter out generic events for avoid redundant reconciliation and be sure that cloud config changed if event was emitted
+		GenericFunc: func(e event.GenericEvent) bool { return false },
```
JoelSpeed, line 172: I disagree, I think we should be reconciling all events. Extra reconciles are not harmful if a controller is designed appropriately.

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+		configMap, ok := obj.(*corev1.ConfigMap)
+
+		if ok {
+			isCloudConfName := configMap.GetName() == infraCloudConfigName || configMap.GetName() == managedCloudConfigMapName
+			isCloudConfNamespace := configMap.GetNamespace() == openshiftManagedConfigNamespace || configMap.GetNamespace() == openshiftConfigNamespace
+
+			return isCloudConfName && isCloudConfNamespace
+		}
+
+		return ok
```
JoelSpeed, line 187: I would reverse the if statement here to reduce indentation
```suggestion
		configMap, ok := obj.(*corev1.ConfigMap)
		if !ok {
			return false
		}
		
		isCloudConfName := configMap.GetName() == infraCloudConfigName || configMap.GetName() == managedCloudConfigMapName
		isCloudConfNamespace := configMap.GetNamespace() == openshiftManagedConfigNamespace || configMap.GetNamespace() == openshiftConfigNamespace

		return isCloudConfName && isCloudConfNamespace
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func ownCloudConfigPredicate(targetNamespace string) predicate.Funcs {
+	isOwnCloudConfigMap := func(obj runtime.Object) bool {
+		configMap, ok := obj.(*corev1.ConfigMap)
+		return ok && configMap.GetNamespace() == targetNamespace && configMap.GetName() == cloudConfigMapName
+	}
+
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool { return false },
```
JoelSpeed, line 168: We should reconcile on create still, no harm in it

controllers/cloud_config_sync_controller_test.go
```diff
@@ -0,0 +1,205 @@
...
+	infraCloudConfKey := "foo"
+
+	infrastructureResource := &configv1.Infrastructure{
+		Spec: configv1.InfrastructureSpec{
+			CloudConfig: configv1.ConfigMapFileReference{
+				Name: infraCloudConfName,
+				Key:  infraCloudConfKey,
+			},
+			PlatformSpec: configv1.PlatformSpec{
+				Type: "VSphere",
```
JoelSpeed, line 46: In the configv1 package there are a number of different constants, would prefer to use one of those here instead if possible. Also, what is reading from the platform spec? We normally read from the platform status?

controllers/cloud_config_sync_controller_test.go
```diff
@@ -0,0 +1,205 @@
...
+			Client:                 cl,
+			Scheme:                 scheme.Scheme,
+			Recorder:               rec,
+			InfrastructureResource: infrastructureResource,
+			TargetNamespace:        targetNamespaceName,
+		}
+
+		// First reconcile
+		_, err := r.Reconcile(ctx, getRequestForObject(infraCloudConfig))
+		Expect(err).Should(Succeed())
```
JoelSpeed, line 78: Why are we manually calling reconcile instead of running it in the background like a manager in other tests?

JoelSpeed (commented):
I think we may be able to improve the testing of this reconciler. Would be good to see some unit tests for individual functions as well as full reconciler tests when it's running under the manager. Please see my comments inline for some suggestions on improvements

cmd/cluster-cloud-controller-manager-operator/main.go
lobziik: Initial motivation was about setting up predicates. Need to grab CM name from infra resource for this

https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/61/files#diff-978f2fef08d39cafbbbfe8a75f1bf39c59f7e1a7a3c7ba1f74fd4979975cd22bR181

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// Use kube-cloud-config from openshift-config-managed namespace as default source.
+	// If it is not exists try to use cloud-config reference from infra resource.
+	// https://github.com/openshift/library-go/blob/master/pkg/operator/configobserver/cloudprovider/observe_cloudprovider.go#L82
+	defaultSourceCMObjectKey := client.ObjectKey{
+		Name: managedCloudConfigMapName, Namespace: openshiftManagedConfigNamespace,
+	}
+	sourceCM := &corev1.ConfigMap{}
+	if err := r.Get(ctx, defaultSourceCMObjectKey, sourceCM); errors.IsNotFound(err) {
+		klog.Warningf("managed cloud-config is not found trying to get infra one")
```
JoelSpeed, line 48: Possibly a better message?
```suggestion
		klog.Warningf("managed cloud-config is not found, falling back to infrastructure config")
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,158 @@
...
+			klog.Errorf("infrastructure resource not found")
+			return ctrl.Result{}, err
+		}
+
+		openshiftUnmanagedCMKey := client.ObjectKey{Name: infra.Spec.CloudConfig.Name, Namespace: openshiftConfigNamespace}
+		if err := r.Get(ctx, openshiftUnmanagedCMKey, sourceCM); err != nil {
+			klog.Errorf("unable to get cloud-config for sync")
+			return ctrl.Result{}, err
+		}
+		sourceCM, err = r.prepareSourceConfigMap(sourceCM, infra)
```
JoelSpeed, line 60: Do we only prepare the CM when it's from the infrastructure?

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,158 @@
...
+		return ctrl.Result{}, err
+	}
+
+	targetCM := &corev1.ConfigMap{}
+	targetConfigMapKey := client.ObjectKey{
+		Namespace: r.TargetNamespace,
+		Name:      cloudConfigMapName,
+	}
+
+	if err := r.Get(ctx, targetConfigMapKey, targetCM); err != nil && !errors.IsNotFound(err) {
```
JoelSpeed, line 75:
```suggestion
	// If the config does not exist, it will be created later, so we can ignore a Not Found error
	if err := r.Get(ctx, targetConfigMapKey, targetCM); err != nil && !errors.IsNotFound(err) {
```

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,158 @@
...
+	return r.Update(ctx, target)
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *CloudConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+
+	build := ctrl.NewControllerManagedBy(mgr).
+		For(
+			&corev1.ConfigMap{},
+			builder.WithPredicates(ownCloudConfigPredicate(r.TargetNamespace)),
```
JoelSpeed, line 144: I think there's a way you can `Or` predicates, which would mean you don't need the first set of `Watches` below

controllers/common_constants.go
```diff
@@ -0,0 +1,5 @@
+package controllers
```
JoelSpeed, line 1: Typically this file would be `common_consts.go` I think, not major, happy to keep as is, but I think that's the convention I've seen in the past

controllers/cloud_config_sync_controller_test.go
```diff
@@ -0,0 +1,275 @@
...
+func makeManagedCloudConifg() *corev1.ConfigMap {
+	return &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
+		Name:      managedCloudConfigMapName,
+		Namespace: openshiftManagedConfigNamespace,
+	}, Data: map[string]string{"cloud.conf": "bar"}}
+}
+
+func TestCloudisCloudConfigEqual(t *testing.T) {
+	reconciler := &CloudConfigReconciler{}
+	assert.True(t, reconciler.isCloudConfigEqual(makeManagedCloudConifg(), makeManagedCloudConifg()))
```
JoelSpeed, line 61: Can we use gomega here instead so we don't have a mix of two testing frameworks?

controllers/cloud_config_sync_controller_test.go
```diff
@@ -0,0 +1,275 @@
...
+			apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(infra), infra)),
+		).Should(BeTrue())
+	})
+
+	It("config should be synced up after first reconcile", func() {
+		syncedCloudConfigMap := &corev1.ConfigMap{}
+		Eventually(func() error {
+			return cl.Get(ctx, syncedConfigMapKey, syncedCloudConfigMap)
+		}).Should(Succeed())
+		Expect(syncedCloudConfigMap.Data[defaultConfigKey]).Should(BeEquivalentTo("bar"))
```
JoelSpeed, line 202: Perhaps move the equivalence check inside the eventually? Only needs one assertion here as we discussed yesterday

controllers/cloud_config_sync_controller_test.go
```diff
@@ -0,0 +1,275 @@
...
+	})
+
+	It("config should not be updated if source and target config content are identical", func() {
+		syncedCloudConfigMap := &corev1.ConfigMap{}
+		Expect(cl.Get(ctx, syncedConfigMapKey, syncedCloudConfigMap)).Should(Succeed())
+		initialCMresourceVersion := syncedCloudConfigMap.ResourceVersion
+
+		request := reconcile.Request{NamespacedName: client.ObjectKey{Name: "foo", Namespace: "bar"}}
+		_, err := reconciler.Reconcile(ctx, request)
+		Expect(err).Should(Succeed())
```
JoelSpeed, line 251: The manager is handling this in the background, shouldn't need to call the reconciler directly ever

controllers/cloud_config_sync_controller.go
lobziik: Yes. In openshift-config-managed its already prepared.

controllers/cloud_config_sync_controller_test.go
lobziik: Here I wanted to ensure that reconcile was definitely run and finished successfully and check things after. it's most convinient way which i found... Just for make this particular test scope bit more narrow.

lobziik:
cc @elmiko :)

elmiko (commented):
this mostly makes sense to me, i just have a couple questions.

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		return cloudConfCm, nil
+	}
+	return nil, fmt.Errorf(
+		"key %s specified in infra resource does not found in source configmap %s",
+		infraConfigKey, client.ObjectKeyFromObject(source),
+	)
+}
+
+func (r *CloudConfigReconciler) isCloudConfigEqual(source *corev1.ConfigMap, target *corev1.ConfigMap) bool {
+	return source.Immutable == target.Immutable &&
```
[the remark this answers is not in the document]
lobziik: https://kubernetes.io/docs/concepts/configuration/configmap/ - here is some info.

I just compare all meaningful fields there.

controllers/cloud_config_sync_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		reflect.DeepEqual(source.Data, target.Data) && reflect.DeepEqual(source.BinaryData, target.BinaryData)
+}
+
+func (r *CloudConfigReconciler) syncCloudConfigData(ctx context.Context, source *corev1.ConfigMap, target *corev1.ConfigMap) error {
+
```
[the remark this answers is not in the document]
lobziik: >any reasoning here?

controllers/cloud_config_sync_controller_test.go
JoelSpeed: Bit of an odd pattern, maybe worth a comment to explain the reasoning as you've just said there

JoelSpeed (commented):
/approve

One nit in the tests but otherwise this is good to go I think, thanks @lobziik

elmiko (approved):
thanks for the link @lobziik !
/lgtm

PR #62, opened 2021-06-18
Azure CCM assets

lobziik:
depends on https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/86

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_03_rbac_provider.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_03_rbac_provider.yaml
@@ -187,3 +187,46 @@ subjects:
 - kind: ServiceAccount
   name: cloud-controller-manager
   namespace: openshift-cloud-controller-manager
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+  name: cloud-node-manager
+  namespace: openshift-cloud-controller-manager
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cloud-node-manager
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+rules:
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["watch","list","get","update", "patch"]
+  - apiGroups: [""]
+    resources: ["nodes/status"]
+    verbs: ["patch"]
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: cloud-node-manager
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cloud-node-manager
+subjects:
+  - kind: ServiceAccount
+    name: cloud-node-manager
+    namespace: openshift-cloud-controller-manager
--- a/pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
+++ b/pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
@@ -1,0 +1,97 @@
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: azure-cloud-controller-manager
+  namespace: openshift-cloud-controller-manager
+  labels:
+    app: azure-cloud-controller-manager
+spec:
+  selector:
+    matchLabels:
+      app: azure-cloud-controller-manager
+  template:
+    metadata:
+      labels:
+        app: azure-cloud-controller-manager
+    spec:
+      hostNetwork: true
+      serviceAccountName: cloud-controller-manager
+      priorityClassName: system-cluster-critical
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - topologyKey: "kubernetes.io/hostname"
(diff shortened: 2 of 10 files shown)


lobziik:
cc @Fedosin @elmiko @JoelSpeed 

Folks, PTAL when you will have time :)

JoelSpeed (commented):
/approve

Looks ok to me, how has this been tested?

lobziik:
/hold

Found problem with cloud-conf syncing

pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+    metadata:
+      labels:
+        app: azure-cloud-controller-manager
+    spec:
+      hostNetwork: true
+      serviceAccountName: cloud-controller-manager
+      priorityClassName: system-cluster-critical
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      tolerations:
```
elmiko, line 22: does this also need to tolerate `not-ready` ?

seems like aws does, https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/aws/assets/deployment.yaml#L55

pkg/cloud/azure/assets/cloud-node-manager-daemonset.yaml
```diff
@@ -0,0 +1,56 @@
...
+        k8s-app: azure-cloud-node-manager
+      annotations:
+        cluster-autoscaler.kubernetes.io/daemonset-pod: "true"
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: cloud-node-manager
+      hostNetwork: true   # required to fetch correct hostname
+      nodeSelector:
+        kubernetes.io/os: linux
+      tolerations:
```
elmiko, line 25: same question here, i'm guessing since this uses the host network that it will need to tolerate `not-ready` as well.

elmiko (commented):
couple questions

pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
JoelSpeed: None of our components should be tolerating that taint, if we do, we will schedule onto a node without networking, which will likely cause our pods to crash loop as they can't contact the APIs they need to contact

pkg/cloud/azure/assets/cloud-node-manager-daemonset.yaml
JoelSpeed: Host network maybe means we can allow it to tolerate the taint, though IMO, we should only add the toleration if we are certain we need it. I'd suggest being conservative initially

lobziik:
Updated, depends on #86 now. Changes with inline arguments and check for CNI was also ported there

pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
JoelSpeed: It will now need to tolerate the not-ready based on the additions made in #76

JoelSpeed (commented):
You'll need to look at the test failures here and make sure you are meeting the spec for the daemonset and deployment pod specs https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-cloud-controller-manager-operator/62/pull-ci-openshift-cluster-cloud-controller-manager-operator-master-verify/1413158319933100032#1:build-log.txt%3A278

lobziik:
https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/77
This needed as well

pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: azure-cloud-controller-manager
+  template:
+    metadata:
+      labels:
+        app: azure-cloud-controller-manager
+    spec:
+      hostNetwork: true
```
Danil-Grigorev, line 17: Suppose hostNetwork is not required for other controllers except `cloud-node`

pkg/cloud/azure/assets/cloud-node-manager-daemonset.yaml
```diff
@@ -0,0 +1,78 @@
...
+          command:
+            - /bin/bash
+            - -c
+            - |
+              #!/bin/bash
+              set -o allexport
+              if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+                source /etc/kubernetes/apiserver-url.env
+              else
+                URL_ONLY_KUBECONFIG=/etc/kubernetes/kubeconfig
```
Danil-Grigorev, line 56: This line is no longer needed upon #88

manifests/0000_26_cloud-controller-manager-operator_03_rbac_provider.yaml
```diff
@@ -187,3 +187,47 @@ subjects:
 - kind: ServiceAccount
   name: cloud-controller-manager
   namespace: openshift-cloud-controller-manager
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
```
Danil-Grigorev, line 195: Those are annotations and not labels @lobziik This is why this resource is not created by CVO. Same thing for other resources in this file.

pkg/config/config.go
```diff
@@ -20,11 +20,12 @@ type imagesReference struct {
...
-	ManagedNamespace string
-	ControllerImage  string
-	CloudNodeImage   string
-	IsSingleReplica  bool
-	Platform         configv1.PlatformType
+	ManagedNamespace   string
+	ControllerImage    string
+	CloudNodeImage     string
+	IsSingleReplica    bool
+	InfrastructureName string
```
Danil-Grigorev, line 27: Would be better to name this var `ClusterName` as it is later passed to the `--cluster-name` flag.

lobziik: Would prefer keep `InfrastructureName` consistent across our code since it titled this way in infrastructure resource. Renamed variable in manifests for better consistency.

PR #65, opened 2021-06-23
No-op change. Do not merge

Danil-Grigorev:
This PR is made for the e2e tests history

/hold


Fedosin:
/close

I think we don't need this anymore

PR #66, opened 2021-06-24
Add golangci-lint checks

Fedosin:
This commit adds lint validations from golangci-lint project. Also it introduces "verify" target, that combines "fmt", "vet" and "lint".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,11 @@ endif
 
 all: build
 
+verify: fmt vet lint
+
 # Run tests
 ENVTEST_ASSETS_DIR=/tmp/testbin
-test: generate fmt vet manifests
+test: generate verify manifests
 	mkdir -p ${ENVTEST_ASSETS_DIR}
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,8 +71,13 @@ CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
 controller-gen:
 	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1)
 
+# Download controller-gen locally if necessary
+GOLINT = $(shell pwd)/bin/golint
```
JoelSpeed, line 75: Why does this comment have `controller-gen` in it? 🤔

pkg/dependencymagnet/dependencymagnet.go
```diff
@@ -0,0 +1,7 @@
+// +build tools
```
JoelSpeed, line 1: Is this where we want to keep this? I've seen in a lot of places to just have a `tools.go` in the root of the project, WDYT?

JoelSpeed (commented):
At the moment it seems to be linting the vendor directory too, can we make sure that's not the case?

JoelSpeed:
Yeah this looks good now, thanks

I think at some point we should review the linters we have enabled and create a full config as a team with standards that we want to uphold for all of our repos

/approve

elmiko:
talked with @Fedosin in slack, overriding the failing tests since they are not related to this change.
/override ci/prow/e2e-aws-ccm-install

PR #69, opened 2021-06-25
Enable errcheck linter

Fedosin:
This commit enables `errcheck` linter for golangci-lint and additionally corrects all cases pointed out by the linter.

### DIFF
--- a/cmd/cluster-cloud-controller-manager-operator/main.go
+++ b/cmd/cluster-cloud-controller-manager-operator/main.go
@@ -63,7 +63,7 @@ func init() {
 }
 
 func main() {
-	flag.Set("logtostderr", "true")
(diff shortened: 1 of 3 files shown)


cmd/cluster-cloud-controller-manager-operator/main.go
```diff
@@ -64,7 +64,10 @@ func init() {
 }
 
 func main() {
-	flag.Set("logtostderr", "true")
+	if err := flag.Set("logtostderr", "true"); err != nil {
```
lobziik, line 67: Will constantly fail if `-logtostderr` flag not set. CI failing because of this as well.

lobziik: For `render` command its the same.

Can we to remove this block? `klog` put logs to stderr by default anyway.

PR #71, opened 2021-06-29
Move controllers under pkg folder

Fedosin:
Now all operator's modules, except "controllers", are located in "pkg" folder. To keep it unified, this commit puts "controllers" in the same place.


lobziik:
/lgtm

I'm for keeping things unified :)

PR #73, opened 2021-06-30
Add an OpenStack bootstrap static pod manifest

Fedosin:
It will be used to provision a node controller on the bootstrap node to register masters.


JoelSpeed:
We aren't doing bootstrap resources anymore, so this is no longer required

PR #74, opened 2021-06-30
Use upstream openshift images from quay.io

Fedosin:
Instead of using images from third-party registries and our CI registry, it's better to provide links to upstream images from quay.io


Danil-Grigorev (approved):
/approve

PR #77, opened 2021-07-02
Revendor library-go and openshift/api for Azure Stack Hub and Azure support

Danil-Grigorev:
Includes revendoring after openshift/library-go#895 and openshift/library-go#1077 changes were merged.


pkg/controllers/clusteroperator_controller.go
```diff
@@ -97,20 +97,8 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
-	platform, err := config.GetProviderFromInfrastructure(infra)
```
lobziik, line 100: Why this check was removed?

JoelSpeed: Agreed, this check should say so that compose config can take the platform directly still, otherwise you could hit nil panics there

lobziik:
> /test verify

Rebase needed

lobziik:
/lgtm cancel

Would be nice to have another pair of eyes

pkg/config/config_test.go
```diff
@@ -341,6 +372,21 @@ func TestComposeConfig(t *testing.T) {
 			Platform:         configv1.OpenStackPlatformType,
 			IsSingleReplica:  true,
```
JoelSpeed, line 373: You've removed the `isSingleReplica` field above but don't seem to have updated the tests to have a new test for when it's a single replica. This is reducing the test coverage and should be fixed

pkg/controllers/clusteroperator_controller.go
```diff
@@ -97,20 +97,8 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
...
-		if err := r.setStatusAvailable(ctx); err != nil {
-			klog.Errorf("Unable to sync cluster operator status: %s", err)
-			return ctrl.Result{}, err
-		}
-		return ctrl.Result{}, nil
-	}
-
 	// Verify FeatureGate ExternalCloudProvider is enabled for operator to work in TP phase
-	external, err := cloudprovider.IsCloudProviderExternal(platform, featureGate)
+	external, err := cloudprovider.IsCloudProviderExternal(infra.Status.PlatformStatus, featureGate)
```
JoelSpeed, line 101: `infra.Status.PlatformStatus` Can this nil pointer deref?

lobziik: There is a check in library-go https://github.com/openshift/library-go/blob/master/pkg/cloudprovider/external.go#L21

pkg/config/config_test.go
Danil-Grigorev: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/77/files#diff-29cea1b5b831c8655c7155f43e2367cec73e66d1e338f8b3c7877a2f339b8811R359

It is set correctly

pkg/controllers/clusteroperator_controller.go
Danil-Grigorev: There is one above in compose config itself, and it does nil check.

JoelSpeed:
@Danil-Grigorev PTAL, this is failing the verify step for some reason

PR #79, opened 2021-07-05
Allow the operator to list deployments in config namespaces

Fedosin:
Now CCCMO fails because it can list deployments in "openshift-config" and "openshift-config-managed" namespaces.

```txt
E0705 11:26:19.089134       1 reflector.go:138] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:241: Failed to watch *v1.Deployment: failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:openshift-cloud-controller-manager-operator:cluster-cloud-controller-manager" cannot list resource "deployments" in API group "apps" in the namespace "openshift-config"
```

To prevent this, we add required rules to "cluster-cloud-controller-manager" roles in the related namespaces.

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_02_rbac_operator.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_02_rbac_operator.yaml
@@ -107,6 +107,14 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - apps
+    resources:
+      - deployments
+    verbs:
+      - get
+      - list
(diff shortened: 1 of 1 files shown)


JoelSpeed:
/approve
/lgtm

This is a temporary fix, we should extract the cloud config sync to its own binary, run it as a separate container and then undo this RBAC change

Fedosin:
golangci-lint

PR #82, opened 2021-07-06
Inline container args into command

JoelSpeed:
Because we are now using the script to set the environment, args weren't being passed through correctly. To resolve that, lets just inline them and ensure that the args are always inlined going forward

CC @Fedosin As you found the issue in your testing

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
@@ -36,10 +36,9 @@ spec:
           else
             URL_ONLY_KUBECONFIG=/etc/kubernetes/kubeconfig
           fi
-          exec /cluster-controller-manager-operator
-        args:
-        - --leader-elect
-        - "--images-json=/etc/cloud-controller-manager-config/images.json"
+          exec /cluster-controller-manager-operator \
+          --leader-elect \
+          "--images-json=/etc/cloud-controller-manager-config/images.json"
         env:
         - name: RELEASE_VERSION
           value: "0.0.1-snapshot"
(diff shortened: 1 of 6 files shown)


manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
```diff
@@ -36,10 +36,9 @@ spec:
           else
             URL_ONLY_KUBECONFIG=/etc/kubernetes/kubeconfig
           fi
-          exec /cluster-controller-manager-operator
-        args:
-        - --leader-elect
-        - "--images-json=/etc/cloud-controller-manager-config/images.json"
+          exec /cluster-controller-manager-operator \
```
Fedosin, line 39: I think we don't need `exec` here. It's possible just to call the binary as `/cluster-controller-manager-operator \`

JoelSpeed: I copied this over from CNO, what difference does it make to the operation? Have you tested this without the exec? Otherwise I'll spin up a cluster and have a play

Fedosin: yep, I tested it without `exec`. Basically it doesn't matter whether we have it or not, so it's fine to keep it as-is.

JoelSpeed: Ok, let's keep as is for now as I know that's been tested, we can change later if we find a need to

elmiko: i think `exec` is proper here, i have to imagine CNO uses it because it will take over the shell that is running the process instead of spawning a new one. which makes exit and cleanup fairly easy. as long as we don't have anything else to run (and i don't think we do), `exec` is probably preferred here. just my 2 cents

Fedosin (approved):
Tested on OpenStack, with this PR all CCM arguments have been applied correctly.

/lgtm
/approve

PR #84, opened 2021-07-07
Use just one CCM replica for Single Node deployments

Fedosin:
In case of Single Node deployments we need to create just one CCM replica, because there will be a port conflict if there are many
of them.

To do so we observe the Infrastructure resource to check what kind of control plane topology the cluster has. If it's `SingleReplicaTopologyMode` then the operator creates just one CCM replica.

### DIFF
--- a/pkg/substitution/substitution.go
+++ b/pkg/substitution/substitution.go
@@ -5,6 +5,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/substitution/substitution.go
```diff
@@ -49,6 +49,9 @@ func FillConfigValues(config config.OperatorConfig, templates []client.Object) [
 		switch obj := templateCopy.(type) {
 		case *appsv1.Deployment:
 			obj.Spec.Template.Spec = setCloudControllerImage(config, obj.Spec.Template.Spec)
+			if config.IsSingleReplica {
+				obj.Spec.Replicas = &[]int32{1}[0]
```
JoelSpeed, line 53: Replicas is a slice? I thought this should just be an int pointer?

Fedosin: It's a semi-idiomatic way to get a pointer to the integer (kind of &1). Here we create a slice of int32 with just one element, and then we get a pointer to the first element.

JoelSpeed: That's quite hard to understand IMO, would prefer to use something like https://pkg.go.dev/k8s.io/utils/pointer#Int32 which I know we are using elsewhere for this purpose

Fedosin: thanks! this is a more convenient way get int pointers

elmiko (approved):
makes sense to me
/lgtm

PR #85, opened 2021-07-08
Remove bootstrapping render implementation

Fedosin:
With https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/76 we don't need to render static pod manifests on the bootstrap machine. So, "render" binary and all related code can be removed from the operator.


elmiko (approved):
/lgtm

i guess we will need to remove the render language from the enhancement as well

PR #87, opened 2021-07-08
Rename CCCMO deployment

Fedosin:
This PR sets CCCMO deployment name to cluster-cloud-controller-manager-operator.

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
@@ -1,11 +1,129 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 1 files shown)


elmiko (approved):
i think this is a nice change, thanks Mike
/lgtm

are there any other components that still expect the original name? (installer, mco)

Fedosin:
@elmiko fortunately all other components don't rely on this name, so it can be safely changed.

JoelSpeed:

This is going to cause no issue for production users as we don't have this component in nightly builds yet. Once this is merged the upgrade job will use a newer image and we should have no further issues

Danil-Grigorev:
/hold 

@Fedosin This PR will break periodics for 4.7->4.8 if merged as is. Please add tombstone object to remove old deployment before rename.

PR #88, opened 2021-07-09
Update the container init script to not use the legacy workaround

JoelSpeed:
This updates the init script that must be present on pods/containers to remove the legacy fallback. I've also added a note to the test that links to where the current solution was implemented in MCO. ie https://github.com/openshift/machine-config-operator/pull/2232

### DIFF
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -174,12 +174,14 @@ func checkVolumes(t *testing.T, podSpec corev1.PodSpec) {
 func checkContainerCommand(t *testing.T, podSpec corev1.PodSpec) {
 	binBash := "/bin/bash"
 	dashC := "-c"
+	// This script should be present on every node.
+	// https://github.com/openshift/machine-config-operator/pull/2232
+	// The script sets the API server URL environment variables that
+	// the client SDK detects automatically.
 	setAPIEnv := `#!/bin/bash
 set -o allexport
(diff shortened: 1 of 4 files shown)


pkg/cloud/cloud_test.go
```diff
@@ -240,12 +240,14 @@ func checkVolumes(t *testing.T, podSpec corev1.PodSpec) {
 func checkContainerCommand(t *testing.T, podSpec corev1.PodSpec) {
 	binBash := "/bin/bash"
 	dashC := "-c"
+	// This script should be present on every node.
```
Fedosin, line 243:
```suggestion
	// This script should be present on every master node.
```

Danil-Grigorev: I wonder if it matters that masters and workers will communicate with API Server using different load balancers. Should it be documented somewhere? That if worker is running CNM replica (Azure) then it is connected to a different endpoint? @JoelSpeed It is mostly an implementational detail, but still.

JoelSpeed: We will need this to be every node for Azure, so we are fixing that on the MCO side already. This comment will be updated to reflect the changes and the differences once that is merged

JoelSpeed:

This PR serves no functional changes as we never observed the value we were setting here anyway, no real need for the upgrade to pass as this won't affect upgrades

lobziik:

We didn't ever handle this variable, shouldn't have any effect at all.

PR #89, opened 2021-07-12
[OCPCLOUD-1202] Allow CCMs to observe cluster wide proxy

Fedosin:
To conform with existing components within OpenShift, this commit allows the operator to watch for ClusterWideProxy configuration and inject the relevant environment variables into the containers for the CCMs.

This ensures that cloud provider traffic is routed via customers proxies as per the rest of the outbound traffic from the OpenShift components.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -221,6 +221,7 @@ func TestComposeConfig(t *testing.T) {
 		name          string
 		namespace     string
 		infra         *configv1.Infrastructure
+		clusterProxy  *configv1.Proxy
 		imagesContent string
 		expectConfig  OperatorConfig
 		expectError   string
@@ -331,6 +332,35 @@ func TestComposeConfig(t *testing.T) {
 			},
 		},
 		expectError: "platform status is not populated on infrastructure",
+	}, {
+		name:      "Unmarshal images from file for AWS",
+		namespace: defaultManagementNamespace,
+		infra: &configv1.Infrastructure{
+			Status: configv1.InfrastructureStatus{
+				PlatformStatus: &configv1.PlatformStatus{
+					Type: configv1.AWSPlatformType,
+				},
+			},
+		},
+		clusterProxy: &configv1.Proxy{
+			Status: configv1.ProxyStatus{
+				HTTPProxy: "http://squid.corp.acme.com:3128",
+			},
+		},
+		imagesContent: `{
+    "cloudControllerManagerAWS": "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
+    "cloudControllerManagerOpenStack": "registry.ci.openshift.org/openshift:openstack-cloud-controller-manager"
+}`,
+		expectConfig: OperatorConfig{
+			ControllerImage:  "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
+			ManagedNamespace: defaultManagementNamespace,
+			Platform:         configv1.AWSPlatformType,
+			ClusterProxy: &configv1.Proxy{
+				Status: configv1.ProxyStatus{
+					HTTPProxy: "http://squid.corp.acme.com:3128",
+				},
+			},
+		},
 	}}
 
 	for _, tc := range tc {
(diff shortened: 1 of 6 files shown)


pkg/config/config_test.go
```diff
@@ -285,6 +286,29 @@ func TestComposeConfig(t *testing.T) {
...
+		platform:  configv1.AWSPlatformType,
+		clusterProxy: &configv1.Proxy{
+			Status: configv1.ProxyStatus{
+				HTTPProxy: "http://squid.corp.acme.com:3128",
+			},
+		},
+		imagesContent: `{
+    "cloudControllerManagerAWS": "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
+    "cloudControllerManagerOpenStack": "registry.ci.openshift.org/openshift:openstack-cloud-controller-manager"
+}`,
```
elmiko, line 301: seems like the indent is a little off here?

pkg/substitution/substitution_test.go
```diff
@@ -116,7 +117,142 @@ func TestSetCloudControllerImage(t *testing.T) {
...
+				}},
+		}},
+		config: config.OperatorConfig{
+			ClusterProxy: &configv1.Proxy{
+				Status: configv1.ProxyStatus{
+					HTTPProxy: "http://squid.corp.acme.com:3128",
+				},
+			},
+		},
+	}, {
```
elmiko, line 191: should we have an explicit https test as well?

elmiko (commented):
this mostly makes good sense to me, just a comment and a question

pkg/config/config_test.go
Fedosin: It's in all places. https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/config/config_test.go#L37-L58
we can fix it in a separate commit.

pkg/substitution/substitution_test.go
Fedosin: done

elmiko (commented):
thanks Mike!
/lgtm

Fedosin:
Fixed a merge conflict

elmiko (commented):
/lgtm

pkg/controllers/clusteroperator_controller.go
```diff
@@ -97,6 +98,19 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	clusterProxy := &configv1.Proxy{}
+	if err := r.Get(ctx, client.ObjectKey{Name: proxyResourceName}, clusterProxy); errors.IsNotFound(err) {
+		clusterProxy = &configv1.Proxy{}
```
JoelSpeed, line 103: Why is this initialisation required? I think line 101 should be sufficient to ensure it's not a nil later, and `Get` will only modify the object if get succeeds, so I don't think this is actually required is it?

pkg/substitution/substitution.go
```diff
@@ -39,6 +40,51 @@ func setCloudControllerImage(config config.OperatorConfig, p corev1.PodSpec) cor
...
+// setProxySettings substitutes controller containers in provided pod specs with cluster wide proxy settings
+func setProxySettings(config config.OperatorConfig, p corev1.PodSpec) corev1.PodSpec {
+	clusterProxyEnvVars := getProxyArgs(config.ClusterProxy)
+	if len(clusterProxyEnvVars) == 0 {
+		return p
+	}
+
+	updatedPod := *p.DeepCopy()
+	for i, container := range p.Containers {
+		klog.Infof("Substituting proxy settings for container %q with %+v", container.Name, clusterProxyEnvVars)
```
JoelSpeed, line 52: Do we definitely want to log this information? Perhaps the specific details could go at a higher log level?

pkg/substitution/substitution_test.go
```diff
@@ -116,7 +117,196 @@ func TestSetCloudControllerImage(t *testing.T) {
...
+						},
+					},
+				},
+			}
+
+			spec := setProxySettings(tc.config, deploy.Spec.Template.Spec)
+			assert.EqualValues(t, spec.Containers, tc.expectedContainers)
+
+			spec = setProxySettings(tc.config, ds.Spec.Template.Spec)
+			assert.EqualValues(t, spec.Containers, tc.expectedContainers)
```
JoelSpeed, line 307: I don't think there's a need to have the deployment or daemonset in this, nothing specific about them is being observed here is it? I think we can just pass the podSpec and run this once

JoelSpeed (commented):
Looks good, a few nits below, PTAL

elmiko (commented):
/lgtm

PR #90, opened 2021-07-13
Check diff after generate and vendor tasks

Danil-Grigorev:
Add a simple script to check diff in generated/vendored files in CI.


Makefile
```diff
@@ -60,10 +60,12 @@ vendor:
 	go mod tidy
 	go mod vendor
 	go mod verify
+	./hack/verify-diff.sh
```
mdbooth, line 63: I don't think you want this here? This will cause `make vendor` to return an error if it updates anything. I think we only want this behaviour in the vendor test.

Makefile
```diff
@@ -60,10 +60,12 @@ vendor:
...
 
 # Generate code
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	./hack/verify-diff.sh
```
mdbooth, line 68: Ditto for generate.

mdbooth (changes requested):
The errors returned by vendor and generate don't feel right to me. Do you want an additional make target, perhaps, that runs verify-diff.sh? Then the test can be explicit and run:

```bash
make vendor verifydiff
```

Makefile
Danil-Grigorev: Updated the script to error out only in CI.

elmiko (commented):
/lgtm

Fedosin:
/close

This change is implemented here: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/124

PR #91, opened 2021-07-14
Substitution module cleanup

Fedosin:
Since we removed "render" code, we don't work with plain pods anymore. It means that FillConfigValues from "substitute" should work with Deployments and DaemonSets only.

This commit removes the Pod case from the function, and also updates the tests accordingly.

### DIFF
--- a/pkg/substitution/substitution_test.go
+++ b/pkg/substitution/substitution_test.go
@@ -81,39 +81,11 @@ func TestSetCloudControllerImage(t *testing.T) {
 
 	for _, tc := range tc {
 		t.Run(tc.name, func(t *testing.T) {
-			deploy := &v1.Deployment{
-				Spec: v1.DeploymentSpec{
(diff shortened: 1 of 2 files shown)


pkg/substitution/substitution_test.go
```diff
@@ -80,39 +80,13 @@ func TestSetCloudControllerImage(t *testing.T) {
...
-				Spec: v1.DaemonSetSpec{
-					Template: corev1.PodTemplateSpec{
-						Spec: corev1.PodSpec{
-							Containers: tc.containers,
-						},
-					},
-				},
-			}
-
 			pod := &corev1.Pod{
```
JoelSpeed, line 83: Why are we still using a pod? Should we not just initialise the pod spec? The pod wrapper here doesn't add anything

Fedosin: okay, why not...

PR #92, opened 2021-07-15
Fix fmt test in CI

Fedosin:
Currently we expect that "make fmt" just checks the code and returns an error when it's badly formatted. Unfortunately, it just fixes the code and always finishes successfully.

To fix if, this commit introduces a new script "verify-gofmt.sh" that is used in the original "fmt" target. The "fmt" itself is renamed
to "update-fmt" to comply with what it actually does.

### DIFF
--- a/hack/verify-gofmt.sh
+++ b/hack/verify-gofmt.sh
@@ -1,0 +1,20 @@
+
+#!/bin/bash
+
+function list_go_src_files() {
+	find . -not \( \
+		\( \
+		-wholename './.*' \
+		-o -wholename '*/vendor/*' \
+		\) -prune \
+	\) -name '*.go' | sort -u
+}
+
+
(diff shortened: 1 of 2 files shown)


hack/verify-gofmt.sh
```diff
@@ -0,0 +1,20 @@
...
+	find . -not \( \
+		\( \
+		-wholename './.*' \
+		-o -wholename '*/vendor/*' \
+		\) -prune \
+	\) -name '*.go' | sort -u
+}
+
+
+bad_files=$(list_go_src_files | xargs gofmt -s -l)
```
lobziik, line 14: thinking about replace `gofmt` with `goimports`, WDYT?

elmiko: i think if we are going to use the `verify-diff.sh` script from #90, then we should centralize all the logic there.

re: `goimports`, i haven't used it before but it seem nice to me

PR #94, opened 2021-07-19
support alibaba cloud ccm

gujingit:


vendor/github.com/openshift/api/config/v1/types_infrastructure.go
```diff
@@ -160,6 +160,9 @@ const (
 
 	// EquinixMetalPlatformType represents Equinix Metal infrastructure.
 	EquinixMetalPlatformType PlatformType = "EquinixMetal"
+
+	// AlibabaCloudPlatformType represents Alibaba Cloud infrastructure
+	AlibabaCloudPlatformType PlatformType = "AlibabaCloud"
```
JoelSpeed, line 165: Are there corresponding changes to the go.mod and go.sum file for this vendor update?

pkg/cloud/alibaba/alibaba_test.go
```diff
@@ -0,0 +1 @@
+package alibaba
```
JoelSpeed, line 1: Please add a test similar to https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/aws/aws_test.go

pkg/cloud/alibaba/assets/cloud-controller-manager-daemonset.yaml
```diff
@@ -0,0 +1,96 @@
...
+metadata:
+  name: cloud-config
+  namespace: kube-system
+data:
+  cloud-config.conf: |-
+    {
+        "Global": {
+            "routeTableIDs": ""
+        }
+    }
```
JoelSpeed, line 12: Cloud config is typically handled by a separate component and not CCCMO, I wouldn't expect this to be here

pkg/cloud/alibaba/assets/cloud-controller-manager-daemonset.yaml
```diff
@@ -0,0 +1,96 @@
...
+---
+apiVersion: apps/v1
+kind: DaemonSet
```
JoelSpeed, line 15: This should probably be a deployment to match the other platforms in this project. Please copy one of the other clouds and adjust as necessary for Ali Cloud

vendor/github.com/openshift/api/config/v1/types_infrastructure.go
kwoodson: github.com/openshift/api/pull/970 Is the dependent pull request. We are waiting for the branch to open. @gujingit We will need to update the go.mod when the upstream api changes merge.

pkg/cloud/alibaba/alibaba_test.go
kwoodson: @gujingit bump

vendor/github.com/openshift/api/config/v1/types_infrastructure.go
gujingit: ok. I will update the go.mod when the upstream api changes merge.

pkg/cloud/alibaba/alibaba_test.go
gujingit: ok, I will add it.

pkg/cloud/alibaba/alibaba_test.go
```diff
@@ -0,0 +1,24 @@
...
+	for _, r := range resources {
+		names = append(names, r.GetName())
+		kinds = append(kinds, r.GetObjectKind().GroupVersionKind().Kind)
+	}
+
+	assert.Contains(t, names, "alibaba-cloud-controller-manager")
+	assert.Contains(t, kinds, "Deployment")
+
+	assert.Contains(t, names, "alibaba-cloud-controller-manager-config")
+	assert.Contains(t, kinds, "ConfigMap")
```
JoelSpeed, line 22: Doesn't the installer normally create the cloud config? Is there WIP for that already? 

If they are creating it, then we have a mechanism in place already to copy this over and make it available to the CCM, so it wouldn't be needed here

pkg/cloud/cloud_test.go
```diff
@@ -84,6 +84,9 @@ func TestGetResources(t *testing.T) {
 	}, {
 		name:     "BareMetal resources are empty",
 		platform: configv1.BareMetalPlatformType,
+	}, {
+		name:     "AlibabaCloud resources are empty",
```
JoelSpeed, line 88: This test should look more like the AWS or Azure tests above as we actually have resources to compare with

pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,100 @@
...
+          operator: Exists
+          tolerationSeconds: 120
+        - effect: NoSchedule
+          key: node.cloudprovider.kubernetes.io/uninitialized
+          operator: Exists
+        - effect: NoSchedule
+          key: node.kubernetes.io/not-ready
+          operator: Exists
+      containers:
+        - command:
```
JoelSpeed, line 49: Please check the test results on this file as there are a number of requirements we set out via tests for the CCMs. The messages should hopefully be self explanatory, if they aren't, you can look to AWS as an example for what some of these might mean

pkg/cloud/alibaba/alibaba_test.go
gujingit: The installer does not create the cloud config, and does not plan to do so. OpenStack also contains a configmap. Can we add the cloud config here?

JoelSpeed: As far as I can tell the installer does still create the cloud config, based on this code https://github.com/openshift/installer/blob/af373729b31fcc6a5af9b6d0a708a44dc2ca1cb2/pkg/asset/manifests/cloudproviderconfig.go#L91-L204

Have you spoken with the installer team about this already? If so, let me know where so I can grab that extra context

patrickdillon: Looking at openshift/installer#5018: Alibaba is currently creating an empty cloud provider config: https://github.com/openshift/installer/blob/c8edf03a275d5025efaf375f3c04499681a61901/pkg/asset/manifests/alibabacloud/cloudproviderconfig.go

If there are values needed from the cloud provider config, they should be generated in the installer.

pkg/cloud/alibaba/alibaba.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+var (
+	//go:embed assets/*
+	aliFS embed.FS
+
+	aliResources []client.Object
+
+	aliSources = []common.ObjectSource{
+		{Object: &appsv1.Deployment{}, Path: "assets/cloud-controller-manager-daemonset.yaml"},
```
JoelSpeed, line 19: Just noticed this doesn't match the filename in the folder, needs to be deployment
```suggestion
		{Object: &appsv1.Deployment{}, Path: "assets/cloud-controller-manager-deployment.yaml"},
```

pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+              set -o allexport
+              if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+                source /etc/kubernetes/apiserver-url.env
+              fi
+              exec /cloud-controller-manager \
+              --kubeconfig=/etc/kubernetes/cloud-controller-manager.conf \
+              --address=127.0.0.1 \
+              --allow-untagged-cloud=true \
+              --leader-elect=true \
+              --cloud-provider=alicloud \
```
kwoodson, line 63: @elmiko @JoelSpeed Is this string here correct? `alicloud`? Should it be `AlibabaCloud` or `alibabacloud` or `""` or `external`?

elmiko: looks like `alicloud` is the correct one, see https://github.com/kubernetes/cloud-provider-alibaba-cloud/blob/master/docs/examples/cloud-controller-manager.yml#L165

pkg/cloud/alibaba/assets/config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
kwoodson, line 1: Delete this file. The config map exists and is created by the installer.

pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+        - hostPath:
+            path: /etc/kubernetes
+          name: k8s
+        - hostPath:
+            path: /etc/ssl/certs
+          name: certs
+        - hostPath:
+            path: /etc/pki
+          name: pki
+        - name: cloud-config
```
kwoodson, line 103: The name of the configmap created by the installer is `cloud-conf`. 
```
          configMap:
            name: cloud-conf
            items:
              - key: cloud.conf
                path: cloud-config.conf
```

pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+          configMap:
```
kwoodson, line 104: The name of the configmap created by the installer is cloud-conf.
```
          configMap:
            name: cloud-conf
            items:
              - key: cloud.conf
                path: cloud-config.conf
```

kwoodson, line 104: The name of the configmap created by the installer is cloud-conf.
```
          configMap:
            name: cloud-conf
            items:
              - key: cloud.conf
                path: cloud-config.conf
```

pkg/cloud/alibaba/assets/config.yaml
gujingit: ok, I will delete it.

pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
gujingit: got it.

gujingit: The configmap created by the installer in the ns 'openshift-config', but the ccm pod is in the ns 'openshift-cloud-controller-manager'. Can the configmap be mounted to the ccm pod?

bd233: @kwoodson  I noticed that the config part of the cloud-provider-config.yaml file generated by the installer is similar to cloud-conf. Is there any connection between the two?

JoelSpeed: This is part of the design of openshift, consumers of the configmap need to copy it to their own namespace. The CCCMO already handles this for you so you don't need to worry about it

bd233: Ok, thanks,  But I am not sure if this helps to answer the two questions of @gujingit  above.

> Is the name of the configmap created by the installer "cloud-provider-config" and the key is "config" ? see https://github.com/openshift/installer/blob/2e99509e39d0aad420b7893a2454e619bdc9cab9/pkg/asset/manifests/cloudproviderconfig.go#L38-L52

> configMap:
>             name: cloud-conf
>             items:
>               - key: cloud.conf
>                 path: cloud-config.conf

I think the name here should be "cloud-conf" and the key should be "cloud.conf" of the configmap. Is my idea correct?

> The configmap created by the installer in the ns 'openshift-config', but the ccm pod is in the ns 'openshift-cloud-controller-manager'. Can the configmap be mounted to the ccm pod?

@JoelSpeed @kwoodson  Can you help me answer this question?

JoelSpeed: > @JoelSpeed @kwoodson Can you help me answer this question?

This is what I was answering with the comment below:

> This is part of the design of openshift, consumers of the configmap need to copy it to their own namespace. The CCCMO already handles this for you so you don't need to worry about it

You can see we do this in Azure as an example https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/7d9a672c483f77ac788b1a52b944730889b347b4/pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml#L98-L103

kwoodson: @JoelSpeed Thanks for the example here. 

I had to go back and read the documentation on configmaps and their documentation wasn't crystal clear.

@gujingit @bd233 
The `configMap.name` is the name of the configmap and it should be `cloud-conf`.  This is the copy that exists in the `openshift-cloud-controller-manager` namespace. The one the operator does for you.
```
oc edit cm -n openshift-cloud-controller-manager  cloud-conf 
configmap/cloud-conf edited
```

The `- key:` references the `configMap.items[0].key` in the `data` dictionary section of the configmap. The configmap looks like this:
```
apiVersion: v1
data:
  cloud.conf: "{\n\t\"Global\": {\n\t\t\"kubernetesClusterTag\": \"\",\n\t\t\"nodeMonitorPeriod\":
    0,\n\t\t\"nodeAddrSyncPeriod\": 0,\n\t\t\"uid\": \"\",\n\t\t\"vpcid\": \"\",\n\t\t\"region\":
    \"us-east-1\",\n\t\t\"zoneid\": \"\",\n\t\t\"vswitchid\": \"\",\n\t\t\"clusterID\":
    \"test-h44cl\",\n\t\t\"routeTableIDs\": \"\",\n\t\t\"serviceBackendType\": \"\",\n\t\t\"disablePublicSLB\":
    false,\n\t\t\"accessKeyID\": \"<redacted>\",\n\t\t\"accessKeySecret\":
    \"<redacted>\"\n\t}\n}\n"
kind: ConfigMap
metadata:
  creationTimestamp: "2021-09-02T16:46:04Z"
  name: cloud-conf
  resourceVersion: "56192"
  uid: 47124e85-d887-4c70-b081-22db7cfbae31
```

The last `configMap.items[0].path` represents the mount path that exists on the filesystem. This value is what the file will show up as on the filesystem.  This value `--cloud-config=/etc/kubernetes/config/cloud-config.conf` from [here]( https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/94/files/b71a2f190d38485031e011e2e349aa06f9e9f9dc#diff-3385f0f98dcbbb6f0c5813829f456118119e9f5bcf8c76f8867b22fba9a725d6R65). The **`cloud-config.conf`** section is the `path`. As long as the values from the `--cloud-config` parameter to the `/cloud-controller-manager` container match this `path` then it should line up. 

Please follow the above information as well as the Azure example and I will test it tomorrow and verify it is all working properly.

kwoodson:
@bd233 @gujingit Please rebase this PR to resolve the conflicts.

kwoodson:
@bd233 @gujingit Please also update the go.mod with the latest `openshift/api` updates. This should include the AliabaCloud platform type.
```
go get -u github.com/openshift/api@master
go mod tidy && go mod vendor && go mod verify
git add go.mod go.sum vendor && git commit -C --amend && git push -f
```

PR #95, opened 2021-07-19
Fix container names in azure manifests

lobziik:


elmiko (approved):
thanks Denis
/approve

Fedosin (approved):
/lgtm

Fedosin:
Do we plan to update other platforms accordingly? 
AWS: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/aws/assets/deployment.yaml#L38
OpenStack: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/openstack/assets/deployment.yaml#L45

lobziik:
@Fedosin No. https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/substitution/substitution.go#L17 Needed for azure only

PR #96, opened 2021-07-19
Azure Stack separate assets

lobziik:
There is plenty of changes between public Azure and Stack one. With current substitution approach it is not convenient to handle it. Extracted to separate module.

### DIFF
--- a/pkg/cloud/azurestack/azurestack.go
+++ b/pkg/cloud/azurestack/azurestack.go
@@ -27,7 +27,7 @@ func init() {
 	utilruntime.Must(err)
 }
 
-// GetResources returns a list of AWS resources for provisioning CCM in running cluster
+// GetResources returns a list of Azrue Stack Hub resources for provisioning CCM in running cluster
 func GetResources() []client.Object {
 	resources := make([]client.Object, len(azureResources))
 	for i := range azureResources {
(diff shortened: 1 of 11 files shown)


JoelSpeed (commented):
I think we can simplify this a bit, please see my comments below

pkg/cloud/azurestack/azurestack.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+)
+
+func init() {
+	var err error
+	azureResources, err = common.ReadResources(azureFs, azureSources)
+	utilruntime.Must(err)
+}
+
+// GetResources returns a list of AWS resources for provisioning CCM in running cluster
```
alexander-demichev, line 30:
```suggestion
// GetResources returns a list of Azure resources for provisioning CCM in running cluster
```

Fedosin (approved):
/lgtm

JoelSpeed:
We are aware that the tests are not currently passing because of a relatively major bug in the Azure CCM when running on OpenShift (this is to do with how openshift is deployed on Azure). We believe we have a working solution, and at this point believe this PR is correct.

We will continue to work on the bugs over the next few days and hope to have the Azure/ASH CCM stable by the end of next week.

sdodson:
Accepting this feature PR after feature freeze because this is the last critical piece in delivering an urgent feature. The work was done but delayed due to a strong desire to test more thoroughly and coordinate with other teams. Overriding bugzilla/valid-bug label.

PR #97, opened 2021-07-19
Add initial ibm pkg assets and embed files

shalver:
Issue: https://github.ibm.com/alchemy-containers/armada-update/issues/2136

### DIFF
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/aws"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/azure"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/azurestack"
+	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/ibm"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/openstack"
 	"github.com/stretchr/testify/assert"
 	appsv1 "k8s.io/api/apps/v1"
@@ -73,8 +74,9 @@ func TestGetResources(t *testing.T) {
 		name:     "OVirt resources are empty, as the platform is not yet supported",
 		platform: configv1.OvirtPlatformType,
 	}, {
-		name:     "IBMCloud resources are empty, as the platform is not yet supported",
+		name:     "IBMCloud resources returned as expected",
 		platform: configv1.IBMCloudPlatformType,
+		expected: ibm.GetResources(),
 	}, {
 		name:     "Libvirt resources are empty",
 		platform: configv1.LibvirtPlatformType,
--- a/pkg/cloud/ibm/assets/deployment.yaml
+++ b/pkg/cloud/ibm/assets/deployment.yaml
@@ -1,0 +1,126 @@
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: ibm-cloud-controller-manager
+  namespace: openshift-cloud-controller-manager
+  labels:
+    k8s-app: ibm-cloud-controller-manager
+spec:
+  replicas: 2
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      k8s-app: ibm-cloud-controller-manager
+  template:
+    metadata:
+      labels:
+        k8s-app: ibm-cloud-controller-manager
+    spec:
+      priorityClassName: system-cluster-critical
+      hostNetwork: true
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - topologyKey: "kubernetes.io/hostname"
+            labelSelector:
+              matchLabels:
+                k8s-app: ibm-cloud-controller-manager
+      tolerations:
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/master
+        operator: Exists
+      - effect: NoExecute
+        key: node.kubernetes.io/unreachable
+        operator: Exists
+        tolerationSeconds: 120
+      - effect: NoExecute
+        key: node.kubernetes.io/not-ready
+        operator: Exists
+        tolerationSeconds: 120
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        operator: Exists
+      - effect: NoSchedule
+        key: node.kubernetes.io/not-ready
+        operator: Exists
+      serviceAccountName: cloud-controller-manager
+      terminationGracePeriodSeconds: 90
+      containers:
+      # Container is required to specify name in order for image substitution
+      # to take place. Cloud-controller-manager is substituted with
+      # real CCM image from the cluster
+      - name: cloud-controller-manager
+        image: quay.io/openshift/origin-ibm-cloud-controller-manager
+        imagePullPolicy: IfNotPresent
+        env:
+          - name: POD_IP_ADDRESS
+            valueFrom:
+              fieldRef:
+                fieldPath: status.podIP
+          - name: VPCCTL_CLOUD_CONFIG
+            value: /etc/ibm/cloud.conf
+        command:
+          - /bin/bash
+          - -c
+          - |
+            #!/bin/bash
+            set -o allexport
+            if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+              source /etc/kubernetes/apiserver-url.env
+            fi
+            exec /bin/ibm-cloud-controller-manager \
+            --port=0 \
+            --bind-address=$(POD_IP_ADDRESS) \
+            --use-service-account-credentials=true \
+            --cloud-provider=ibm \
+            --cloud-config=/etc/ibm/cloud.conf \
+            --feature-gates=IPv6DualStack=false \
+            --profiling=false \
+            --leader-elect=true \
+            --leader-elect-lease-duration=137s \
+            --leader-elect-renew-deadline=107s \
+            --leader-elect-retry-period=26s \
+            --leader-elect-resource-namespace=openshift-cloud-controller-manager \
+            --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_AES_128_GCM_SHA256,TLS_CHACHA20_POLY1305_SHA256,TLS_AES_256_GCM_SHA384 \
+            --v=2
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 10258
+            scheme: HTTPS
+          initialDelaySeconds: 300
+          timeoutSeconds: 160
+        ports:
+          - containerPort: 10258
+            name: https
+            protocol: TCP
+        resources:
+          # Container is required to set pod requests, but not limits to satisfy CI 
+          # requirements for OpenShift on every scheduled workload in the cluster
+          requests:
+            cpu: 75m
+            memory: 60Mi
+        volumeMounts:
+          - mountPath: /etc/kubernetes
+            name: host-etc-kube
+            readOnly: true
+          - mountPath: /etc/ibm
+            name: cloud-conf
+          - mountPath: /etc/vpc
+            name: ibm-cloud-credentials
+      volumes:
+        - name: host-etc-kube
+          hostPath:
+            path: /etc/kubernetes
+            type: Directory
+        - configMap:
+            defaultMode: 420
+            name: cloud-conf
+          name: cloud-conf
+        - secret:
+            defaultMode: 420
+            secretName: ibm-cloud-credentials
+          name: ibm-cloud-credentials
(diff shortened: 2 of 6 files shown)


JoelSpeed:
IIUC, we need to have an entry in the images map to tell CCCMO which image needs to be substituted. I don't think this will work until that is added.

Have you tried to test this manually yet?

Once you've added that image reference, we won't be able to merge this until the IBM CCM image is included in the OCP release payload, to prevent accidental merges, I'm going to hold this until that process has been done. 

/hold

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+          - /bin/bash
+          - -c
+          - |
+            #!/bin/bash
+            set -o allexport
+            if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+              source /etc/kubernetes/apiserver-url.env
+            fi
+            exec /bin/ibm-cloud-controller-manager \
+            --port=0 \
```
Fedosin, line 77: We also need to add `--leader-elect-resource-namespace=openshift-cloud-controller-manager` because by default it uses `kube-system` namespace.

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+        command:
+          - /bin/bash
+          - -c
+          - |
+            #!/bin/bash
+            set -o allexport
+            if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+              source /etc/kubernetes/apiserver-url.env
+            fi
+            exec /bin/ibm-cloud-controller-manager \
```
rtheis, line 76: Please set `--tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_AES_128_GCM_SHA256,TLS_CHACHA20_POLY1305_SHA256,TLS_AES_256_GCM_SHA384` to use TLS 1.3 ciphers.

rtheis (changes requested):

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+            --port=0 \
+            --use-service-account-credentials=true \
+            --kubeconfig=/etc/kubernetes/kubeconfig \
+            --authentication-kubeconfig=/etc/kubernetes/kubeconfig \
+            --authorization-kubeconfig=/etc/kubernetes/kubeconfig \
+            --client-ca-file=/etc/kubernetes/ca.crt \
+            --requestheader-client-ca-file=/etc/kubernetes/ca.crt \
+            --cloud-provider=ibm \
+            --cloud-config=/etc/kubernetes/cloud.conf \
+            --feature-gates=LegacyNodeRoleBehavior=false,IPv6DualStack=false \
```
rtheis, line 86: `LegacyNodeRoleBehavior=false` is no longer a feature gate for our 1.22 version of the CCM.  Please remove it.

rtheis (changes requested):

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+        key: node.kubernetes.io/not-ready
+        operator: Exists
+        tolerationSeconds: 120
+      - effect: NoSchedule
+        key: node.cloudprovider.kubernetes.io/uninitialized
+        operator: Exists
+      - effect: NoSchedule
+        key: node.kubernetes.io/not-ready
+        operator: Exists
+      serviceAccountName: cloud-controller-manager
```
rtheis, line 52: The default RBAC for this service account is not sufficient for the IBM CCM.  We'll either need a base RBAC update or a custom addition for the IBM CCM.  I assume that the later will be preferred.

rtheis (changes requested):

pkg/cloud/ibm/assets/deployment.yaml
rtheis: Looks like we can update similar to https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/81

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+            --client-ca-file=/etc/kubernetes/ca.crt \
+            --requestheader-client-ca-file=/etc/kubernetes/ca.crt \
+            --cloud-provider=ibm \
+            --cloud-config=/etc/kubernetes/cloud.conf \
+            --feature-gates=LegacyNodeRoleBehavior=false,IPv6DualStack=false \
+            --profiling=false \
+            --leader-elect=true \
+            --leader-elect-lease-duration=25s \
+            --leader-elect-renew-deadline=15s \
+            --leader-elect-retry-period=3s \
```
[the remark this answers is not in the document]
shalver: updated

pkg/cloud/ibm/assets/deployment.yaml
shalver: updated

pkg/cloud/ibm/assets/deployment.yaml
shalver: updated

pkg/cloud/ibm/assets/deployment.yaml
shalver: updated

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+          - |
+            #!/bin/bash
+            set -o allexport
+            if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+              source /etc/kubernetes/apiserver-url.env
+            fi
+            exec /bin/ibm-cloud-controller-manager \
+            --port=0 \
+            --use-service-account-credentials=true \
+            --kubeconfig=/etc/kubernetes/kubeconfig \
```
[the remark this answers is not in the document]
shalver: updated

JoelSpeed:
Are we happy that this is working now? Seems ok from what I can tell, though I haven't run it.

If we need to update the RBAC, maybe we should just include that in another commit on this PR?

rtheis (approved):
/lgtm

rtheis:
@JoelSpeed I am happy with this PR as-is.  Once we hack together the needed configurations in our test cluster, this is working as we expect.  As for the RBAC, we don't have a full list yet.

JoelSpeed:
/hold cancel
/lgtm

I believe this is safe to merge

JoelSpeed:

We know Azure is broken right now, no need to keep retesting

pkg/cloud/cloud_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/aws"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/azure"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/azurestack"
+  "github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/ibm"
```
JoelSpeed, line 11: Nit: Indentation is off here

shalver: fixed

pkg/cloud/ibm/assets/deployment.yaml
JoelSpeed: Why port 0? Default port is 10258, which is reserved for CCMs.
You will need to include a port stanza in the pod spec as well to make the verify task happy

pkg/cloud/ibm/assets/deployment.yaml
rtheis: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/100 should fix

pkg/cloud/ibm/assets/deployment.yaml
rtheis: @JoelSpeed I believe `--port=0` is disabling the insecure port.  Eventually this option will be removed from Kubernetes for all controllers.  The tentative removal is scheduled for v1.24.

JoelSpeed: Good point, well made! I forgot they have a separate secure port flag, ignore me

> You will need to include a port stanza in the pod spec as well to make the verify task happy

This is still relevant though

shalver: port stanza added to container

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,132 @@
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: ibm-cloud-controller-manager
+  namespace: openshift-cloud-controller-manager
+  labels:
+    k8s-app: ibm-cloud-controller-manager
+spec:
+  replicas: 3
```
JoelSpeed, line 9: We use 2 replicas on other clouds, might be good to be consistent 🤔 

(Should probably add another test case for this one)

JoelSpeed (commented):
I think this is good to go pretty much, one final thing about replicas, though it's not super important/can be fixed later if need be

pkg/cloud/ibm/assets/deployment.yaml
shalver: replicas updated

JoelSpeed:
/lgtm

Thanks @shalver

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,132 @@
...
+  namespace: openshift-cloud-controller-manager
+  labels:
+    k8s-app: ibm-cloud-controller-manager
+spec:
+  replicas: 2
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
+      maxSurge: 3
```
rtheis, line 14: Max surge of 3 doesn't make sense if you can only have 2 replicas.  This should be at most 2 now.

shalver: updated

rtheis (approved):
Thanks Steve.

pkg/cloud/ibm/assets/deployment.yaml
```diff
@@ -0,0 +1,132 @@
...
+      serviceAccountName: cloud-controller-manager
+      terminationGracePeriodSeconds: 90
+      containers:
+      # Container is required to specify name in order for image substitution
+      # to take place. Cloud-controller-manager is substituted with
+      # real CCM image from the cluster
+      - name: cloud-controller-manager
+        image: quay.io/openshift/origin-ibm-cloud-controller-manager
+        imagePullPolicy: IfNotPresent
+        securityContext:
```
elmiko, line 61: i think we will need to drop this securityContext. the other clouds do not have this and @cjschaef is seeing some root-related issues when running the ccm.

cjschaef: I'm will be testing using the [IBM CCM Dockerfile.openshift](https://github.com/openshift/cloud-provider-ibm/blob/master/openshift-hack/images/Dockerfile.openshift) and not limiting the container to `runAsNonRoot` to check if things work as expected.

elmiko: i have a feeling that the `runAsNonRoot` might be causing issues because we do not specify a `runAsUser`. the product docs say this:

> MustRunAsNonRoot - Requires that the pod be submitted with a non-zero runAsUser or have the USER directive defined in the image. No default provided.

ref: https://docs.openshift.com/container-platform/4.8/authentication/managing-security-context-constraints.html#authorization-SCC-strategies_configuring-internal-oauth

elmiko: also, looking at the image layers it appears that we only define user `0`. so, i imagine this context is forcing the check, which then fails. i'm still digging to learn a little more about how we specify the user.

shalver: securityContext removed from deployment.yaml

elmiko (commented):
thanks for all the hard work everyone
/lgtm

JoelSpeed:
/test verify

I think this will fail now as there are changes to the base branch since this was pushed

JoelSpeed:
Yep it failed:
```
cloud_test.go:365: Deployment should set strategy type to "Recreate"
```
Please rebase on the latest changes and make sure this is included.

Please also take a look at https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/114 as this will affect IBM too and will likely merge first

elmiko (commented):
thanks for the update @shalver 
/lgtm

JoelSpeed:
Thanks @shalver, this should be ready to merge as soon as the branch opens up on monday, I'll remove the hold when I get the email through that confirms it's opened

shalver:
@elmiko @JoelSpeed    apologies for another push,  somehow had lost the volumes and volume mounts in deployment.yaml associated with the CredentialsRequest

rtheis (commented):
@shalver nit: the updated files are missing new lines at the end of them.  I'm not sure if this is going to cause problems for linting checks.

JoelSpeed:
> nit: the updated files are missing new lines at the end of them.

May not break the linter, but it hurts my OCD 😉  Can we get that fixed up, else it can cause weird git diff issues down the line

shalver:
new lines addded to deployment.yaml and 0000_26_cloud-controller-manager-operator_15_credentialsrequest-ibm.yaml

JoelSpeed (commented):
/lgtm

rtheis (approved):
/lgtm

JoelSpeed:
/hold cancel

The master branch is now open for 4.10 development. Having reviewed this PR, we do not believe this will cause any breakage to the existing platforms, as such, I think this is good to go!

Thanks for all the work on this from the IBM side

PR #98, opened 2021-07-20
Use klog v2 only

Fedosin:
Now we combine klog v1 and v2 in our code, which is not desired.

This commit deletes all v1 uses from the code and removes the library from dependencies.


elmiko (approved):
/approve

PR #101, opened 2021-07-23
Bug 1985366: Use registered ports and ensure that ports are defined in pod specs

JoelSpeed:
Per the guidelines in https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md, we should only be using defined ports when running on host network.

This implements the requirements of that guide based off of the ports registered in https://github.com/openshift/enhancements/pull/847

/hold until the registered ports are agreed upon

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
@@ -67,8 +67,15 @@ spec:
             fi
             exec /cloud-config-sync-controller \
             --leader-elect \
-            --metrics-bind-address=:8081 \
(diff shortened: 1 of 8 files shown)


JoelSpeed:
/hold cancel

Enhancement PR has merged, these port numbers are now official

Fedosin (approved):
/approve
/hold
/lgtm

Fedosin:
hold is there until tests are green

JoelSpeed:

I've checked the test output here and I don't believe this PR is causing the issues. We have known problems with Azure right now which we are already working on. As the AWS CCM tests passed, I'm pretty confident this is working as expected

PR #104, opened 2021-08-10
Bug 1993087: cloud-config Azure credentials injector

lobziik:
This pr introduce:
- azure-config-credentials-injector component within CCCMO image for injecting azure credenials to cloud config from secret
    - description docs for it
    - tests
    - Cobra library dependency
- CredentialsRequest manifest for Azure platform
- CCCMO image was included to images.json configmap for using it in initContainer for Azure Stack CCM/CNM
- InitContainer substitution for `azure-credentials-inject ` container

Need to be merged before https://github.com/openshift/installer/pull/5138. Tested with installer built from this pr.


cmd/azure-config-credentials-injector/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+	preparedCloudConfig, err := prepareCloudConfig(cloudConfig, azureClientId, azureClientSecret)
+	if err != nil {
+		return err
+	}
+
+	outputPath := args[1]
+	if err := writeCloudConfig(outputPath, preparedCloudConfig); err != nil {
+		return err
+	}
```
elmiko, line 63: i'm curious about these failure conditions, do we not need a log message?

elmiko (commented):
this generally looks good to me, i'm curious about the error logs for the config injector, and i'm also wondering if we should add a unit test to exercise the injection mechanics (essentially testing `mergeCloudConfig`), and also we should add a doc to describe what this new binary does and how it fits with the init container.

elmiko (commented):
the tests and doc updates look really nice @lobziik , what more did you want to add before removing WIP status ?

cmd/azure-config-credentials-injector/main.go
lobziik: It would fail in case if file path is not valid (e.g. directory), permissions issues and so on. Error would be propagated to toplevel and printed.

elmiko: got it, thanks!

manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
 data:
   images.json: >
     {
+      "cloudControllerManagerOperator": "quay.io/openshift/origin-cluster-cloud-controller-manager-operator",
```
JoelSpeed, line 12: Let's hope we never have a cloud provider call operator 😉

pkg/config/config_test.go
```diff
@@ -20,10 +20,12 @@ func TestGetImagesFromJSONFile(t *testing.T) {
 		name: "Unmarshal images from file",
 		path: "images_file",
 		imagesContent: `{
-    "cloudControllerManagerAWS": "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
-    "cloudControllerManagerOpenStack": "registry.ci.openshift.org/openshift:openstack-cloud-controller-manager"
-}`,
+			"cloudControllerManagerOperator": "quay.io/openshift/origin-cluster-cloud-controller-manager-operator",
+    		"cloudControllerManagerAWS": "registry.ci.openshift.org/openshift:aws-cloud-controller-manager",
```
JoelSpeed, line 24: Should we make sure the pattern matches the others? Use the registry.ci tag instead?

pkg/controllers/fixtures/images.json
```diff
@@ -1,4 +1,5 @@
 {
+    "cloudControllerManagerOperator": "quay.io/openshift/origin-cluster-cloud-controller-manager-operator",
```
JoelSpeed, line 2: Should we make sure the pattern matches the others? Use the registry.ci tag instead?

cmd/azure-config-credentials-injector/credentials_injector_test.go
```diff
@@ -0,0 +1,150 @@
+package main
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"os"
+	"testing"
```
JoelSpeed, line 8: Nit: standard library imports should come first, with a gap, then the additional imports
```suggestion
	"io/ioutil"
	"os"
	"testing"
	
	"github.com/spf13/cobra"
	"github.com/stretchr/testify/assert"
```

cmd/azure-config-credentials-injector/credentials_injector_test.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+func Test_mergeCloudConfig(t *testing.T) {
+
+	tmpDir, err := ioutil.TempDir("", "cccmo-azure-creds-injector")
+	if err != nil {
+		t.Fatal(err)
+	}
```
JoelSpeed, line 16: You could use `require.NoError(t, err)`, require is the same as assert except it calls fatal under the hood instead of error.

cmd/azure-config-credentials-injector/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+	clientIDCloudConfigKey     = "aadClientId"
+	clientSecretCloudConfigKey = "aadClientSecret"
+)
+
+func main() {
+	rootCmd := &cobra.Command{
+		Use:   "azure-config-credentials-injector [cloud-config-file-path] [output-file-path]",
+		Short: "Cloud config credentials injection tool for azure cloud platform",
+		Args:  cobra.ExactArgs(2),
```
JoelSpeed, line 25: Wondering if it's worth using named arguments rather than positional arguments, any thoughts?

cmd/azure-config-credentials-injector/main.go
JoelSpeed: Could you expand the errors here using `fmt.Errorf` to provide additonal detail about what exactly is failing, will almost certainly make it easier to find the issue when you can tie an individual error message to a particular if statement

JoelSpeed (commented):
As a suggestion for simplifying the test suite, have you looked at `github.com/testify/require`, it's basically the same as assert, can replace a lot of the `if err != nil` blocks you have here.

Apart from the test improvement suggestions, and improving error messages, I think this looks good to go.

Very much appreciate the readme you added, thank you for that!

JoelSpeed:
/close

We will continue this work in #106

PR #110, opened 2021-08-30
OCPCLOUD-1189: Move cloud provider specific logic behind interface

lobziik:
Quite big refactoring with the aim of encapsulate provider specific logic
within respective provider package.

This introduce bunch of changes:
- asaskevich/govalidator added to project dependencies
- substitution module was removed, part of it was moved to pkg/cloud/common package
- _GetResources_ from _pkg/cloud_ now expect whole _OperatorConfig_, not only _PlatformStatus_
- Specific images references was removed from _OperatorConfig_ and replaced with generic
_ImagesReference_. Now its provider package responsibility to know and pick needed
images from there.
- _CloudProviderAssets_ and _TemplateBasedAssets_ interfaces was introduced, all providers
was refactored in order to use it.
- _text/template_ usage was introduced in every provider package, switch-case based
substitution was removed
- validation for provider specific imageRefereces and templateValues was added
into each cloud provider package
- some tests was fixed

UPD: 
_TemplateBasedAssets_ was removed based on discussion within this PR.  https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/110/commits/7a3f69f86b2b89e275bcce323508d3eea54eedaa


lobziik:
сс @elmiko @Fedosin PTAL :)

pkg/config/config.go
```diff
@@ -45,48 +46,22 @@ func GetProviderFromInfrastructure(infra *configv1.Infrastructure) (configv1.Pla
...
-	case configv1.AzurePlatformType:
-		return images.CloudNodeManagerAzure
-	default:
-		return ""
-	}
-}
-
 // ComposeConfig creates a Config for operator
 func ComposeConfig(infrastructure *configv1.Infrastructure, clusterProxy *configv1.Proxy, imagesFile, managedNamespace string) (OperatorConfig, error) {
-	platform, err := GetProviderFromInfrastructure(infrastructure)
```
lobziik, line 89: From chat conversation with @Fedosin: need to rewrite this function, in order to make it's purpose more clean. Currently it serves for `infrastructure` object validation purposes before further usage.

JoelSpeed (commented):
I've only left comments on the first of the providers, but they apply across the board

I'm wondering if we can simplify the proxy args injection as well by requiring that the deployment yamls have a 
```
env:{{ .proxyArgs }}
```
Which we then substitute as appropriate?

pkg/cloud/common/substitution.go
```diff
@@ -0,0 +1,79 @@
...
+			if config.IsSingleReplica {
+				obj.Spec.Replicas = pointer.Int32(1)
+			}
+		case *appsv1.DaemonSet:
+			obj.Spec.Template.Spec = setProxySettings(config, obj.Spec.Template.Spec)
+		}
+		substitutedObjects[i] = templateCopy
+	}
+	return substitutedObjects
+}
```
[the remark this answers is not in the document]
lobziik: >I'm wondering if we can simplify the proxy args injection as well by requiring that the deployment yamls have a
>
>env:{{ .proxyArgs }}
>
>Which we then substitute as appropriate?

My original thought was that we want to enforce proxy and single replica settings to all containers/pods which managing by CCCMO.  Here it's a single point for doing such things. We could refactor it, but this approach seemd to me bit more convenient.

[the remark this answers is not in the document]
JoelSpeed: Ack, fair enough, SGTM

elmiko (commented):
i think this looks good @lobziik , nice improvements.
/lgtm

it would be cool if we had a "fake" provider in the future so we could have an example of the minimum necessary changes to add a new provider. maybe something for the future

pkg/cloud/common/interfaces.go
```diff
@@ -0,0 +1,25 @@
...
+// TemplateBasedAssets interface for specify assets object which leveraging golang standard template/text library
+// under the hood.
+type TemplateBasedAssets interface {
+	CloudProviderAssets
+	// GetTemplates returns list of templates for further rendering.
+	GetTemplates() []ObjectTemplate
+	// GetTemplateValues returns map with values for further templates rendering.
+	// This map is global for all templates within assets object and intended to be applied to each ObjectTemplate
+	// for single CloudProviderAssets
+	GetTemplateValues() (TemplateValues, error)
```
JoelSpeed, line 24: Where are these used? Don't we just call `GetRenderedResources` and cut out the middle man?

pkg/cloud/common/templates.go
```diff
@@ -0,0 +1,79 @@
...
+
+	err = yaml.UnmarshalStrict(buf.Bytes(), object)
+	if err != nil {
+		klog.Errorf("Cannot decode data from embedded resource %v: %v", tmpl.EmbedFsPath, err)
+		return nil, err
+	}
+	return object.(client.Object), nil
+}
+
+func RenderTemplates(provider TemplateBasedAssets) ([]client.Object, error) {
```
JoelSpeed, line 46: Do we need to use an interface in here? Why not just pass through the two types that we need? Might make testing easier as you then don't need to mock this interface

JoelSpeed (commented):
I think we can simplify some of this by having all of the actions that might cause an error happen during the construction phase. And given that, do we actually need to have a struct and interface at all? Each provider could just have one function that returns the rendered assets right?

hack/unit-tests.sh
```diff
@@ -52,7 +52,7 @@ function runTests() {
     if [ "$OPENSHIFT_CI" == "true" ]; then
         runTestCI
     else
-        go_test
+        go_test -test.short
```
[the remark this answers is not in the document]
lobziik: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/110/commits/75ec1e4133a9d7574bad0076ba66fbe964a7ac21

Put short there for disable bit heavy test during local runs.

pkg/cloud/aws/aws.go
```diff
@@ -2,36 +2,89 @@ package aws
...
+	return renderedObjects, nil
+}
+
+func (a *awsAssets) GetTemplates() []common.ObjectTemplate {
+	return a.templates
+}
+
+func (a *awsAssets) GetTemplateValues() (common.TemplateValues, error) {
+	return a.templateValues, nil
+}
```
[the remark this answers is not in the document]
lobziik: Leave it there intentionally, for be able to implement somewhat logic for construct values for templates in this method.

pkg/cloud/common/interfaces.go
lobziik: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/110/files#diff-5e1ede80f1538a3bc199f37052f71e420c9aa8505982c96ad0119f829ab67da6R28

Idea behind this method - encapsulate knowledge about needed variables inside `Assets` object and have a proper place to construct such map. 

Hm, looking on your comment on `RenderTemplate` function makes me think that this interface might be rethinked a bit.

pkg/cloud/aws/aws.go
```diff
@@ -2,36 +2,89 @@ package aws
...
+	templateValues common.TemplateValues
+}
+
+func (a *awsAssets) GetRenderedResources() ([]client.Object, error) {
+	renderedObjects, err := common.RenderTemplates(a)
+	if err != nil {
+		return nil, err
+	}
+	return renderedObjects, nil
+}
```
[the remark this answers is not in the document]
lobziik: >Doesn't look to me like a can be changed once its created can it?

pkg/cloud/aws/aws.go
```diff
@@ -2,36 +2,85 @@ package aws
...
+func (a awsAssets) GetOperatorConfig() config.OperatorConfig {
+	return a.operatorConfig
+}
+
+func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets, error) {
+	images := &imagesReference{
+		CloudControllerManager: config.ImagesReference.CloudControllerManagerAWS,
+	}
+
+	_, err := govalidator.ValidateStruct(images)
```
alexander-demichev, line 73:
```suggestion 
if _, err := govalidator.ValidateStruct(images); err != nil {
```

pkg/cloud/azure/azure.go
```diff
@@ -2,37 +2,89 @@ package azure
...
-	azureSources = []common.ObjectSource{
-		{Object: &appsv1.DaemonSet{}, Path: "assets/cloud-node-manager-daemonset.yaml"},
-		{Object: &appsv1.Deployment{}, Path: "assets/cloud-controller-manager-deployment.yaml"},
+func (assets *azureAssets) GetTemplateValues() (common.TemplateValues, error) {
+	values := common.TemplateValues{
+		"images":             assets.images,
+		"infrastructureName": assets.operatorConfig.InfrastructureName,
 	}
-)
+	_, err := govalidator.ValidateMap(values, templateValuesValidationMap)
```
alexander-demichev, line 56:
```suggestion 
	if _, err := govalidator.ValidateMap(values, templateValuesValidationMap); err != nil {
```

pkg/cloud/azurestack/azurestack.go
```diff
@@ -2,37 +2,91 @@ package azurestack
...
-	azureSources = []common.ObjectSource{
-		{Object: &appsv1.DaemonSet{}, Path: "assets/cloud-node-manager-daemonset.yaml"},
-		{Object: &appsv1.Deployment{}, Path: "assets/cloud-controller-manager-deployment.yaml"},
+func (assets *azurestackAssets) GetTemplateValues() (common.TemplateValues, error) {
+	values := common.TemplateValues{
+		"images":             assets.images,
+		"infrastructureName": assets.operatorConfig.InfrastructureName,
 	}
-)
+	_, err := govalidator.ValidateMap(values, templateValuesValidationMap)
```
alexander-demichev, line 57:
```suggestion 
	if _, err := govalidator.ValidateMap(values, templateValuesValidationMap); err != nil {
```

pkg/cloud/azurestack/azurestack.go
```diff
@@ -2,37 +2,91 @@ package azurestack
...
-	resources := make([]client.Object, len(azureResources))
-	for i := range azureResources {
-		resources[i] = azureResources[i].DeepCopyObject().(client.Object)
+func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets, error) {
+	images := &imagesReference{
+		Operator:               config.ImagesReference.CloudControllerManagerOperator,
+		CloudControllerManager: config.ImagesReference.CloudControllerManagerAzure,
+		CloudNodeManager:       config.ImagesReference.CloudNodeManagerAzure,
+	}
+	_, err := govalidator.ValidateStruct(images)
```
alexander-demichev, line 78:
```suggestion 
	if _, err := govalidator.ValidateStruct(images); err != nil {
```

pkg/cloud/common/templates.go
```diff
@@ -0,0 +1,79 @@
...
+	templateContent *template.Template
+}
+
+func (tmpl *ObjectTemplate) render(providerAssets TemplateBasedAssets) (client.Object, error) {
+	buf := &bytes.Buffer{}
+	templateValues, err := providerAssets.GetTemplateValues()
+	if err != nil {
+		return nil, fmt.Errorf("can not get values for template rendering: %s", err)
+	}
+	err = tmpl.templateContent.Execute(buf, templateValues)
```
alexander-demichev, line 32:
```suggestion 
	if err := tmpl.templateContent.Execute(buf, templateValues); err != nil {
```

pkg/cloud/common/templates.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("can not render template: %s", err)
+	}
+	object := tmpl.ReferenceObject.DeepCopyObject()
+
+	err = yaml.UnmarshalStrict(buf.Bytes(), object)
```
alexander-demichev, line 38:
```suggestion 
	if err = yaml.UnmarshalStrict(buf.Bytes(), object); err != nil {
```

pkg/cloud/openstack/openstack.go
```diff
@@ -2,36 +2,86 @@ package openstack
...
-// GetResources returns a list of OpenStack resources for provisioning CCM in running cluster
-func GetResources() []client.Object {
-	resources := make([]client.Object, len(openStackResources))
-	for i := range openStackResources {
-		resources[i] = openStackResources[i].DeepCopyObject().(client.Object)
+func (o *openstackAssets) GetTemplateValues() (common.TemplateValues, error) {
+	values := common.TemplateValues{
+		"images": o.images,
 	}
+	_, err := govalidator.ValidateMap(values, templateValuesValidationMap)
```
alexander-demichev, line 63:
```suggestion 
	if _, err := govalidator.ValidateMap(values, templateValuesValidationMap); err != nil {
```

pkg/cloud/openstack/openstack.go
```diff
@@ -2,36 +2,86 @@ package openstack
...
-	return resources
+func (o *openstackAssets) GetOperatorConfig() config.OperatorConfig {
+	return o.operatorConfig
+}
+
+func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets, error) {
+	images := &imagesReference{
+		CloudControllerManager: config.ImagesReference.CloudControllerManagerOpenStack,
+	}
+	_, err := govalidator.ValidateStruct(images)
```
alexander-demichev, line 74:
```suggestion 
	if _, err := govalidator.ValidateStruct(images); err != nil {
```

pkg/config/config.go
```diff
@@ -61,7 +61,7 @@ func getImagesFromJSONFile(filePath string) (ImagesReference, error) {
 
 // ComposeConfig creates a Config for operator
 func ComposeConfig(infrastructure *configv1.Infrastructure, clusterProxy *configv1.Proxy, imagesFile, managedNamespace string) (OperatorConfig, error) {
-	_, err := GetProviderFromInfrastructure(infrastructure)
+	err := checkInfrastructureResource(infrastructure)
```
alexander-demichev, line 64:
```suggestion 
	if err := checkInfrastructureResource(infrastructure); err != nil {
```

lobziik:
> I would start by looking at the render part of this and trying to simplify that, I don't think we need to use an interface into that part of the code, that to me just muddys the boundary a little. If you can re-look at that part, lets see where we get to after that

@JoelSpeed PTAL https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/110/commits/7a3f69f86b2b89e275bcce323508d3eea54eedaa

pkg/cloud/common/interfaces.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CloudProviderAssets represent cloud provider specific assets object. Such as Azure/Aws/OpenStack/etc.
+// Intended to hide provider specific resource management, manifests checks and rendering behind this interface.
+type CloudProviderAssets interface {
+	// GetRenderedResources intended to return ready to apply list of resources.
+	// Expected that resources returned by this method is completely prepared and do not need additional handling.
+	GetRenderedResources() ([]client.Object, error)
```
JoelSpeed, line 12: I still think that, until we have a reason to do otherwise, this should be. There's no need to keep the error in case we use it in the future, we can add it back if we need it ever
```suggestion
	GetRenderedResources() []client.Object
```

pkg/cloud/common/templates.go
```diff
@@ -0,0 +1,72 @@
...
+
+type TemplateSource struct {
+	ReferenceObject client.Object
+	EmbedFsPath     string
+}
+
+type ObjectTemplate struct {
+	TemplateSource
+	templateContent *template.Template
+}
```
JoelSpeed, line 24: Please add proper comments/documentation for this package

pkg/cloud/cloud.go
```diff
@@ -1,39 +1,72 @@
...
 // would be preserved.
-func GetResources(platformStatus *configv1.PlatformStatus) []client.Object {
+func GetResources(operatorConfig config.OperatorConfig) []client.Object {
+	assets, err := getAssets(operatorConfig)
+	if err != nil {
+		klog.Errorf("can not get assets: %s", err.Error())
+		return nil
+	}
+	renderedObjects := assets.GetRenderedResources()
+	if err != nil {
```
JoelSpeed, line 34: This error check isn't needed anymore

pkg/cloud/common/interfaces.go
lobziik: Fair point at the moment. I expect that it _may_ be little bit problematic to get it back in future, but  agree, at this point this err is not needed. Removed in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/110/commits/5c1cb86916a79392d22b6a64392b3d2c26c5d14d

JoelSpeed (commented):
Please squash the commits into logical commits and either merge alex's suggestions for inlining the error checks or resolve those as wont fix, then I think we are good to merge

pkg/cloud/aws/aws.go
lobziik: Bit opinionated but I don't like such construction here. Sorry, wont fix it :) The same for other comments around errors inlining for validation.

pkg/cloud/cloud.go
```diff
@@ -1,39 +1,68 @@
...
-// provisioning CCM instance in the cluster for the given platform type
+// GetResources selectively returns a list of resources required for
+// provisioning CCM instance in the cluster for the given OperatorConfig.
 //
 // These resources will be actively maintained by the operator, preventing
 // changes in their spec. However you can extend any resource spec with
 // values not specified in the provided source resource. These changes
 // would be preserved.
-func GetResources(platformStatus *configv1.PlatformStatus) []client.Object {
+func GetResources(operatorConfig config.OperatorConfig) []client.Object {
```
lobziik, line 27: Lets hold merge for a while. This need return error, IMO.

lobziik:
UPD: 
- Commits squashed with recent changes around removing _TemplateBasedAssets_ interface and docstrings for templates part of common module
- Proper error handling in GetResources in cloud module added (https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/110/commits/2d0a68b13f3e9de511dc4c9aa0a53383c9e46089)

Commit history before squash here: https://github.com/lobziik/cluster-cloud-controller-manager-operator/commits/provider-interface-commit-hist

pkg/cloud/cloud.go
JoelSpeed: Why do you think it needs an error? We are handling all of the things that could possibly return an error already in the constructor, this is just returning a pre-rendered value.

I don't want us to over-optimise here just in case this comes up in the future. Changing the signature back in a future PR should be simple enough

alexander-demichev (approved):
/approve

Fedosin (approved):
/lgtm

PR #111, opened 2021-08-31
Bug 1999018: Ensure rollingUpdate strategy is cleared to allow upgrades

JoelSpeed:
Because we moved the strategy to `Recreate` in #108, the rollingUpdate field must be empty. If it is not empty, this prevents the upgrade from going through. This should force it to empty to allow the upgrade.


Fedosin (approved):
/lgtm
/approve

PR #114, opened 2021-09-01
Bug 2000191: Ensure CCCMO and CCMs adheres to OpenShift recommended leader election

JoelSpeed:
To ensure consistency with other components within OpenShift, this PR enforces (via unit tests) a consistent set of leader election flags which adhere to the guidelines set out in the conentions (https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#handling-kube-apiserver-disruption)

### DIFF
--- a/cmd/cloud-config-sync-controller/main.go
+++ b/cmd/cloud-config-sync-controller/main.go
@@ -21,22 +21,26 @@ import (
 	"os"
 	"time"
 
+	"github.com/spf13/pflag"
+
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
-
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/klog/v2"
-	"k8s.io/klog/v2/klogr"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/component-base/config"
+	"k8s.io/component-base/config/options"
+	"k8s.io/klog/v2"
+	"k8s.io/klog/v2/klogr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/controllers"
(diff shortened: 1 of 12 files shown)


manifests/0000_26_cloud-controller-manager-operator_11_deployment.yaml
```diff
@@ -40,7 +40,11 @@ spec:
             URL_ONLY_KUBECONFIG=/etc/kubernetes/kubeconfig
           fi
           exec /cluster-controller-manager-operator \
-          --leader-elect \
+          --leader-elect=true \
+          --leader-elect-lease-duration=137s \
+          --leader-elect-renew-deadline=107s \
```
Fedosin, line 45: CCCMO doesn't have these like leader-elect-renew-deadline and leader-elect-retry-period, these values are taken directly from the code. https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/cmd/cloud-config-sync-controller/main.go#L117-L118

Fedosin:
We also need to consider cloud-config-sync-controller and update durations there as well

cmd/cloud-config-sync-controller/main.go
```diff
@@ -21,33 +21,40 @@ import (
...
-	// The default durations for the leader electrion operations.
-	leaseDuration = 120 * time.Second
-	renewDealine  = 110 * time.Second
-	retryPeriod   = 90 * time.Second
+	leaderElectionConfig = config.LeaderElectionConfiguration{
+		LeaderElect:   true,
+		LeaseDuration: util.LeaseDuration,
+		RenewDeadline: util.RenewDeadline,
+		RetryPeriod:   util.RetryPeriod,
+		ResourceName:  "cluster-cloud-controller-manager-leader",
```
lobziik, line 56: Should it be different with operator resource name?

lobziik:
I checked couple of test runs, all of them seems failing because of no permissions for lease resource.

 ```E0901 18:07:10.684821       1 leaderelection.go:330] error retrieving resource lock openshift-cloud-controller-manager/cluster-cloud-controller-manager-leader: leases.coordination.k8s.io "cluster-cloud-controller-manager-leader" is forbidden: User "system:serviceaccount:openshift-cloud-controller-manager-operator:cluster-cloud-controller-manager" cannot get resource "leases" in API group "coordination.k8s.io" in the namespace "openshift-cloud-controller-manager"```

cmd/cloud-config-sync-controller/main.go
JoelSpeed: Good catch!

JoelSpeed:
Nice catch on the RBAC, I had put the wrong namespace down for the operator and config sync controller, should be fixed now

lobziik:
/lgtm

I didn't find anything criminal connected to CCMO/CCMs  in CI runs.

PR #117, opened 2021-09-04
Deploy CCMs only if KCMO doesn't own cloud controllers

Fedosin:
To support migration of cloud controllers from KCM to CCM we need to make sure that KCM doesn't own them anymore before deploying CCMs.

KCM sets condition CloudControllerOwner to False when it's finished the migration on its side. CCCMO observes this condition and doesn't perform any action until it is False.

Requires https://github.com/openshift/cluster-kube-controller-manager-operator/pull/558 to work successfully


lobziik:
Just wondering, might it make sense to revive this old pr: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/27 and extract pre-provisioning checks into separate function? WDYT?

Fedosin:
Agree, the function is getting bigger, so adding provisioningAllowed would be nice.

JoelSpeed (commented):
Should we have some sort of check that covers this code to say, if we have already deployed, keep managing even if KCM has gone backwards somehow?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +256,48 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
+		return false, err
+	}
+
+	// If there is no condition, we assume that KCM owns the Cloud Controller
+	ownedByKCM := true
+
+	if kcm.Status.Conditions != nil {
+		for _, cond := range kcm.Status.Conditions {
+			if cond.Type == KCMCloudControllerOwnershipCondition {
+				ownedByKCM = cond.Status == operatorv1.ConditionTrue
```
elmiko, line 297: i couldn't quite tell from the unit tests, but is there a test case that exercises this condition?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +256,48 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
+		return false, nil
+	}
 
 	return true, nil
 }
+
+func (r *CloudOperatorReconciler) isCloudControllerOwnedByKCM(ctx context.Context) (bool, error) {
+	kcm := &operatorv1.KubeControllerManager{}
+	err := r.Get(ctx, client.ObjectKey{Name: kcmResourceName}, kcm)
+	if err != nil && !errors.IsNotFound(err) {
```
JoelSpeed, line 281: What happens when the object doesn't exist or it hasn't set the condition yet (because it's not running?)

pkg/controllers/clusteroperator_controller_test.go
```diff
@@ -404,6 +446,50 @@ var _ = Describe("Component sync controller", func() {
...
+								Status:             operatorv1.ConditionTrue,
+								LastTransitionTime: metav1.Now(),
+							},
+						},
+					},
+				},
+			},
+			expectProvisioned: false,
+		}),
+		Entry("Should not provision resources because KCMO hasn't set CloudControllerOwner condition", testCase{
```
JoelSpeed, line 474: How does this affect bootstrapping of the cluster? We need to verify that this won't break bootstrapping. IIUC KCMO will only start updating it's ClusterOperator after CCM has initialized the masters

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +256,55 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
+func (r *CloudOperatorReconciler) isCloudControllersOwnedByKCM(ctx context.Context) (bool, error) {
+	kcm := &operatorv1.KubeControllerManager{}
+	err := r.Get(ctx, client.ObjectKey{Name: kcmResourceName}, kcm)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// If KCM object doesn't exist, we assume that we are in bootstrapping process and
+			// the controllers are not owned by KCM.
+			klog.Info("Kube Config Manager object not found. It's considered that KCM does not own the controllers.")
+			return false, nil
+		}
```
JoelSpeed, line 287: I don't think this will work

The object you are fetching (https://github.com/openshift/cluster-kube-controller-manager-operator/blob/e52078fbafd5bc04210185ce546c89cbc83de9b2/manifests/0000_25_kube-controller-manager-operator_01_operator.cr.yaml#L1) is created by CVO before the CCCMO is created.

In this case, I think what we need to do is check whether the status is empty. CVO won't create a status, only KCMO will do that, does that make more sense as a check? Perhaps instead we should make sure status is empty and there's no observed config? 

Could this lead to a race condition during startup? if CCCMO starts before KCMO but isn't supposed to? Perhaps we need to check for KCMO's pod to be pending to be very certain?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +256,52 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
+		klog.Errorf("Unable to retrive Kube Config Manager object: %v", err)
+
+		if err := r.setStatusDegraded(ctx, err); err != nil {
+			klog.Errorf("Error syncing ClusterOperatorStatus: %v", err)
+			return false, fmt.Errorf("error syncing ClusterOperatorStatus: %v", err)
+		}
+		return false, err
+	}
+
+	if kcm.Status.Conditions == nil {
```
JoelSpeed, line 291: Is there a semantic difference here between nil and and empty slice that could trap us? IIUC this field will be there no matter as there's no pointers involved in the reference, so perhaps it's safer to check the length is zero?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -40,6 +41,10 @@ import (
 const (
 	externalFeatureGateName = "cluster"
 	proxyResourceName       = "cluster"
+	kcmResourceName         = "cluster"
+
+	// Condition type for Cloud Controller ownership
+	CloudControllerOwnershipCondition = "CloudControllerOwner"
```
JoelSpeed, line 47: Does this need to be exported?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +269,77 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
+	}
 
 	return true, nil
 }
+
+func (r *CloudOperatorReconciler) isCloudControllersOwnedByKCM(ctx context.Context) (bool, error) {
+	kcm := &operatorv1.KubeControllerManager{}
+	err := r.Get(ctx, client.ObjectKey{Name: kcmResourceName}, kcm)
+	if err != nil {
+		klog.Errorf("Unable to retrive Kube Config Manager object: %v", err)
```
JoelSpeed, line 295: Nit
```suggestion
		klog.Errorf("Unable to retrive KubeControllerManager object: %v", err)
```

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +269,77 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
+			klog.Errorf("Error syncing ClusterOperatorStatus: %v", err)
+			return false, fmt.Errorf("error syncing ClusterOperatorStatus: %v", err)
+		}
+		return false, err
+	}
+
+	if len(kcm.Status.Conditions) == 0 {
+		// If KCM Conditions object doesn't exist, we assume that we are in bootstrapping process and
+		// the controllers are not owned by KCM.
+		klog.Info("Kube Config Manager status not found. It's considered that KCM does not own the controllers.")
```
JoelSpeed, line 307: I think this might be a better message that explains why we are going ahead and deploying
```suggestion
		klog.Info("KubeControllerManager status not found, cluster is bootstrapping with external cloud providers")
```

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +269,77 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
+		}
+	}
+
+	return ownedByKCM, nil
+}
+
+func (r *CloudOperatorReconciler) isCloudControllersOwnedByCCM(ctx context.Context) (bool, error) {
+	co, err := r.getOrCreateClusterOperator(ctx)
+	if err != nil {
+		klog.Errorf("Unable to retrive Cluster Operator object: %v", err)
```
JoelSpeed, line 325:
```suggestion
		klog.Errorf("Unable to retrive ClusterOperator object: %v", err)
```

pkg/controllers/clusteroperator_controller.go
```diff
@@ -251,7 +269,77 @@ func (r *CloudOperatorReconciler) provisioningAllowed(ctx context.Context, infra
...
 	}
 
-	// TODO: Add part confirming migration is finished for KCM before allowing provisioning
+	ownedByKCM, err := r.isCloudControllersOwnedByKCM(ctx)
+	if err != nil {
+		return false, err
+	}
+	if ownedByKCM {
+		// KCM resource found and it owns Cloud provider
+		klog.Infof("Kube Controller Manager still owns Cloud Controllers. Skipping...")
```
JoelSpeed, line 278:
```suggestion
		klog.Infof("KubeControllerManager still owns Cloud Controllers. Skipping...")
```

PR #119, opened 2021-09-13
support alibaba cloud ccm

gujingit:


pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+          name: certs
+        - hostPath:
+            path: /etc/pki
+          name: pki
+        - name: cloud-config
+          configMap:
+            name: cloud-conf
+            items:
+              - key: cloud.conf
+                path: cloud-config.conf
```
JoelSpeed, line 108: Please ensure all files end with a new line (it helps with git diffs in the future)

pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,108 @@
...
+              #!/bin/sh
+              set -o allexport
+              if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+                source /etc/kubernetes/apiserver-url.env
+              fi
+              exec /cloud-controller-manager \
+              --kubeconfig=/etc/kubernetes/cloud-controller-manager.conf \
+              --address=127.0.0.1 \
+              --allow-untagged-cloud=true \
+              --leader-elect=true \
```
JoelSpeed, line 62: There are more leader election flags required to be configured, the unit tests should explain the issues (you can also copy/paste the leader election flags from any other platform)

pkg/cloud/cloud_test.go
```diff
@@ -1,6 +1,7 @@
 package cloud
 
 import (
+	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/alibaba"
```
JoelSpeed, line 4: Please move this import into the group with the rest of the cloud providers, it should come before the AWS import on line 9

pkg/cloud/cloud_test.go
```diff
@@ -67,6 +68,10 @@ func TestGetResources(t *testing.T) {
 		platform:     configv1.AzurePlatformType,
 		isAzureStack: true,
 		expected:     azurestack.GetResources(),
+	}, {
+		name:     "AlibabaCloud resources are empty",
```
JoelSpeed, line 72: This test description isn't correct
```suggestion
		name:     "AlibabaCloud resources returned as expected",
```

JoelSpeed (commented):
Please take a look at the cloud_test file again. There are several tests in there that the alibaba platform is needed to be added to, but I think you've only added the new platform to half of them

JoelSpeed:
> There are several tests in there that the alibaba platform is needed to be added to, but I think you've only added the new platform to half of them

Never mind this comment, I just merged #121 which removes the extra lists. I think you should be fine after a rebase and all the tests should be covered

/ok-to-test

JoelSpeed:
Changes look good, but you'll need to run `go mod vendor` and then add the resulting changes as an additional commit to pass the verify step. Once that's done I think we are good to go :)

pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,112 @@
...
+            - /bin/sh
+            - -c
+            - |
+              #!/bin/sh
+              set -o allexport
+              if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+                source /etc/kubernetes/apiserver-url.env
+              fi
+              exec /cloud-controller-manager \
+              --kubeconfig=/etc/kubernetes/cloud-controller-manager.conf \
```
elmiko, line 59: do we need this explicit kubeconfig flag given that we specify `--use-service-account-credentials=true` below?

gujingit: It's ok to remove the kubeconfig flag

elmiko:
i think we might be having some issues with our usage of `--use-service-account-credentials=true`, i've tried to capture what i think is happening in this issue https://github.com/openshift/cloud-provider-alibaba-cloud/issues/6

elmiko:
i think this will need to update the https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/manifests/0000_26_cloud-controller-manager-operator_01_images.configmap.yaml file as well.

gujingit:
Hi, what's the default image reference for alibaba ccm?  [https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/0083ec2523f754f7e9e1b6bf861c20db25cf297e/pkg/cloud/cloud_test.go#L44](url)

lobziik:
@gujingit Hi!

`quay.io/repository/openshift/origin-alibaba-cloud-controller-manager` i think

JoelSpeed:
/approve

/hold

I just want to check that we have the alibaba CCM included in the release payload before we go ahead with merging this. Otherwise I think it's good to go

kwoodson:
@JoelSpeed How do we verify the Alibaba CCM is in the release payload?

JoelSpeed:
@Fedosin has put in the requests for this to be done, perhaps he can give an update

kwoodson:
@JoelSpeed I have seen the latest image is being built in the 4.10 image. I saw the `ose-alibaba-cloud-controller-manager.yaml`.

I think this PR is ready.

elmiko (commented):
/lgtm

are we including the images yet @JoelSpeed ?

JoelSpeed:
The images are still marked as WIP, we need to bring in some changes from the upstream provider to fix the build before we can merge this

kwoodson:
@elmiko @JoelSpeed Looks like the fix to the `-mod readonly` has merged here https://github.com/kubernetes/cloud-provider-alibaba-cloud/commit/ef08fd22df7dc602d684827d06381ad6767ebebe. What is the process to force an update?

elmiko:
> @elmiko @JoelSpeed Looks like the fix to the `-mod readonly` has merged here [kubernetes/cloud-provider-alibaba-cloud@ef08fd2](https://github.com/kubernetes/cloud-provider-alibaba-cloud/commit/ef08fd22df7dc602d684827d06381ad6767ebebe). What is the process to force an update?

good question, i imagine we will need to update our code repo to match but the complicating factor here is the refactor of the code base. if we are good to move forward with testing the latest releases to the ccm then we could manually make a merge from the upstream to our copy. i _think_ @JoelSpeed was looking into this.

kwoodson:
@gujingit Now that the CCM has been updated, I found that there are a couple of options that are not supported. These two options are causing a failure to start the CCM:
```
unknown flag: --allow-untagged-cloud
unknown flag: --allocate-node-cidrs
```

Is there a plan to revisit this PR and test that it works? I also noticed that when launching the container that I received the following error message:
```
/bin/sh: /cloud-controller-manager: not found
```

I was able to fix this by using `CGO_ENABLED=0`. This enabled me to run the container.

gujingit:
> @gujingit Now that the CCM has been updated, I found that there are a couple of options that are not supported. These two options are causing a failure to start the CCM:
> 
> ```
> unknown flag: --allow-untagged-cloud
> unknown flag: --allocate-node-cidrs
> ```
> 
> Is there a plan to revisit this PR and test that it works? I also noticed that when launching the container that I received the following error message:
> 
> ```
> /bin/sh: /cloud-controller-manager: not found
> ```
> 
> I was able to fix this by using `CGO_ENABLED=0`. This enabled me to run the container.

@kwoodson @mtulio we have added these options, please pull the upstream

kwoodson:
Barring any other feedback I believe this is ready to go.

cc @JoelSpeed @elmiko

PR #120, opened 2021-09-13
Bug 2001825: Enforce the cloud-route controller disabled across platforms

JoelSpeed:
The cloud-route controller is disabled and forced disabled across the board today in OpenShift. This used to be enforced by setting `allocate-node-cidrs=false` but this was removed in https://github.com/kubernetes/cloud-provider/commit/f9aa959a5054f6160c1f5a7577220fbd104cd9bf (Note this is still true with in-tree cloud providers, just out of tree this no longer works).

Now we must set `configure-cloud-routes=false` to ensure that the route controller is not started. This should be a simpler approach to ensure the route controller is forced off than forcing the use of the `--controllers=•,-route` style flags.


Fedosin (approved):
/approve
/lgtm

PR #121, opened 2021-09-13
Remove duplicated platform list in cloud tests

JoelSpeed:
While reviewing the alibaba introduction to CCCMO I realised we have this same list duplicated across 3 different test suites. The likelihood of them getting out of sync is very high as we add more platforms.
To make this less likely, I've moved the list into a function so that we just call the function to return all of the platforms across each of the tests that need this list.


Fedosin (approved):
/approve
/lgtm

PR #124, opened 2021-09-14
Check diff after generate and vendor tasks

Fedosin:
Now our targets just fix related issues in the code. For CI it means that it doesn't report any errors, as jobs always finish successfully.

This commit adds a script that checks the output of git diff command. If the diff is not empty, then the code has been modified, and therefore it has some errors.

Running this script in CI will force it to report real errors in the jobs logs.

Co-authored-by: Danil Grigorev <dgrigore@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,10 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+verify-%:
+	make $*
+	./hack/verify-diff.sh
+
 all: build
 
 verify: fmt vet lint
--- a/hack/verify-diff.sh
+++ b/hack/verify-diff.sh
@@ -1,0 +1,9 @@
+FILE_DIFF=$(git ls-files -o --exclude-standard)
+
+if [ "$FILE_DIFF" != "" ]; then
+  echo "Found untracked files:"
+  echo $FILE_DIFF
+  exit 1
+fi
+
+git diff --exit-code


Makefile
```diff
@@ -45,23 +45,27 @@ manifests: controller-gen
...
 	( GOLANGCI_LINT_CACHE=$(PROJECT_DIR)/.cache $(GOLANGCI_LINT) run --timeout 10m )
+	./hack/verify-diff.sh
 
 # Run go mod
 .PHONY: vendor
 vendor:
 	go mod tidy
 	go mod vendor
 	go mod verify
+	./hack/verify-diff.sh
```
JoelSpeed, line 68: What about adding a new make target than can run other targets

```
verify-%:
  make $*
  ./hack/verify-diff.sh
```
Then we can have the ability to run `make vendor` without the verify (which would be useful)

JoelSpeed (commented):
I would like to see separate targets that check the verification so that the existing targets can still be used during development

Makefile
Fedosin: currently the behaviour is set by OPENSHIFT_CI env variable in the script. If it's true, then we perform verification, otherwise omit it. Isn't it enough for us?
So, on your local machine `make vendor` will update the code and return successfully, but on CI it will report an error if something is wrong.

JoelSpeed: Changing the behaviour based on an env var like that seems very much like magic and is very non-obvious (I will not remember this in 6 months), I would rather we be explicit about having separate tasks. 

As an example, if I wanted to verify there was no diff locally as well, I would have to try and work out what magic env var I have to set, which is not user friendly.

IMO we should be explicit here and have seperate make tasks for verification

Fedosin: ah, I got what you mean. I also don't like that variable

JoelSpeed (commented):
/approve

Thanks @Fedosin, can you make the appropriate openshift/release PR to get the targets updated for CI please and then link it here

PR #125, opened 2021-09-14
Minimize IAM scope for IBM Cloud CredReqs

cjschaef:
Minimize the IAM scope for the IBM Cloud CCM, based on required
resources and permissions.


cjschaef:
Need some confirmation from IBM network team on which roles are applicable and verify there are no additional services that need permissions.

elmiko (commented):
this makes sense, although i am no expert on the ibm policy permissions. would like to get an additional review on this.
/approve

rtheis (approved):
/lgtm

shalver:
/lgtm
but honestly this is not something I am very familiar with

PR #128, opened 2021-09-16
Bug 2004924: Enforce the cloud-route controller disabled across platforms

JoelSpeed:
Manual cherry-pick of #120 

We must disable the route controller across platforms as it is not supported within OpenShift.


Fedosin (commented):
/lgtm
/approve

hardys:
Discussed and agreed blocker status, adding staff-eng-approved

JoelSpeed:
/override ci/prow/e2e-openstack-ccm

Have discussed this with @pierreprinetti and @mdbooth on slack. This job is currently perma-failing due to a ci-config issue and as such will never pass. It was also accidentally marked as required. We are moving the job back to optional until we are confident in its stability

PR #130, opened 2021-09-22
Vsphere ccm support

lobziik:
Add bits for deploy ccm on vsphere platform

Successfully migrated OCP cluster on my lab vSphere once.

### DIFF
--- a/pkg/cloud/vsphere/assets/cloud-controller-manager-daemonset.yaml
+++ b/pkg/cloud/vsphere/assets/cloud-controller-manager-daemonset.yaml
@@ -1,0 +1,118 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: vsphere-cloud-controller-manager
+  namespace: openshift-cloud-controller-manager
+  labels:
+    app: vsphere-cloud-controller-manager
+spec:
+  selector:
+    matchLabels:
+      app: vsphere-cloud-controller-manager
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: vsphere-cloud-controller-manager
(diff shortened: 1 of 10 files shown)


Fedosin (approved):
/approve

pkg/cloud/vsphere/assets/cloud-controller-manager-daemonset.yaml
```diff
@@ -0,0 +1,118 @@
+kind: DaemonSet
```
JoelSpeed, line 1: Please use a Deployment to be consistent with other providers

JoelSpeed (commented):
I think it's better to be consistent with the other platforms and use a Deployment.
In some scenarios there will be 5 control plane nodes, we don't need 5 CCM pods sat there with 4 doing nothing. A deployment with 2 replicas should suffice

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -0,0 +1,119 @@
...
+              if [[ -f /etc/kubernetes/apiserver-url.env ]]; then
+                source /etc/kubernetes/apiserver-url.env
+              fi
+              exec /bin/vsphere-cloud-controller-manager \
+                --v=3 \
+                --cloud-config=$(CLOUD_CONFIG) \
+                --cloud-provider=vsphere \
+                --controllers=* \
+                --configure-cloud-routes=false \
+                --bind-address=127.0.0.1 \
```
JoelSpeed, line 86: Does this prevent the pod being scraped for metrics?

Fedosin (approved):
/lgtm

PR #131, opened 2021-09-23
Add GCP CCM support

Fedosin:
This commit allows to deploy GCP cloud controller manager on OpenShift.

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
@@ -1,0 +1,24 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  name: openshift-gcp-ccm
+  namespace: openshift-cloud-credential-operator
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  serviceAccountNames:
+  - cloud-controller-manager
+  secretRef:
+    name: gcp-ccm-cloud-credentials
+    namespace: openshift-cloud-controller-manager
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+      - roles/compute.instanceAdmin
+      - roles/compute.loadBalancerAdmin
+      - roles/iam.serviceAccountUser
+    # If set to true, don't check whether the requested
+    # roles have the necessary services enabled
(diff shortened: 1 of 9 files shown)


pkg/cloud/gcp/gcp.go
```diff
@@ -0,0 +1,80 @@
...
+	_, err := govalidator.ValidateStruct(images)
+	if err != nil {
+		return nil, fmt.Errorf("%s: missed images in config: %v", providerName, err)
+	}
+	assets := &GCPAssets{
+		operatorConfig: config,
+	}
+	objTemplates, err := common.ReadTemplates(assetsFs, templates)
+	if err != nil {
+		return nil, err
```
JoelSpeed, line 68:
```suggestion
		return nil, fmt.Errorf("could not read templates: %v", err)
```

pkg/cloud/gcp/gcp.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	templateValues, err := getTemplateValues(images, config)
+	if err != nil {
+		return nil, fmt.Errorf("can not construct template values for %s assets: %v", providerName, err)
+	}
+
+	assets.renderedResources, err = common.RenderTemplates(objTemplates, templateValues)
+	if err != nil {
+		return nil, err
```
JoelSpeed, line 77:
```suggestion
		return nil, fmt.Errorf("could not render templates: %v", err)
```

JoelSpeed:
Changes LGTM, please add appropriate PR description including details of how this has been tested

lobziik:
/hold
https://github.com/kubernetes/cloud-provider-gcp/pull/287 - this need to be merged and rebased. I thought that it is already done.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
```diff
@@ -0,0 +1,24 @@
...
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+      - roles/compute.instanceAdmin
+      - roles/compute.loadBalancerAdmin
+      - roles/iam.serviceAccountUser
+    # If set to true, don't check whether the requested
+    # roles have the necessary services enabled
+    skipServiceCheck: true
```
JoelSpeed, line 24: Nit, please add a new line to prevent future spurious git diffs

JoelSpeed (commented):
Other than this one nit, I don't think theres anything stopping us from merging this as is. While the provider won't actually work right now, we know that this configuration should work once the provider itself is fixed.

I think we should merge this and work to fix the issues with the provider itself.

Any objections @lobziik @Fedosin ?

PR #132, opened 2021-09-23
Fix "vendor" job

Fedosin:
Now the job is broken because "make verify-vendor" returns an error. This commit fixes the problem by updating go.mod file.

The code is automatically generated by `make vendor`.


alexander-demichev (approved):
/approve

PR #137, opened 2021-10-13
Replace fmt with goimports

lobziik:
For check imports ordering during ci runs.
Now `make fmt` and `make test` would leverage goimports instead.


Fedosin (approved):
/approve

PR #139, opened 2021-10-25
Get rid of server-side apply

Fedosin:
Currently the operator uses Server-Side Apply (SSA)[1] to manage its resources. Unfortunately, this approach has some drawbacks as it allows users to modify resources independently from the operator.

To prevent this behavior, this commit adds custom apply functions inspired by ones from resourseapply module in library-go[2].

[1] https://kubernetes.io/docs/reference/using-api/server-side-apply/
[2] https://github.com/openshift/library-go/tree/master/pkg/operator/resource/resourceapply

### DIFF
--- a/pkg/controllers/clusteroperator_controller.go
+++ b/pkg/controllers/clusteroperator_controller.go
@@ -36,6 +36,7 @@ import (
 
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/config"
+	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/controllers/resourceapply"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


pkg/controllers/clusteroperator_controller.go
```diff
@@ -151,39 +153,30 @@ func (r *CloudOperatorReconciler) sync(ctx context.Context, config config.Operat
...
 	for _, resource := range resources {
-		resourceExisting := resource.DeepCopyObject().(client.Object)
-		err := r.Get(ctx, client.ObjectKeyFromObject(resourceExisting), resourceExisting)
-		if errors.IsNotFound(err) {
-			klog.Infof("Resource %s %q needs to be created, operator progressing...", resource.GetObjectKind().GroupVersionKind(), client.ObjectKeyFromObject(resource))
-			updated = true
-		} else if err != nil {
-			r.Recorder.Event(resource, corev1.EventTypeWarning, "Update failed", err.Error())
-			return false, err
+		switch t := resource.(type) {
```
lobziik, line 162: I'm thinking about moving this switch to 'resourceapply' package as a separate function. WDYT?

Fedosin: +1

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -0,0 +1,227 @@
...
+	err = client.Update(ctx, toWrite)
+	if err != nil {
+		recorder.Event(required, corev1.EventTypeWarning, "Update failed", err.Error())
+		return false, err
+	}
+	recorder.Event(toWrite, corev1.EventTypeNormal, "Updated successfully", "Resource was successfully updated")
+	return true, err
+}
+
+func ApplyDeployment(ctx context.Context, client coreclientv1.Client, recorder record.EventRecorder, requiredOriginal *appsv1.Deployment) (bool, error) {
```
lobziik, line 125: Do we need to have this functions public?

lobziik (commented):
Looks good to me, just one question about exporting the whole bunch of different fuctions from `resourceapply` module. Might it make sense to export only `ApplyResource` out of it?

pkg/controllers/resourceapply/resourceapply.go
Fedosin: It's not necessary but I think it logically makes sense to keep these functions public even if we don't call them directly.

JoelSpeed: We can always make them public later without breaking our API surface, what we can't do is make them private, as otherwise people make have imported them and this would be a breaking change. They should be private until explicitly required to be public IMO

Fedosin: they are private now :)

PR #140, opened 2021-10-26
Bug 2017565: [OCPCLOUD-1306] Add user defined trusted CA from cloud-config to ccm-trusted-ca bundle 

lobziik:
Additional trusted CA from cloud-config should also be counted during
ccm-trusted-ca sync procedure. Due to OCP installer nuances,
'additionalTrustBundle' not always ends up in the proxy object.

For handling such situation this patch introduces support of 'ca-bundle.pem'
key in cloud-config. CA from there will be added to ccm-trusted-ca along with
additional CA bundle which reffered by proxy object.

For additional context see:
openshift/installer#5251 (comment)

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller_test.go
+++ b/pkg/controllers/cloud_config_sync_controller_test.go
@@ -123,7 +123,6 @@ var _ = Describe("Cloud config sync controller", func() {
 
 	var infraCloudConfig *corev1.ConfigMap
 	var managedCloudConfig *corev1.ConfigMap
-	var syncedCloudConfigMap *corev1.ConfigMap
 
 	var reconciler *CloudConfigReconciler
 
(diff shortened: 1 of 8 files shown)


lobziik:
Fixed tests in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/140/commits/0d96a1790803463bd18809616bca6cad0a21965e

/hold cancel

elmiko (commented):
thanks Denis
/approve

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -164,24 +174,12 @@ var _ = Describe("Cloud config sync controller", func() {
...
 			Eventually(
-				apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(managedCloudConfig), &corev1.ConfigMap{})),
-			).Should(BeTrue())
-		}
-
-		if syncedCloudConfigMap != nil {
-			Expect(cl.Delete(ctx, syncedCloudConfigMap, deleteOptions)).To(Succeed())
-			Eventually(
-				apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(syncedCloudConfigMap), &corev1.Namespace{})),
+				apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(&cm), &corev1.ConfigMap{})),
```
JoelSpeed, line 182: As mentioned in slack, but for the benefit or future selves. You could also use a `DeleteAllOf` call here, which will delete all of the configmaps in a particular namespace. For example this could be done:
```
By("Cleanup resources")
for _, ns := range []string{targetNamespaceName, OpenshiftConfigNamespace, OpenshiftManagedConfigNamespace} {
	Expect(cl.DeleteAllOf(ctx, &corev1.ConfigMap{},
		client.GracePeriodSeconds(0),
		client.InNamespace(ns),
	)).To(Succeed())
	Eventually(func() (bool, error) {
		cmList := &corev1.ConfigMapList{}
		if err := cl.List(ctx, cmList, client.InNamespace(ns)); err != nil {
			return false, err
		}
		return len(cmList.Items) == 0, nil
	},
	).Should(BeTrue())
}
```

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -198,7 +196,7 @@ var _ = Describe("Cloud config sync controller", func() {
 
 	It("config should be synced up after first reconcile", func() {
 		Eventually(func() (bool, error) {
-			syncedCloudConfigMap := &corev1.ConfigMap{}
+			syncedCloudConfigMap = &corev1.ConfigMap{}
```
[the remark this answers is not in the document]
lobziik: Not needed actually, will remove

lobziik:
I squashed commits and cleaned up tests for cloud-conf syncer a bit, IMHO it's ready for final review.
@JoelSpeed @Fedosin @elmiko PTAL

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -164,30 +173,17 @@ var _ = Describe("Cloud config sync controller", func() {
...
-				apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(managedCloudConfig), &corev1.ConfigMap{})),
-			).Should(BeTrue())
-		}
-
-		if syncedCloudConfigMap != nil {
-			Expect(cl.Delete(ctx, syncedCloudConfigMap, deleteOptions)).To(Succeed())
+		allCMs := &corev1.ConfigMapList{}
+		Expect(cl.List(ctx, allCMs)).To(Succeed())
+		for _, cm := range allCMs.Items {
+			Expect(cl.Delete(ctx, &cm, deleteOptions)).To(Succeed())
```
JoelSpeed, line 179: It's bad practice to take the location of a range variable, I'd recommend doing a `cm.DeepCopy()` here instead, same on line 181

JoelSpeed (commented):
One nit, I don't think it makes too much difference here but could cause a bug down the line if you were to create a list from within the loop or something

Fedosin (approved):
/lgtm

PR #141, opened 2021-11-01
Add exclusion annotation for hosted deployments

jonesbr17:
Configures the CVO to exclude manifests in deployments with external control planes
See https://github.com/openshift/cluster-version-operator/pull/252


JoelSpeed:
Just to be very explicitly clear in my head, this is intended to not deploy the operator on hypershift guest clusters because something else will be deploying the CCM within the hypershift host cluster?

jonesbr17:
After more [discussion](https://github.com/openshift/ibm-roks-toolkit/pull/284#discussion_r740198650), it sounds like we won't be going down this path

PR #142, opened 2021-11-09
Bump library-go for support Power VS out-of-tree cloud providers

Karthik-K-N:
This PR updates the library-go to support Power VS out-of-tree cloud providers


Fedosin (approved):
/lgtm
/approve

PR #146, opened 2021-11-11
Allow Azure node manager to be deployed on Windows

Fedosin:
This commit removes restrictions that doesn't allow the manager to work on Windows machines.


JoelSpeed:
We can't run the Azure node manager as a deamonset on windows nodes as the node manager binary is compiled for linux. We do not and will not ship a container image for windows for this. We must run the node manager as a windows service instead.

JoelSpeed (changes requested):
We cannot merge this as it will not work on OpenShift
/hold

PR #150, opened 2021-11-26
Bug 2025884: Inject openstack CCM image within deployment template

JoelSpeed:
We are supposed to substitute the image based on the image references we are passed. This is implemented using templating in all of the other providers, but it was missed within OSP.

CC @mdbooth


Fedosin (approved):
/approve
/lgtm

JoelSpeed:
Image was substituted correctly
/hold cancel

PR #151, opened 2021-12-03
Alibaba Cloud Provider: fix the path to the binary to match the build…

kwoodson:
… from the Dockerfile

Since there has been a small amount of churn there is a mismatch in the name and path of the binary from the Dockerfile.

Here is the corresponding Dockerfile:
https://github.com/openshift/cloud-provider-alibaba-cloud/blob/master/openshift-hack/images/Dockerfile.openshift#L12


elmiko (commented):
/approve
/lgtm

PR #153, opened 2021-12-09
Skip cloud-config sync on aws platform

lobziik:
AWS does not use cloud config at all at the moment,
skip its sync for this platform

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -35,6 +35,21 @@ type CloudConfigReconciler struct {
 func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	klog.Infof("%s emitted event, syncing cloud-conf ConfigMap", req)
 
+	infra := &configv1.Infrastructure{}
+	if err := r.Get(ctx, client.ObjectKey{Name: infrastructureResourceName}, infra); err != nil {
+		klog.Errorf("infrastructure resource not found")
+		return ctrl.Result{}, err
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -89,6 +99,18 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+func (r *CloudConfigReconciler) isCloudConfigSyncNeeded(platformStatus *configv1.PlatformStatus) (bool, error) {
+	if platformStatus == nil {
+		return false, fmt.Errorf("platformStatus is required")
+	}
+	switch platformStatus.Type {
+	case configv1.AWSPlatformType: // aws ccm does not use cloud-config at the moment
+		return false, nil
+	default:
+		return true, nil
+	}
```
JoelSpeed, line 111: Are we sure we want to make this an exclusive list rather than an inclusive list? How many platforms do vs don't use the cloud config?

lobziik: Only AWS at the moment

Fedosin (changes requested):
With your change you also need to update the failing test (TestAPI) and pre-create an infrastructure resource here: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/controllers/suite_test.go#L107
Like we do it here: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/controllers/clusteroperator_controller_test.go#L306

lobziik:
Hm, wondering why it does not fail on my local machine... Looking

Fedosin (approved):
/lgtm

PR #154, opened 2021-12-10
Bug 2032926: Bump envtest version

Fedosin:
Instead of a script, envtest started shipping as a compiled binary, so we have to updated our Makefile accordingly.
    
This bump is required to port CCCMO to k8s 1.23, because unit tests fail otherwise.


JoelSpeed:
Please add appropriate PR descriptions to all PRs, even if they are minimal. I can see what you are doing with this, but I don't know why.

Makefile
```diff
@@ -22,8 +30,8 @@ verify: fmt vet lint
 # Run tests
 test: generate verify manifests unit
 
-unit:
-	hack/unit-tests.sh
+unit: envtest
+	PATH=$(PATH):$(KUBEBUILDER_ASSETS) go test ./... -coverprofile cover.out
```
damdo, line 34: It looks like to override the default binaries path (`/usr/local/kubebuilder/bin`) we need to set the `KUBEBUILDER_ASSETS` env var rather than adding its value to the `PATH` env.
ref: https://github.com/kubernetes-sigs/controller-runtime/blob/11917e9d90625b51e5dab16b4cf20934971e2f08/pkg/envtest/server.go#L44
```suggestion
        PATH="$(PATH)" KUBEBUILDER_ASSETS="$(KUBEBUILDER_ASSETS)" go test ./... -coverprofile cover.out
```

Makefile
```diff
@@ -91,6 +99,14 @@ GOLANGCI_LINT = $(shell pwd)/bin/golangci-lint
 golangci-lint:
 	$(call go-get-tool,$(GOLANGCI_LINT),github.com/golangci/golangci-lint/cmd/golangci-lint)
 
+ENVTEST = $(shell pwd)/bin/setup-envtest
+.PHONY: envtest
+envtest: ## Download envtest-setup locally if necessary.
+	$(call go-get-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
```
damdo, line 105: Is it worth pinning `setup-envtest` here to the current `controller-runtime` version we use in `go.mod` (i.e. `v0.11.0`? Maybe we could have it in a variable at the top next to the `ENVTEST_K8S_VERSION` one.

damdo (commented):
I tried to fix the unit tests that are failing for this PR, locally, and with the two changes I'm suggesting above, they are now passing for me.

Makefile
Fedosin: Currently kubebuilder generates it with `latest` for all new projects. I don't think we should change this pattern.

damdo: I see, ok sounds good then 👍

elmiko (commented):
i gave this a test drive locally, everything worked as expected, thanks Mike!
/approve

PR #156, opened 2021-12-17
Bug 2033489: allow baremetal platform to skip syncing

elmiko:
on baremetal there is no credential to sync from, this change allows
cccmo to skip syncing on baremetal. it also adds a test clause for
sync skipping on baremetal.


elmiko:
i am approving this as it fixes a critical bug that is holding up ci, and also because i will likely be asleep when @zaneb's testing is concluded. i am also placing a hold on this that folks should feel free to remove when, and hopefully if, the tests are passing.
/approve
/hold until zaneb says so

zaneb:
Local test on metal succeeded
/hold cancel

PR #157, opened 2021-12-17
Bug 2028054: Ensure old deployment is deleted before new deployment creation

JoelSpeed:
We have seen instances during upgrades from 4.8 to 4.9 in which the deployments end up stuck on leader election due to the new deployment being created before the old deployment (possibly).

At the moment, the CVO doesn't wait on deployments to become ready, but if it did, our current ordering (tombstone after new deployment) could block the application of the manifests.

To avoid this possible conflict, we should make sure that the tombstone comes before the deployment manifest.

@wking Would you be able to sanity check this from the CVO side, I've moved the tombstone to a lower numbered manifest which I believe should guarantee it to be actioned before the new deployment


Fedosin (approved):
/approve

wking (approved):
/lgtm

PR #158, opened 2021-12-17
Bug 2033489: Use a list of platforms where config sync is required

Fedosin:
Currently we have a list of platforms where cloud sync is not required. Unfortunately, if we add a platform where it's not required and it's not in the list, the operator fails to start, and the whole cluster is degraded.

It's better to create a list of the platform where the sync is necessary, and ignore the others.


JoelSpeed (commented):
/approve
/lgtm

elmiko:
/title Bug 2033489: Use a list of platforms where config sync is required

elmiko:
/retitle Bug 2033489: Use a list of platforms where config sync is required

PR #155, opened 2021-12-15
Bug 2032926: feat: library-go bump and leader election config

eggfoobar:
Bumping the library-go dependency to latest, this should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242) and performance improvements in this [PR](https://github.com/openshift/library-go/pull/1235).

Changes:
  - updated to latest library-go
  - updated start commands to use leaderelection configs from library-go
  - updated controller-runtime to k8s 0.23.0 support
  - fixed test for new logr interface

Signed-off-by: ehila <ehila@redhat.com>

PR #159, opened 2021-12-17
Bug 2032926: Fix CA bundle tests to use async assertions correctly

JoelSpeed:
You should not use gomega assertions within async assertions as was previously done. This meant that the async assertions actually had nothing to assert against. A bug in gomega allowed the tests to pass though they shouldn't have done.

### DIFF
--- a/pkg/controllers/trusted_ca_bundle_controller_test.go
+++ b/pkg/controllers/trusted_ca_bundle_controller_test.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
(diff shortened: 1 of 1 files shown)


Fedosin (approved):
/approve
/lgtm

Fedosin:
/retitle Bug 2032926: Fix CA bundle tests to use async assertions correctly

Without this PR we can't upgrade to 1.23 because unit tests are broken there. Required by https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/155

PR #160, opened 2021-12-20
Bug 2032926: Fix config sync tests to use async assertions correctly

Fedosin:
You should not use gomega assertions within async assertions as was previously done. This meant that the async assertions actually had nothing to assert against. A bug in gomega allowed the tests to pass though they shouldn't have done.


elmiko (commented):
wow, nice find!
/approve

PR #161, opened 2021-12-20
Use "go install" to fetch binaries

Fedosin:
"go get" is deprecated and will be removed in the next release. We should use "go install" instead.


Makefile
```diff
@@ -92,28 +92,28 @@ push:
...
 .PHONY: envtest
 envtest: ## Download envtest-setup locally if necessary.
-	$(call go-get-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
+	$(call go-install-tool,$(ENVTEST),sigs.k8s.io/controller-runtime/tools/setup-envtest@latest)
 
-# go-get-tool will 'go get' any package $2 and install it to $1.
+# go-install-tool will 'go install' any package $2 and install it to $1.
 PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
-define go-get-tool
+define go-install-tool
```
lobziik, line 109: does `go install` needs all this squats with creating dummy module in tmp? Can we throw out this thing?

JoelSpeed: AFAIK go install won't need all the temp dir stuff, @Fedosin could you try and remove it and verify that things still work?

Fedosin: done!

Makefile
```diff
@@ -89,31 +87,19 @@ image:
...
 
-# Download controller-gen locally if necessary
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+
+.PHONY: controller-gen
 CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
-controller-gen:
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1)
+controller-gen: # Download controller-gen locally if necessary
+	GOBIN=$(PROJECT_DIR)/bin go install -mod=readonly sigs.k8s.io/controller-tools/cmd/controller-gen@latest
```
lobziik, line 95: Suggest to pin `controller-gen` and `setup-envtest` versions

lobziik: or better extract this versions into variables and move it to start of makefile

Fedosin: done!

PR #164, opened 2022-01-06
Bug 2037680: Fix CCCMO metric ports configuration

lobziik:
Disable redundant config-sync-controller metrics server,
change health check port from 9259 to 9260.
Reassign mistakenly exposed ports from config-sync-controller to operator.

Also need add 9260 port here: https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md as localhost only

P.S https://github.com/openshift/enhancements/pull/997

### DIFF
--- a/cmd/config-sync-controllers/main.go
+++ b/cmd/config-sync-controllers/main.go
@@ -65,12 +65,6 @@ func init() {
 func main() {
 	klog.InitFlags(nil)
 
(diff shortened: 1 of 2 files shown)


JoelSpeed (commented):
/approve

Thanks @lobziik

elmiko (commented):
/lgtm

PR #165, opened 2022-01-07
Bug 2037689: do not expose metrics port by default

gujingit:


JoelSpeed:
/ok-to-test
/approve
/lgtm

Thanks @gujingit

PR #167, opened 2022-01-31
Bug 2047998: Alibaba should deploy image from release payload

JoelSpeed:
The CCCMO must deploy images from the release payload and not directly from the quay repo


JoelSpeed:
/override ci/prow/e2e-aws-ccm

There is an issue with the config on the CCM, this test has actually passed, will get the test fixed up asap

PR #168, opened 2022-01-31
Bug 2048478: [release-4.10] Alibaba should deploy image from release payload

JoelSpeed:
Manual cherry-pick #167

The CCCMO must deploy images from the release payload and not directly from the quay repo


sdodson:
The upstream PR has merged so setting bugzilla/valid-bug. This is a blocker so adding staff-eng-approved.

JoelSpeed:
/override ci/prow/e2e-aws-ccm

The only test failure is an alert that is firing because the test is relying on a feature gate, we are working to resolve that

PR #171, opened 2022-02-15
Added credential request file for Power VS

Karthik-K-N:
Added credential request yaml file for Power VS platform


elmiko (commented):
i don't know the powervs credentials specifically, but this doesn't look out of band.
/approve

PR #172, opened 2022-02-15
Changes to support config map sych for Power VS Platform

Karthik-K-N:
With the [changes](https://github.com/Karthik-K-N/cluster-cloud-controller-manager-operator/commit/7f9529f78edcb0117f649f7ad23b702f8539c19b) added recently the config map synch was missed for Power VS platform. This PR adds Power VS platform to list of providers which need config map sync


Fedosin (approved):
/approve
/lgtm

PR #173, opened 2022-02-17
Bug 2055723: start watching KubeControllerManager resource

Fedosin:
CCM operator deploys required resources only if KCM doesn't own related controllers. To check it CCM operator looks for
CloudControllerOwner condition with false value on KCM CR in its cache.

Currently we don't watch KCM, and therefore this condition is not present in the cache.

To fix it, this commit adds KCM to the list of watched resources, so the manager will add updates to the cache.


elmiko (commented):
/lgtm

PR #174, opened 2022-02-23
Bug  2051457: CCM PodDisruptionBudgets

lobziik:
This PR introduces PodDisruptionBudgets for cloud-controller-manager pods, it includes
- Resource apply logic for PodDisruptionBudget type
- couple of new labels for all ccm/cnm pods with platform name value, namely:
    * infrastructure.openshift.io/cloud-controller-manager - for ccm
    *  infrastructure.openshift.io/cloud-node-manager - for cnm
- PodDisruptionBudget addition logic for deployments on all platfroms, except single-node OCP topology
- extended resourceapply for recreate Daemonsets and deployments in case if selectors was changed
- refactor resourceapply test to use envtest and gomega matcher


pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -4,17 +4,20 @@ metadata:
...
   namespace: openshift-cloud-controller-manager
   labels:
-    app: azure-cloud-controller-manager
+    k8s-app: azure-cloud-controller-manager
+    infrastructure.openshift.io/cloud-controller-manager: {{ .cloudproviderName }}
 spec:
   selector:
     matchLabels:
-      app: azure-cloud-controller-manager
+      k8s-app: azure-cloud-controller-manager
```
JoelSpeed, line 12: This change will break cluster upgrades, labels on deploymenys are immutable once created. Is there a way we can test this, will we need to delete and recreate to achieve this?

pkg/cloud/common/resources.go
```diff
@@ -0,0 +1,55 @@
...
+		if err != nil {
+			return nil, err
+		}
+		commonResources = append(commonResources, pdb)
+	}
+	return commonResources, nil
+}
+
+func getPDB(config config.OperatorConfig) (*policyv1.PodDisruptionBudget, error) {
+	minAvailable := intstr.FromInt(1)
```
JoelSpeed, line 33: Change this to `maxUnavailable` maybe? Then it would work even with a single replica?

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -224,3 +227,55 @@ func applyDaemonSet(ctx context.Context, client coreclientv1.Client, recorder re
 	recorder.Event(required, corev1.EventTypeNormal, "Updated successfully", "Resource was successfully updated")
 	return true, nil
 }
+
+func applyPodDisruptionBudget(ctx context.Context, client coreclientv1.Client, recorder record.EventRecorder, requiredOriginal *policyv1.PodDisruptionBudget) (bool, error) {
```
JoelSpeed, line 231: Has this been copied from library go?

JoelSpeed (commented):
/hold

We cannot update the label selectors on any deployment for a platform that is already GA'd, eg, AzureStackHub, IBM cloud. We need to make this change in a non-breaking way, or teach our apply logic to be able to detect the label selector config and delete and recreate the deployment

pkg/cloud/common/resources.go
```diff
@@ -0,0 +1,55 @@
...
+	matchLabels := map[string]string{
+		CloudControllerManagerProviderLabel: config.GetPlatformNameString(),
+	}
+	pdbNamePrefix := strings.ToLower(config.GetPlatformNameString())
+	pdbName := fmt.Sprintf("%s-cloud-controller-manager", pdbNamePrefix)
+	return &policyv1.PodDisruptionBudget{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "PodDisruptionBudget",
+			APIVersion: "policy/v1",
+		},
```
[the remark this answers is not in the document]
lobziik: Not absolutely, but It simplify testing drastically.

pkg/cloud/common/resources.go
lobziik: If we will strictly constraint number of ccm replicas to two, yes it might work with single replica.

However, i don't like this way, feel that it's harder to understand and troubleshoot then.

JoelSpeed: `maxUnavailable` tends to be more widely recommended as it is more flexible to changes than `minAvailable` and is in general safer.

In my experience, `minAvailable` is more likely to cause an issue than `maxUnavailable`. IMO this should be `maxUnavailable`, even if we have 3 replicas, yes, it slows down rollouts, but that's not necessarily a bad thing for such an important component

pkg/controllers/resourceapply/resourceapply.go
lobziik: Aligned it with library-go version, removed spec hash and generation annotation

pkg/cloud/azure/assets/cloud-controller-manager-deployment.yaml
lobziik: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/174/commits/0e963afcdeb0d3ead499d4abc7d67911801bc943 - recreate logic with envtest based tests

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -158,6 +164,34 @@ func applyDeployment(ctx context.Context, client coreclientv1.Client, recorder r
...
+			"Delete existing deployment", "Delete existing deployment to recreate it with new parameters",
+		)
+		// Perform dry run creation in order to validate deployment before deleting existing one
+		requiredCopy := required.DeepCopy()
+		requiredCopy.Name = fmt.Sprintf("%s-dry-run", requiredCopy.Name)
+		dryRunOpts := &coreclientv1.CreateOptions{DryRun: []string{metav1.DryRunAll}}
+		err = client.Create(ctx, requiredCopy, dryRunOpts)
+		if err != nil {
+			recorder.Event(existing, corev1.EventTypeWarning, "New resource validation failed", err.Error())
+			return false, err
```
JoelSpeed, line 185: Please wrap this error to provide additional context

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -158,6 +164,34 @@ func applyDeployment(ctx context.Context, client coreclientv1.Client, recorder r
...
+		}
+		err = client.Delete(ctx, existing)
+		if err != nil && !apierrors.IsNotFound(err) {
+			recorder.Event(existing, corev1.EventTypeWarning, "Deletion failed", err.Error())
+			return false, err
```
JoelSpeed, line 190: Please wrap this error to provide additional context

JoelSpeed (commented):
Please review my nits within the apply section and apply the nits to the rest of the code there as well please

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -158,6 +164,34 @@ func applyDeployment(ctx context.Context, client coreclientv1.Client, recorder r
...
+		}
+		return applyDeployment(ctx, client, recorder, required)
```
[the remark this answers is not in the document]
lobziik: >Are we recursing on applyDeployment now?

Yes

>This seems like it could be dangerous, eg, what happens if the deployment doesn't go away immediately and takes some time to go away?

I expect it will be handled the same way as any other error returned from this function. Operator might go to degraded for a while, and reconciled then. https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/ddc23ceec562563659d475a4a913a6d41b719d5b/pkg/controllers/clusteroperator_controller.go#L113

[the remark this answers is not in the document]
JoelSpeed: Is there some way to either avoid this recursion or limit its depth? What are we achieving by recursing here? Is it possible maybe to break the logic into two smaller functions and call the exact logic we need rather than recursing?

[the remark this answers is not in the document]
lobziik: >Is there some way to either avoid this recursion or limit its depth?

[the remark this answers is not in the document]
JoelSpeed: Ok, I think for now we can accept the recursion, but I want a comment on here that explains why the recursion is safe and why the recursion is required. I'd like to see similar comments on the Daemonset version

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -158,6 +164,38 @@ func applyDeployment(ctx context.Context, client coreclientv1.Client, recorder r
...
+	needRecreate := false
+	if !reflect.DeepEqual(existingCopy.Spec.Selector, required.Spec.Selector) {
+		needRecreate = true
+	}
+	if needRecreate {
+		klog.Infof("Deployment need to be recreated with new parameters")
+		// This check need for break a recursion in case if resource deletion was stuck during deletion
+		if existing.DeletionTimestamp != nil {
+			return false, fmt.Errorf("resource was already marked for deletion, returning")
+		}
```
JoelSpeed, line 178: What if the object is coming from a cached client? Is this actually going to work/ could we see stale data and try deleting again?

What API response do you get if you delete an already deleted object?

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -224,3 +294,45 @@ func applyDaemonSet(ctx context.Context, client coreclientv1.Client, recorder re
...
 }
+
+func applyPodDisruptionBudget(ctx context.Context, client coreclientv1.Client, recorder record.EventRecorder, requiredOriginal *policyv1.PodDisruptionBudget) (bool, error) {
+	required := requiredOriginal.DeepCopy()
+
+	existing := &policyv1.PodDisruptionBudget{}
+	err := client.Get(ctx, coreclientv1.ObjectKeyFromObject(required), existing)
+	if apierrors.IsNotFound(err) {
+		err = client.Create(ctx, required)
+		if err != nil {
```
JoelSpeed, line 305: Nit, don't shadow the outer error, inline this please
```suggestion
		if err := client.Create(ctx, required); err != nil {
```

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -224,3 +294,45 @@ func applyDaemonSet(ctx context.Context, client coreclientv1.Client, recorder re
...
+	if !*modified && contentSame {
+		return false, nil
+	}
+
+	// at this point we know that we're going to perform a write.  We're just trying to get the object correct
+	toWrite := existingCopy // shallow copy so the code reads easier
+	toWrite.Spec = *required.Spec.DeepCopy()
+
+	err = client.Update(ctx, toWrite)
+	if err != nil {
```
JoelSpeed, line 332: Nit, don't shadow previous error, declare within the if scope
```suggestion
	if err := client.Update(ctx, toWrite); err != nil {
```

pkg/cloud/alibaba/alibaba.go
```diff
@@ -42,9 +43,10 @@ func (a *alibabaAssets) GetRenderedResources() []client.Object {
 	return a.renderedResources
 }
 
-func getTemplateValues(images *imagesReference) (common.TemplateValues, error) {
+func getTemplateValues(images *imagesReference, operatorConfig config.OperatorConfig) (common.TemplateValues, error) {
 	values := common.TemplateValues{
-		"images": images,
+		"images":            images,
+		"cloudproviderName": operatorConfig.GetPlatformNameString(),
```
alexander-demichev, line 49: what if the platform name is an empty string?

lobziik: Validation will fail

pkg/controllers/resourceapply/resourceapply.go
lobziik: >What API response do you get if you delete an already deleted object?

JoelSpeed: > 404

Is that also when the deleted object has a finalizer, I'm trying to work out if deleting an already "deleted" object that hasn't gone away throws an error

JoelSpeed: I just checked, it doesn't throw an error, so we may loop here a few times still and have a successful "I deleted the resource" if the cache isn't updated yet <- this makes the recursion a possible issue and source of infinite recursion

lobziik: I removed recursion. If we can not rely to DeletionTimestamp, on a data which we obtaining from explicit Get invocation, if we have to suspect cache malfunction everywhere, ok, probably better to avoid somewhat hacky moves.

pkg/controllers/resourceapply/resourceapply.go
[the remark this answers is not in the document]
lobziik: removed recursion https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/174/commits/bec903d29c7b42c79afb801419a03b9fb5510654

elmiko (commented):
this generally looks good to me, +1

i will leave the lgtm label for @JoelSpeed to make sure all this concerns are addressed

lobziik:
>Have you tested the latest iteration of tis PR to verify that it still works as expected on upgrade, eg on Azure

Didn't test it manually. There is a whole bunch of envtest based test: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/174/files#diff-a9b533cdc5d42f96ee87c67af940c0e07dbffc2744eaec276d67b103c851960bR286 IMHO that should be sufficient.

JoelSpeed:
> Didn't test it manually. There is a whole bunch of envtest based test: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/174/files#diff-a9b533cdc5d42f96ee87c67af940c0e07dbffc2744eaec276d67b103c851960bR286 IMHO that should be sufficient.

I'd prefer to see some logs from an actual cluster with some manual testing before we merge this, would be good to create a cluster and then upgrade to this PR. I think cluster bot can do this for you, but it would be good to check the logs and see that the CCMs are up throughout

lobziik:
tested on azure manually, steps:
- install nightly via cluster bot
- applied FG for engaging CCCMO
- scaled down CVO after migration
- checked labels
- replaced CCCMO image with `quay.io/dmoiseev/cluster-cloud-controller-manager-operator:ccm-pdb` - built from this pr
- respective log:
```
dmoiseev@dmoiseev-mac ~ $ oc logs -n openshift-cloud-controller-manager-operator cluster-cloud-controller-manager-operator-5bdbf4f775-jsq8q cluster-cloud-controller-manager -f
I0321 16:15:06.740290       1 request.go:665] Waited for 1.036040111s due to client-side throttling, not priority and fairness, request: GET:https://api-int.ci-ln-mrqp8qk-1d09d.ci.azure.devcluster.openshift.com:6443/apis/operators.coreos.com/v1?timeout=32s
I0321 16:15:08.993975       1 logr.go:249] CCMOperator/controller-runtime/metrics "msg"="Metrics server is starting to listen"  "addr"=":9258"
I0321 16:15:08.994284       1 logr.go:249] CCMOperator/setup "msg"="starting manager"
I0321 16:15:08.994486       1 internal.go:362] CCMOperator "msg"="Starting server" "addr"={"IP":"::","Port":9258,"Zone":""} "kind"="metrics" "path"="/metrics"
I0321 16:15:08.994532       1 internal.go:362] CCMOperator "msg"="Starting server" "addr"={"IP":"127.0.0.1","Port":9259,"Zone":""} "kind"="health probe"
I0321 16:15:08.994552       1 leaderelection.go:248] attempting to acquire leader lease openshift-cloud-controller-manager-operator/cluster-cloud-controller-manager-leader...
I0321 16:18:06.840485       1 leaderelection.go:258] successfully acquired lease openshift-cloud-controller-manager-operator/cluster-cloud-controller-manager-leader
I0321 16:18:06.840763       1 controller.go:178] CCMOperator/controller/clusteroperator "msg"="Starting EventSource" "reconciler group"="config.openshift.io" "reconciler kind"="ClusterOperator" "source"="kind source: *v1.ClusterOperator"
I0321 16:18:06.840807       1 controller.go:178] CCMOperator/controller/clusteroperator "msg"="Starting EventSource" "reconciler group"="config.openshift.io" "reconciler kind"="ClusterOperator" "source"="kind source: *v1.Infrastructure"
I0321 16:18:06.840830       1 controller.go:178] CCMOperator/controller/clusteroperator "msg"="Starting EventSource" "reconciler group"="config.openshift.io" "reconciler kind"="ClusterOperator" "source"="kind source: *v1.FeatureGate"
I0321 16:18:06.840849       1 controller.go:178] CCMOperator/controller/clusteroperator "msg"="Starting EventSource" "reconciler group"="config.openshift.io" "reconciler kind"="ClusterOperator" "source"="kind source: *v1.KubeControllerManager"
I0321 16:18:06.840872       1 controller.go:178] CCMOperator/controller/clusteroperator "msg"="Starting EventSource" "reconciler group"="config.openshift.io" "reconciler kind"="ClusterOperator" "source"="channel source: 0xc0004810e0"
I0321 16:18:06.840908       1 controller.go:186] CCMOperator/controller/clusteroperator "msg"="Starting Controller" "reconciler group"="config.openshift.io" "reconciler kind"="ClusterOperator"
I0321 16:18:06.942863       1 controller.go:220] CCMOperator/controller/clusteroperator "msg"="Starting workers" "reconciler group"="config.openshift.io" "reconciler kind"="ClusterOperator" "worker count"=1
I0321 16:18:07.147634       1 resourceapply.go:173] Deployment need to be recreated with new parameters
I0321 16:18:07.282870       1 resourceapply.go:258] DaemonSet need to be recreated with new parameters
```
- checked labels again
   - selectors are correct
   - labels on pods are correct
   - no additional daemonsets/deployments/replicasets was not detected

@JoelSpeed

pkg/cloud/cloud_test.go
```diff
@@ -624,3 +624,46 @@ func checkDaemonSetSelectorLabels(t *testing.T, ds *appsv1.DaemonSet, platformTy
...
+		})
+	}
+}
+
+func derefReplicas(num *int32) int {
+	if num != nil {
+		return int(*num)
+	}
+	return 1
+}
```
JoelSpeed, line 669: Typically in a deref function you would have a second argument which is the default, might make it more usable later, wdyt?
```suggestion
func derefIntOrDefault(num *int32, default int32) int32 {
	if num != nil {
		return *num
	}
	return default
}
```

pkg/controllers/resourceapply/resourceapply.go
JoelSpeed: Please fix this comment

pkg/controllers/resourceapply/resourceapply.go
JoelSpeed: This hasn't been done

JoelSpeed:
Unit tests appear to be failing here, shall we try to get those resolved?

pkg/controllers/resourceapply/resourceapply_test.go
```diff
@@ -4,18 +4,84 @@ import (
...
+	return cl, teardownFunc
+}
+
+func cleanupResources(t *testing.T, g *WithT, ctx context.Context, cl client.Client, listObject client.ObjectList) {
+	g.Expect(cl.List(ctx, listObject)).To(Succeed())
+	deleteResouce := func(g *WithT, obj client.Object) {
+		key := client.ObjectKeyFromObject(obj)
+		obj.SetFinalizers([]string{})
+		g.Expect(cl.Update(ctx, obj)).To(Succeed())
+		err := cl.Delete(ctx, obj)
```
JoelSpeed, line 49: Would be better to do this asynchronously in case the first time you do it it fails, you could retry

JoelSpeed (commented):
I reviewed the test cases as it has flaked a few times, seems we are attempting to run the test before the default namespace exists on some runs, ie there's a race here.

To improve this, my recommendation would be to create a randomly named namespace for each test (use ObjectMeta.GenerateName) which we create when we set up the test case and then delete afterwards, this means each test has its own namespace, resources from previous tests won't interfere with the existing tests, in general, it should make things cleaner and, means we can avoid this race

pkg/controllers/resourceapply/resourceapply_test.go
lobziik: Honestly i can not imagine why it might fail there.
I would prefer to let it crash instead of retry.

elmiko (commented):
there are several usages of `fmt.Errorf` using the `%v` format for errors, i think we should convert these to `%w`, but i don't think it's a blocker here.
/lgtm

PR #175, opened 2022-03-02
Bug 2059716: Ensure release version is injected into all controller status clients

JoelSpeed:
We must ensure that the release version environment variable is observed in all controllers else the version reported by the clusteroperator will flap during execution


Fedosin (approved):
/lgtm

PR #178, opened 2022-03-14
Add generation of config for OpenStack CCM

stephenfin:
Implement the [OpenStack Cloud Provider Config Upgrade][enhancement]. This is tackled in the following steps:

- Remove code that references the legacy `openshift-config-managed/kube-cloud-config` configmap generated and managed by [Cluster Config Operator (CCO)][cco], since we'll do this ourselves in the future.
- Add machinery to support (validated) "transformation" of user-provided config to configuration that works with [Cluster Controller Manager (CCM)][ccm].
- Implement transformation logic to support [OpenStack CCM][occm].
- ~Implement transformation logic to support AWS and Azure CCMs, based on the CCO implementation~ (Dropped - this is a large task and needs more investigation)

[enhancement]: https://github.com/openshift/enhancements/pull/1009
[cco]: https://github.com/openshift/cluster-config-operator
[ccm]: https://github.com/kubernetes/controller-manager
[occm]: https://github.com/kubernetes/cloud-provider-openstack


pkg/cloud/cloud.go
```diff
@@ -18,6 +19,38 @@ import (
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/vsphere"
 )
 
+// cloudConfigTransformer function transforms the source config map using the input infrastructure.config.openshift.io object.
+// only the data and binaryData field of the output ConfigMap will be respected by consumer of the transformer.
+type cloudConfigTransformer func(source *corev1.ConfigMap, infra *configv1.Infrastructure) (*corev1.ConfigMap, error)
```
JoelSpeed, line 24: Nit, why do we need to return a pointer to the CM, surely the `source` input and configmap return values are the same pointer, or have I misunderstood?

pkg/cloud/cloud.go
```diff
@@ -18,6 +19,38 @@ import (
...
+		return common.NoOpTransformer, nil
+	case configv1.AzurePlatformType:
+		return common.NoOpTransformer, nil
+	case configv1.GCPPlatformType:
+		return common.NoOpTransformer, nil
+	case configv1.IBMCloudPlatformType:
+		return common.NoOpTransformer, nil
+// TODO: Enable this when we turn on syncing for OpenStack
+//	case configv1.OpenStackPlatformType:
+//	 	return openstack.CloudConfigTransformer, nil
```
JoelSpeed: Are we not doing that within this PR?

pkg/cloud/cloud.go
```diff
@@ -70,6 +101,8 @@ func getAssetsConstructor(platformStatus *configv1.PlatformStatus) (assetsConstr
 		return gcp.NewProviderAssets, nil
 	case configv1.IBMCloudPlatformType:
 		return ibm.NewProviderAssets, nil
+	case configv1.OpenStackPlatformType:
```
JoelSpeed, line 103: Is this to make it alphabetical? If so, 👍

pkg/cloud/common/config.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const (
+	DefaultConfigKey = "cloud.conf"
+)
+
+func NoOpTransformer(source *corev1.ConfigMap, infra *configv1.Infrastructure) (*corev1.ConfigMap, error) {
+	// Keys might be different between openshift-config/cloud-config and openshift-config-managed/kube-cloud-config
+	// Always use "cloud.conf" which is default one across openshift
+	output := source.DeepCopy()
```
JoelSpeed, line 18: What is the motivation to not modify the configmap in place? Are we expecting the configmap name/namespace or other to be changed, if so, why aren't we doing that here?

pkg/cloud/common/config_test.go
```diff
@@ -0,0 +1,61 @@
...
+func makeInfraCloudConfig() *corev1.ConfigMap {
+	return &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
+		Name:      infraCloudConfName,
+		Namespace: "openshift-config",
+	}, Data: map[string]string{infraCloudConfKey: "bar"}}
+}
+
+var _ = Describe("NoOpTransformer method", func() {
+	infra := makeInfrastructureResource(configv1.AzurePlatformType)
+	infraCloudConfig := makeInfraCloudConfig()
```
JoelSpeed, line 42: In case any of your tests modify these, it's better to ensure they are freshly initialised on a per test basis

```suggestion
	var infra *configv1.Infrastructure
	var infraCloudConfig *corev1.ConfigMap

	BeforeEach(func() {
		infra = makeInfrastructureResource(configv1.AzurePlatformType)
		infraCloudConfig = makeInfraCloudConfig()
	})
```

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -16,18 +16,22 @@ import (
...
-	defaultConfigKey = "cloud.conf"
-
 	// Controller conditions for the Cluster Operator resource
 	cloudConfigControllerAvailableCondition = "CloudConfigControllerAvailable"
 	cloudConfigControllerDegradedCondition  = "CloudConfigControllerDegraded"
 )
 
+// cloudConfigTransformer function transforms the source config map using the input infrastructure.config.openshift.io object.
+// only the data and binaryData field of the output ConfigMap will be respected by consumer of the transformer.
+type cloudConfigTransformer func(source *corev1.ConfigMap, infra *configv1.Infrastructure) (*corev1.ConfigMap, error)
```
JoelSpeed, line 33: Is there somewhere that this signature can be made public rather than duplicating it between two packages?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -74,8 +78,18 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	cloudConfigTransformerFn, err := cloud.GetCloudConfigTransformer(infra.Status.PlatformStatus)
 	if err != nil {
+		klog.Errorf("unable to get cloud config transformer function; unsupported platform")
+		if err := r.setDegradedCondition(ctx); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
+		}
+		return ctrl.Result{}, err
+	}
+
+	if sourceCM, err = cloudConfigTransformerFn(sourceCM, infra); err != nil {
```
JoelSpeed, line 91: The fact we are reusing the variable here makes me think we don't need to return the configmap at all from the function

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +84,73 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
 	return assets, nil
 }
+
+// CloudConfigTransformer implements the cloudConfigTransformer. It uses the source ConfigMap and
+// to create a new config that adheres to the configuration for CCM
+// It returns an error if the platform is not OpenStackPlatformType.
+func CloudConfigTransformer(source *corev1.ConfigMap, infra *configv1.Infrastructure) (*corev1.ConfigMap, error) {
+	if !(infra.Status.PlatformStatus != nil &&
+		infra.Status.PlatformStatus.Type == configv1.OpenStackPlatformType) {
+		return nil, fmt.Errorf("invalid platform, expected to be OpenStack")
```
JoelSpeed, line 94: Nit, use `errors.New` if you aren't using a format string

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +84,73 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+	if v, ok := output.Data[common.DefaultConfigKey]; ok {
+		inCfgRaw = []byte(v)
+	}
+
+	cfg, err := ini.Load(inCfgRaw)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read the cloud.conf: %w", err)
+	}
+
+	// TODO: Use 'HasSection' when we bump to gopkg.in/ini.v1 v1.67 or later
```
JoelSpeed, line 120: Why can't we bump to that now?

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +84,73 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+		return nil, fmt.Errorf("failed to modify the provided configuration: %w", err)
+	}
+
+	if global != nil {
+		if global.Key("secret-name").String() != "openstack-credentials" {
+			return nil, fmt.Errorf("'[Global] secret-name' is set to a non-default value")
+		}
+		if global.Key("secret-namename").String() != "kube-system" {
+			return nil, fmt.Errorf("'[Global] secret-namespace' is set to a non-default value")
+		}
```
JoelSpeed, line 132: What is the expected behaviour when this happens, will this cause the controller to enter a permanent error backoff? Do we need to log the transformation errors without returning them up the stack?

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +84,73 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+		global, err = cfg.NewSection("Global")
+		if err != nil {
+			return nil, fmt.Errorf("failed to modify the provided configuration: %w", err)
+		}
+	}
+
+	// Add the new required ones
+	global.NewKey("use-clouds", "true")
+	global.NewKey("clouds-file", "/etc/openstack/secret/clouds.yaml")
+	global.NewKey("cloud", "openstack")
```
JoelSpeed, line 149: Does NewKey overwrite existing keys or will it error if the key already exists? (Does it have any return values?)

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +84,73 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+
+	var buf bytes.Buffer
+	cfg.WriteTo(&buf)
```
JoelSpeed, line 152: Does this return an error?

pkg/cloud/openstack/openstack.go
mdbooth: @JoelSpeed I've seen this recommendation in review elsewhere, but I don't understand why. What's the advantage of errors.New?

pkg/cloud/openstack/openstack.go
mdbooth: We don't need it anyway. If there was a HasSection we'd write:

```
if HasSection {
  GetSection
}
```

Without it we write what we've got here:

```
GetSection
if section == nil {...}
```

It would be redundant.

pkg/cloud/openstack/openstack.go
mdbooth: The only resolution in this case would involve the user modifying the source configmap, so I think we:

1. Mark CCCMO non-upgradeable
2. Return a completed reconciliation

We'll expect to be reconciled again when the user modifies the source configmap.

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +84,73 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+	infraConfigKey := infra.Spec.CloudConfig.Key
+	if val, ok := output.Data[infraConfigKey]; ok {
+		output.Data[common.DefaultConfigKey] = val
+		delete(output.Data, infraConfigKey)
+	} else {
+		return nil, fmt.Errorf(
+			"key %s specified in infra resource was not found in source configmap %s",
+			infraConfigKey, client.ObjectKeyFromObject(source),
+		)
+	}
```
mdbooth, line 107: Every transformer is currently doing this. The target CM and key are common in all cases. The CloudConfigTransformer should take a *string* as input containing the input config from the relevant configmap[key], and write a transformed *string*. The parent code should then write that to the target configmap[DefaultConfigKey], which will be the same in all cases(?).

For input we should have 2 cases:
* We're using CCO
* We're doing the transformation in CCCMO

If we're using CCO we should never fall back to the user-managed config. If CCO hasn't done its thing yet we wait until it has. For the first cut we should leave AWS and Azure configs in CCO, but it's good to play with them at the WIP stage so we can increase our confidence that we *can* implement them here.

If we're doing the transformation in CCCMO the source should always be the configmap[key] referenced in infra.

I think this would make the return from GetCloudConfigTransformer something like:

```golang
type CloudConfigTransformer func(config string, infra *configv1.Infrastructure) (string, error)

func GetCloudConfigTransformer(platformStatus *configv1.PlatformStatus) (CloudConfigTransformer, error)
```

If GetCloudConfigTransformer returns nil for the transformer we use CCO as the source CM and the identity transformer. If it returns an actual transformer we use the user config as the source. In both cases the same code writes the returned value to the same destination.

pkg/cloud/cloud.go
stephenfin: I've done this now. This was incomplete.

pkg/cloud/cloud.go
stephenfin: Yup. Some non-foolish consistency, hopefully :wink:

pkg/cloud/common/config.go
stephenfin: I haven't changed anything functional here: this code was previously in `prepareSourceConfigMap` and I simply renamed the variable (`s/cloudConfCm/output/`). Happy to drop this if that's doable. However, note that we do change the name and namespace but that is done separately ([here](https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/fde5bd720458cf2593c75bad7dbc2c9b2d8de9ba/pkg/controllers/cloud_config_sync_controller.go#L177-L181))

pkg/controllers/cloud_config_sync_controller.go
stephenfin: Sure. Will fix.

pkg/cloud/openstack/openstack.go
stephenfin: v1.67 doesn't exist yet. I can remove this `TODO`. It's not important.

pkg/cloud/openstack/openstack.go
stephenfin: What @mdbooth said. Should I note this somewhere?

pkg/cloud/openstack/openstack.go
stephenfin: It appears to override existing keys (see [here](https://github.com/go-ini/ini/blob/v1.66.4/section.go#L65-L94)). There are return values, namely a [`Key`](https://github.com/go-ini/ini/blob/v1.66.4/key.go#L26-L39) and `error`, but we'll never hit those errors. I can add error handling here though, if needed?

pkg/cloud/openstack/openstack.go
stephenfin: Yes, it will return a `bytes.Buffer` and any errors generated by calls to `bytes.Buffer.WriteString` or `bytes.Buffer.Write` (which is [what this is exclusively using under the hood](https://github.com/go-ini/ini/blob/v1.66.4/file.go#L336-L508). I can add handling for this.

pkg/cloud/openstack/openstack.go
stephenfin: I now use `configv1.OpenStackPlatformType` here so this is irrelevant now. Would still be good to know the advantage of `errors.New` though.

pkg/cloud/openstack/openstack.go
stephenfin: Added error handling

pkg/cloud/openstack/openstack.go
stephenfin: Done

pkg/cloud/openstack/openstack.go
stephenfin: I've done just this, I think. Let me know if I missed anything.

pkg/cloud/aws/aws.go
```diff
@@ -76,3 +77,32 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+
+// CloudConfigTransformer implements the cloudConfigTransformer. It uses the source ConfigMap and
+// infra.status.platformStatus.AWS.ServiceEndpoints to create a new config that include the
+// ServiceOverrides sections. The transformer uses infra.status.platformStatus.region as the
+// signing service for all the ServiceOverrides.
+// It returns an error if the platform is not AWSPlatformType.
+func CloudConfigTransformer(source string, infra *configv1.Infrastructure) (string, error) {
+	if !(
+		infra.Status.PlatformStatus != nil &&
+		infra.Status.PlatformStatus.Type == configv1.AWSPlatformType) {
```
JoelSpeed, line 89: Nit, I'm not sure about this indentation, it looks weird to me, hard to find the end of the `(`

Is think this might be clearer?
```suggestion
	if !(infra.Status.PlatformStatus != nil &&
		infra.Status.PlatformStatus.Type == configv1.AWSPlatformType) {
```

pkg/cloud/aws/aws.go
```diff
@@ -76,3 +77,32 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+		return "", fmt.Errorf("invalid platform, expected to be %s", configv1.AWSPlatformType)
+	}
+	if infra.Status.PlatformStatus.AWS == nil || len(infra.Status.PlatformStatus.AWS.ServiceEndpoints) == 0 {
+		return source, nil  // no transformation required
+	}
+
+	var region string
+	if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil {
+		region = infra.Status.PlatformStatus.AWS.Region
+	}
```
JoelSpeed, line 99: Both of these conditions were already checked earlier in this function, you can  just inline this
```suggestion
	region := infra.Status.PlatformStatus.AWS.Region
```

pkg/cloud/aws/aws.go
```diff
@@ -76,3 +77,32 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+	if region == "" {
+		// TODO: CCM used field.Required for this; do we want to add that dep?
```
JoelSpeed, line 101: Which go module does the `field` package come from, if it's a standard K8s like apimachinery, then go ahead

pkg/cloud/common/config.go
```diff
@@ -0,0 +1,9 @@
+package common
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+func NoOpTransformer(source string, infra *configv1.Infrastructure) (string, error) {
```
JoelSpeed, line 7: Godoc on this please

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +83,59 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
 
 	return assets, nil
 }
+
+// CloudConfigTransformer implements the cloudConfigTransformer. It uses the source ConfigMap and
+// to create a new config that adheres to the configuration for CCM
```
JoelSpeed, line 88: It uses the source ConfigMap and what? Think you missed a word or two

pkg/cloud/openstack/openstack_test.go
```diff
@@ -44,3 +52,92 @@ func TestResourcesRenderingSmoke(t *testing.T) {
...
+	}
+
+	for _, tc := range tc {
+		t.Run(tc.name, func(t *testing.T) {
+			actual, err := CloudConfigTransformer(tc.source, tc.infra)
+			if tc.errMsg != "" {
+				assert.EqualError(t, err, tc.errMsg)
+				return
+			} else {
+				assert.NoError(t, err)
```
JoelSpeed, line 133: Nit: As a team, we are trying to move away from testify towards using gomega for assertions. The majority of this project already uses gomega, are you able to rewrite this?

Not sure if you've used gomega so example below in case you haven't

```
import . "github.com/onsi/gomega"

{
  g := NewWithT(t)
  g.Expect(err).ToNot(HaveOccurred())

  g.Expect(actual).To(Equal(expected))
}
```

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -60,39 +62,78 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	sourceCM, err = r.prepareSourceConfigMap(sourceCM, infra)
+	if err != nil {
 		if err := r.setDegradedCondition(ctx); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
 		}
 		return ctrl.Result{}, err
 	}
 
+	if cloudConfigTransformerFn != nil {
+		// TODO: Do we need to care about stuff in output.BinaryData?
```
JoelSpeed, line 126: Binary data isn't allowed to contain any key that overlaps with the data map, so, we aren't expecting users to put their cloud config in binary data are we? I think we can ignore it

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +83,59 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+			"secret-namespace": "kube-system",
+			"kubeconfig-path":  "",
+		} {
+			if global.Key(key).String() != value {
+				return "", fmt.Errorf("'[Global] %s' is set to a non-default value", key)
+			}
+			global.DeleteKey(key)
+		}
+	} else {
+		// Section doesn't exist, ergo no validation to concern ourselves with
```
mandre, line 115: For a moment I was wondering if this should be an error case as there's absolutely no chance the original cloud.conf would have been functional since it had no creds to communicate with OpenStack.
But I understand the intention and agree it's nicer to the user to recover from this (unlikely) situation where the cloud.conf doesn't have a Global section.

pkg/cloud/openstack/openstack.go
```diff
@@ -80,3 +83,59 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+		"use-clouds":  "true",
+		"clouds-file": "/etc/openstack/secret/clouds.yaml",
+		"cloud":       "openstack",
+	} {
+		_, err = global.NewKey(key, value)
+		if err != nil {
+			return "", fmt.Errorf("failed to modify the provided configuration: %w", err)
+		}
+	}
+
```
mandre, line 132: We should also drop everything that is in the BlockStorage section as it's now handled by the CSI driver.

On that note, it seems we're using a [static cloud.conf](https://github.com/openshift/openstack-cinder-csi-driver-operator/blob/master/assets/configmap.yaml) for Cinder CSI, and settings such as `[BlockStorage] node-volume-attach-limit` would be lost. We likely want to perform a similar transformation on the storage side.

mandre (commented):
I gave this PR a spin (release image at [quay.io/mandre/release:cccmo-cloud-conf-upgrade](http://quay.io/mandre/release:cccmo-cloud-conf-upgrade)) and confirms the transformation happens correctly in the happy path scenario with a default cloud.conf upgrade. I'm going to perform more tests today. Good job @stephenfin !

pkg/cloud/openstack/openstack_test.go
stephenfin: Ugh, I hate that style of testing :smile: It's basically forces us to use a (BDD) DSL in test cases rather than Go code like everywhere else. Not my battle to fight though so done

pkg/cloud/aws/aws.go
stephenfin: Sure, done

pkg/cloud/aws/aws.go
stephenfin: I don't think so. Previously we checked `infra.Status.PlatformStatus.Type`, whereas here we're testing `infra.Status.PlatformStatus.AWS`. Are those the same thing?

In any case, this isn't staying in this PR. I just have it here to prove how we'd approach it if/when we do this.

pkg/cloud/openstack/openstack.go
stephenfin: It seems the configuration for cloud-provider-openstack [still has a BlockStorage section](https://github.com/kubernetes/cloud-provider-openstack/blob/033052ee1ccb4986ac652facf30963a83a854d1f/pkg/openstack/openstack.go#L114-L119)? This is presumably [used by cinder CSI](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md#block-storage). Are you sure we want to drop this?

pkg/cloud/aws/aws.go
stephenfin: Wait, never mind, I see we check `infra.Status.PlatformStatus.AWS` in a separate call. My bad. My comment about this not actually being part of the final product still makes sense though.

pkg/cloud/aws/aws.go
stephenfin: Yup, `k8s.io/apimachinery/pkg/util/validation/field`

JoelSpeed: Seems fine to use it then, we already import apimachinery

stephenfin:
I think this is ready for full review/potential merge. The only thing I'm not certain about is my changes to `pkg/cloud/openstack/assets/deployment.yaml`. What I've done _makes_ sense to me but I could have missed something. By all means, feel free to let me know if so!

pkg/cloud/cloud.go
```diff
@@ -18,6 +18,40 @@ import (
...
+	case configv1.AlibabaCloudPlatformType:
+		return common.NoOpTransformer, nil
+	case configv1.AWSPlatformType:
+		// FIXME: We need to implement a transformer for this. Currently we're
+		// relying on CCO to do the heavy lifting for us.
+		return nil, nil
+	case configv1.AzurePlatformType:
+		// FIXME: We need to implement a transformer for this. Currently we're
+		// relying on CCO to do the heavy lifting for us.
+		return nil, nil
```
JoelSpeed, line 38: Should these not be returning the `NoOpTransformer` for now? What happens when this returns a nil, is the caller handling that?

JoelSpeed (commented):
/approve

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -60,39 +62,80 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	if err := r.Get(ctx, defaultSourceCMObjectKey, sourceCM); errors.IsNotFound(err) {
-		klog.Warningf("managed cloud-config is not found, falling back to infrastructure config")
 
-		openshiftUnmanagedCMKey := client.ObjectKey{Name: infra.Spec.CloudConfig.Name, Namespace: OpenshiftConfigNamespace}
-		if err := r.Get(ctx, openshiftUnmanagedCMKey, sourceCM); err != nil {
-			klog.Errorf("unable to get cloud-config for sync")
+	sourceCM := &corev1.ConfigMap{}
+	managedConfigFound := false
+
+	// NOTE: We know that there is some transformation logic in place in CCO
```
lobziik, line 77: @stephenfin can you please specify what CCO means there. Leave repo link in this comment perhaps

pkg/cloud/cloud.go
lobziik: @stephenfin can we return NoOpTransformer there and remain logic around syncing managed config untouched? Or its necessary to apply transformations on unmanaged config?

lobziik (approved):
Codewise looks good to me, @stephenfin thanks a lot!

I have a question about managed/unmanaged config but otherwise its fine.

pkg/cloud/cloud.go
stephenfin: I intentionally wanted to treat these as "special" because they're the only things I can see that *definitely* need to refer to the CCO-generated configmaps. If I didn't have this distinction, I'd have to either (a) pull in CCO's configmap for everything or (b) pull it in for everything *except* the clouds that I have a non-noop transformer for (currently just OpenStack) [*] Both of these will ultimately result in CCO being involved in this loop for way longer than necessary so this helps minimize larger tech debt at the cost of a small bit of (temporary) complexity.

[*] There's also (c), which is a variant of (b): add special transformers for the other clouds != Azure or AWS (or OpenStack). This adds the downside of introducing code duplication since we'd be effectively duplicating the `NoOpTransformer` for these other clouds. It's also kind of a variation of what we're currently doing.

pkg/controllers/cloud_config_sync_controller.go
stephenfin: Whoops, this is [Cluster Config Operator](github.com/openshift/cluster-config-operator/). I can push a follow-up to clarify this if you'd like

pkg/cloud/openstack/openstack.go
stephenfin: Done. Will work on dropping this in `cloud-provider-openstack` next.

pkg/cloud/cloud.go
stephenfin: I've added more comments here to make this a small bit clearer

pkg/controllers/cloud_config_sync_controller.go
stephenfin: Fixed this

pkg/cloud/openstack/openstack_test.go
```diff
@@ -31,16 +39,132 @@ func TestResourcesRenderingSmoke(t *testing.T) {
...
+			name:   "Empty config",
+			source: "",
+			infra:  makeInfrastructureResource(configv1.OpenStackPlatformType),
+		}, {
+			name: "Non-empty config",
+			source: `[Global]
+secret-name = openstack-credentials
+secret-namespace = kube-system
+
+[BlockStorage]
```
mdbooth, line 124: Please can we explicitly test that unknown sections and parameters are passed through unmodified? e.g. add:

```
[blockstorage]
foobarwhizz = 9
```

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -60,39 +62,80 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if cloudConfigTransformerFn != nil {
+		// We ignore stuff in sourceCM.BinaryData. This isn't allowed to
+		// contain any key that overlaps with those found in sourceCM.Data and
+		// we're not expecting users to put their data in the former.
+		output, err := cloudConfigTransformerFn(sourceCM.Data[defaultConfigKey], infra)
+		if err != nil {
+			if err := r.setDegradedCondition(ctx); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
+			}
+			return ctrl.Result{}, err
```
mdbooth, line 134: I think we have a critical problem here: this value of `err` contains information which is required by the user, but we're not giving it to them. I believe controller-runtime will log this value, but logs are not user-visible. We need to somehow relate this error message to the condition and/or generate an event containing it.

I think we can merge this PR without fixing this, but I don't think we can GA without fixing it.

mdbooth (commented):
Some nits, and

/lgtm

However, to re-iterate my comment, I think the issue of making config transformation errors user-visible is *critical*, and a release blocker. Merging this allows us to test, though, so I don't want to hold it up.

mandre (commented):
Re-adding Matt's LGTM after the trivial go fmt fix.
/lgtm

stephenfin:
That's a very unhappy CI... :disappointed: I've no idea how to address these failures. They seem totally unrelated (even the `ci/prow/e2e-openstack-ccm` job)

mandre:
Seems like we've hit quota limits on AWS:
```
level=error msg=Error: [ERR]: Error building changeset: InvalidChangeBatch: [Exceeded limit MAX_RRSETS_BY_ZONE for zone: Z2GYOLTZHS5VK]
```

CI is retesting on AWS only, as `e2e-aws-ccm`, `e2e-aws-upgrade`, and `e2e-aws` are the only required jobs, so until quota issue is resolved your PR is effectively blocked.

JoelSpeed:
/override ci/prow/e2e-aws-ccm

I'm not concerned by the historic failures on the AWS side, this should have no effect to the AWS CCM and it appears from logs to be working as expected

pierreprinetti:
Implements OSASINFRA-2732

PR #182, opened 2022-04-14
Bug 2074471: Update enabled and use-octavia options in cloud-conf config-map

MaysaMacedo:
Since Neutron-LBaaS has been deprecated in OpenStack since queens
release, we must ensure installations use Octavia by default. Also,
when running one installation with Kuryr SDN the cloud provider
shouldn't handle the Services as Kuryr already does that. This commit
fixes the issues by ensuring that use-octavia is set as True and
enabled is set as False.

### DIFF
--- a/manifests/0000_26_cloud-controller-manager-operator_02_rbac_operator.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_02_rbac_operator.yaml
@@ -64,6 +64,15 @@ rules:
   - list
   - watch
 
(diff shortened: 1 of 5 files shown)


MaysaMacedo:
This is highly WIP and for some reason the clusteRole is not reconciled

MaysaMacedo:
closing in favor of https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/183

PR #185, opened 2022-04-22
OpenStack: ensure config-map is updated only when needed

MaysaMacedo:
The managed config-map is often updated even when all the
keys have the correct value, this behaviour is caused
by the fact that the order of the iteration of config-map
keys not preserved when using golang Map, consequently the final
version of the config-map would be considered different
from the previous one, when shouldn't. This commit fixes
the issue by ensuring that the order of the keys is
preserved in the config-map.

### DIFF
--- a/pkg/cloud/openstack/openstack.go
+++ b/pkg/cloud/openstack/openstack.go
@@ -125,12 +125,13 @@ func CloudConfigTransformer(source string, infra *configv1.Infrastructure) (stri
 		}
 	}
 
-	for key, value := range map[string]string{
-		"use-clouds":  "true",
-		"clouds-file": "/etc/openstack/secret/clouds.yaml",
-		"cloud":       "openstack",
+	// Use a slice to preserve keys order
+	for _, o := range []struct{ k, v string }{
+		{"use-clouds", "true"},
+		{"clouds-file", "/etc/openstack/secret/clouds.yaml"},
+		{"cloud", "openstack"},
 	} {
-		_, err = global.NewKey(key, value)
+		_, err = global.NewKey(o.k, o.v)
 		if err != nil {
 			return "", fmt.Errorf("failed to modify the provided configuration: %w", err)
 		}
--- a/pkg/cloud/openstack/openstack_test.go
+++ b/pkg/cloud/openstack/openstack_test.go
@@ -1,11 +1,11 @@
 package openstack
 
 import (
+	"strings"
 	"testing"
 
 	. "github.com/onsi/gomega"
 	configv1 "github.com/openshift/api/config/v1"
-	ini "gopkg.in/ini.v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/config"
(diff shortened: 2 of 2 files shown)


pkg/cloud/openstack/openstack.go
```diff
@@ -125,12 +125,14 @@ func CloudConfigTransformer(source string, infra *configv1.Infrastructure) (stri
 		}
 	}
 
-	for key, value := range map[string]string{
+	configKeys := []string{"cloud", "clouds-file", "use-clouds"}
+	configMap := map[string]string{
```
mdbooth, line 129: @pierreprinetti is going to hate this too, but how about:

```golang
for _, o := range []struct{ k, v string } {
  {"use-clouds", "true"},
  ...
} {
  _, err = global.NewKey(o.k, o.v)
  ...
}
```

mdbooth (changes requested):
We need to fix the unit test too: https://github.com/shiftstack/cluster-cloud-controller-manager-operator/blob/d2f73eae92113ff6ee46e89f57f8867b8773b107/pkg/cloud/openstack/openstack_test.go#L141-L147

We were actually working round the fact that the order was changing in the unit test instead of noticing that this was a huge problem 🤦

pkg/cloud/openstack/openstack.go
mandre: Yeah, we'll need to use a slice to preserve the keys order.
We should perhaps add a comment too to make it more obvious:

```go
// Use a slice to preserve keys order
for _, o := range []struct{ k, v string }{
	{"use-clouds", "true"},
	{"clouds-file", "/etc/openstack/secret/clouds.yaml"},
	{"cloud", "openstack"},
} {
	_, err = global.NewKey(o.k, o.v)
	if err != nil {
		return "", fmt.Errorf("failed to modify the provided configuration: %w", err)
	}
}

```

mandre (commented):
> We need to fix the unit test too: https://github.com/shiftstack/cluster-cloud-controller-manager-operator/blob/d2f73eae92113ff6ee46e89f57f8867b8773b107/pkg/cloud/openstack/openstack_test.go#L141-L147
> 
> We were actually working round the fact that the order was changing in the unit test instead of noticing that this was a huge problem facepalm

My bad, I'm largely responsible for this.

If we simplify the tests, the patch should be:
```diff
diff --git i/pkg/cloud/openstack/openstack_test.go w/pkg/cloud/openstack/openstack_test.go
index ec5a911..3e8c242 100644
--- i/pkg/cloud/openstack/openstack_test.go
+++ w/pkg/cloud/openstack/openstack_test.go
@@ -1,11 +1,11 @@
 package openstack
 
 import (
+       "strings"
        "testing"
 
        . "github.com/onsi/gomega"
        configv1 "github.com/openshift/api/config/v1"
-       ini "gopkg.in/ini.v1"
        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
        "github.com/openshift/cluster-cloud-controller-manager-operator/pkg/config"
@@ -132,39 +132,18 @@ ignore-volume-az = true
        for _, tc := range tc {
                t.Run(tc.name, func(t *testing.T) {
                        g := NewWithT(t)
-                       output, err := CloudConfigTransformer(tc.source, tc.infra)
+                       actual, err := CloudConfigTransformer(tc.source, tc.infra)
                        if tc.errMsg != "" {
                                g.Expect(err).Should(MatchError(tc.errMsg))
                                return
                        } else {
                                g.Expect(err).ToNot(HaveOccurred())
-                               // The output is unsorted so we must reload and reparse the
-                               // strings
-                               expected, err := ini.Load([]byte(`[Global]
+                               expected := `[Global]
 use-clouds  = true
 clouds-file = /etc/openstack/secret/clouds.yaml
-cloud       = openstack`))
-                               g.Expect(err).ToNot(HaveOccurred())
-                               actual, err := ini.Load([]byte(output))
-                               g.Expect(err).ToNot(HaveOccurred())
-
-                               // Because things aren't sorted, we need to manually iterate
-                               // over sections and keys
-                               g.Expect(expected.SectionStrings()).Should(ConsistOf(actual.SectionStrings()))
-                               for _, sectionName := range expected.SectionStrings() {
-                                       expectedSection, err := expected.GetSection(sectionName)
-                                       g.Expect(err).ToNot(HaveOccurred())
-                                       actualSection, err := actual.GetSection(sectionName)
-                                       g.Expect(err).ToNot(HaveOccurred())
-                                       g.Expect(expectedSection.KeyStrings()).Should(ConsistOf(actualSection.KeyStrings()))
-                                       for _, keyName := range expectedSection.KeyStrings() {
-                                               expectedKey, err := expectedSection.GetKey(keyName)
-                                               g.Expect(err).ToNot(HaveOccurred())
-                                               actualKey, err := actualSection.GetKey(keyName)
-                                               g.Expect(err).ToNot(HaveOccurred())
-                                               g.Expect(expectedKey.String()).Should(Equal(actualKey.String()))
-                                       }
-                               }
+cloud       = openstack`
+                               actual := strings.TrimSpace(actual)
+                               g.Expect(actual).Should(Equal(expected))
                        }
                })
        }
```

PR #176, opened 2022-03-03
Bug 2059716: Ensure release version is set on config sync controller

JoelSpeed:
Follow up to #175, I missed this controller and only added the release version to one of the two controllers in this manager

PR #187, opened 2022-04-28
Bug 2079791: Ensure release version is injected into all controller status clients

JoelSpeed:
Manual cherry-pick of #175 and #176

> We must ensure that the release version environment variable is observed in all controllers else the version reported by the clusteroperator will flap during execution


elmiko (commented):
/lgtm

PR #193, opened 2022-05-19
[OCPCLOUD-1541] Bug 2089334: Use service account credentials for all providers

Fedosin:
OpenShift enforces the `--use-service-account-credentials` flag on CCMs.

This commit adds a test that makes sure this is configured correctly on each cloud provider except `vSphere` platform.

On vSphere they changed the behavior from what they have in the in-tree cloud provider. Code in this CCM executed by the `service-controller` runs as `vsphere-cloud-controller-manager` in the `kube-system` namespace, which is [a dynamically created but hardcoded service account](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/cloud.go#L108). But in the in-tree cloud provider the client is called [vsphere-legacy-cloud-provider](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/vsphere/vsphere.go#L278)

### DIFF
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -338,6 +338,11 @@ func TestRenderedResources(t *testing.T) {
 				checkLeaderElection(t, podSpec)
 				checkCloudControllerManagerFlags(t, podSpec)
 				checkTrustedCAMounted(t, podSpec)
+				if platform.platformStatus.Type != configv1.VSpherePlatformType {
+					// we don't use service account credentials on vSphere as its CCM behavior
+					// is different from the in-tree cloud provider.
+					checkUseServiceAccountCredentials(t, podSpec)
+				}
 			}
 		})
 	}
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
/lgtm
/approve

JoelSpeed:
/hold
/lgtm cancel
/approve cancel

The CI is showing that the vsphere CCM is not working correctly, the clusters have not completed bootstrapping as the nodes are left uninitialized, @Fedosin can you investigate this by spinning up your own vSphere cluster and inspecting why the pod isn't working correctly?

pkg/cloud/cloud_test.go
```diff
@@ -338,6 +338,11 @@ func TestRenderedResources(t *testing.T) {
 				checkLeaderElection(t, podSpec)
 				checkCloudControllerManagerFlags(t, podSpec)
 				checkTrustedCAMounted(t, podSpec)
+				if platform.platformStatus.Type != configv1.VSpherePlatformType {
+					// we don't use service account credentials on vSphere as its CCM behavior
+					// is different from the in-tree cloud provider.
```
JoelSpeed, line 343: Can you explain in the PR description why this is the case, what does it do differently, would be good to know for the future why we have this exception

Fedosin: Done

PR #200, opened 2022-07-27
Reset Go mod after openstack library-go bump

JoelSpeed:
Following on from #199, we force merged dependencies from a branch. This resets back to the updated library-go commit and removes the replace statement.

No diff to the vendor.


elmiko (commented):
/approve

PR #205, opened 2022-11-23
OCPCLOUD-1724: GCP Credentials req. manifest of CCMO to use new API field

odvarkadaniel:
This PR removes all the `predefinedRoles` which give elevated permissions. We get those permissions and put them into the new API field `Permissions []string`. These permissions are known to thanks to the command `gcloud iam roles describe <role>` - this lists all the included permissions we want.


odvarkadaniel:
Got rid of permission that were unnecessary in my opinion. Let's see how e2e goes now.

odvarkadaniel:
I think this can get tested by QE as well. The same introduction as [here](https://github.com/openshift/machine-api-operator/pull/1087), but with different set of permission to satisfy services that are used by CCCMO.

lobziik (commented):
Did you find any documentation about needed permissions for GCP ccm?

https://github.com/kubernetes/kops checking kops might be a good idea, not sure, but there might be some code that sets up permissions during cluster installation.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
```diff
@@ -15,10 +15,107 @@ spec:
...
+    - "compute.regionBackendServices.delete"
+    - "compute.regionBackendServices.get"
+    - "compute.regionBackendServices.getIamPolicy"
+    - "compute.regionBackendServices.list"
+    - "compute.regionBackendServices.setIamPolicy"
+    - "compute.regionBackendServices.setSecurityPolicy"
+    - "compute.regionBackendServices.update"
+    - "compute.regionBackendServices.use"
+    - "compute.instances.attachDisk"
+    - "compute.instances.create"
```
lobziik, line 67: I'm not sure that CCM needs `instance.create` permission...

odvarkadaniel:
> Did you find any documentation about needed permissions for GCP ccm?
> 
> https://github.com/kubernetes/kops checking kops might be a good idea, not sure, but there might be some code that sets up permissions during cluster installation.

Will take a look at that as well. I got through [this](https://github.com/openshift/cloud-provider-gcp/tree/master/providers/gce) to check what permissions we might need here.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
odvarkadaniel: I think it might need it according to this code: https://github.com/openshift/cloud-provider-gcp/blob/master/providers/gce/gce_instances.go#L492-L498

abutcher:
We're leaking these custom roles created by CCO in CI so holding until we're cleaning them up.
/hold

odvarkadaniel:
As you mentioned before, @lobziik, we needed to take a look at `KOPS` to investigate whether there are some additional permission that need to be added. I found out that we were missing a few, which are now in the list. I will have to see if this works once the permissions stuff is resolved.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
lobziik: This thing uses within tests. I don't think that we need to request this permission for production installations.

I suggest to double check any create/delete permissions. Some of the functions which you pointed might not be used at all or be used in a tests code.

lobziik (commented):
I still do think that there are some unnecessary permissions, such as:
- `compute.instances.create`
- `compute.instances.delete`
- "compute.instances.stop"
- "compute.instances.suspend"
- "compute.instances.update" (?) might be needed

I think need to double-check how respective client functions are used within the CCM and remove excessive permissions.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
```diff
@@ -15,10 +15,45 @@ spec:
...
+    - "compute.httpHealthChecks.delete"
+    - "compute.httpHealthChecks.get"
+    - "compute.httpHealthChecks.update"
+    - "compute.instanceGroups.addInstances"
+    - "compute.instanceGroups.create"
+    - "compute.instanceGroups.delete"
+    - "compute.instanceGroups.get"
+    - "compute.instanceGroups.listInstances"
+    - "compute.instanceGroups.removeInstances"
+    - "compute.instances.delete"
```
lobziik, line 44: Why is this needed?

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
```diff
@@ -15,10 +15,45 @@ spec:
...
+    - "compute.httpHealthChecks.create"
+    - "compute.httpHealthChecks.delete"
+    - "compute.httpHealthChecks.get"
+    - "compute.httpHealthChecks.update"
+    - "compute.instanceGroups.addInstances"
+    - "compute.instanceGroups.create"
+    - "compute.instanceGroups.delete"
+    - "compute.instanceGroups.get"
+    - "compute.instanceGroups.listInstances"
+    - "compute.instanceGroups.removeInstances"
```
lobziik, line 43: Don't we need `use`, `list` and `update` permissions?

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
```diff
@@ -15,10 +15,45 @@ spec:
...
+    - "compute.instances.delete"
+    - "compute.instances.get"
+    - "compute.regionBackendServices.create"
+    - "compute.regionBackendServices.delete"
+    - "compute.regionBackendServices.get"
+    - "compute.regionBackendServices.update"
+    - "compute.targetPools.addInstance"
+    - "compute.targetPools.create"
+    - "compute.targetPools.delete"
+    - "compute.targetPools.get"
```
lobziik, line 53: I would expect `update` and `use` here as well.

lobziik (commented):
Looks much better to me!

I'm a bit confused with missing `list`, `update`, and `use` permissions for networking-related things, left questions inline. `instances.delete` is also looks bit strange.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
odvarkadaniel: This is indeed not needed.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
odvarkadaniel: I don't think we need that or at least didn't find any reference of that service call anywhere.

odvarkadaniel: `compute.instanceGroup.List` is used just for tests in this case. `Update` and `Use` are not present.

manifests/0000_26_cloud-controller-manager-operator_16_credentialsrequest-gcp.yaml
odvarkadaniel: These are also not included in the list of service in the code.

PR #206, opened 2022-12-01
vSphere: External CCM translation, Zonal 

jcpowermac:
This PR converts existing INI from in-tree
to external CCM INI.

Also adds the INI section for zonal labels
if FailureDomains is defined and has a length
other than 0.


jcpowermac:
Still continuing on this PR.  Need to work on node networking section and then tests.

jcpowermac:
@JoelSpeed @rvanderp3 can you take a look when you have a moment, thanks!

pkg/cloud/vsphere/vsphere.go
```diff
@@ -89,3 +95,204 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+				datacenters := strings.Join(vcenter.Datacenters[:], ",")
+				_, err = setOrGetKeyValue(vCenterSection, "datacenters", datacenters, true)
+				if err != nil {
+					return "", err
+				}
+			}
+		}
+
+		if infra.Spec.PlatformSpec.VSphere.FailureDomains != nil {
+			if len(infra.Spec.PlatformSpec.VSphere.FailureDomains) != 0 {
```
JoelSpeed, line 155: Is the first of these if statements actually required?

pkg/cloud/vsphere/vsphere.go
```diff
@@ -89,3 +95,204 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+
+				err = setLabelsSection(cfg)
+				if err != nil {
+					return "", err
+				}
```
JoelSpeed, line 160: Inline the error decl/check and wrap the error coming out please

pkg/cloud/vsphere/vsphere.go
```diff
@@ -89,3 +95,204 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+				return err
+			}
+		}
+	}
+
+	// if there are no new or existing keys on [Nodes] section
+	// [Nodes] should not exist, delete section
+	if len(nodesSection.Keys()) == 0 {
+		cfg.DeleteSection("Nodes")
+	}
```
JoelSpeed, line 266: Should this go before the previous loop? And return early?

pkg/cloud/vsphere/vsphere.go
```diff
@@ -89,3 +95,204 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+	return nil
+}
+
+func setOrGetSection(cfg *ini.File, name string) (*ini.Section, error) {
+	if cfg.HasSection(name) {
+		klog.Infof("%s section found...", name)
+		return cfg.GetSection(name)
+	} else {
+		klog.Infof("Creating new section: %s", name)
+		return cfg.NewSection(name)
```
JoelSpeed, line 277: Logs here are probably too chatty at `V(0)`, this is pretty debug level IMO

pkg/cloud/vsphere/vsphere.go
```diff
@@ -89,3 +95,204 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+	}
+
+	if infra.Spec.PlatformSpec.VSphere != nil {
+		err = setNodesSection(cfg, &infra.Spec.PlatformSpec.VSphere.NodeNetworking)
+		if err != nil {
+			return "", err
+		}
+
+		if infra.Spec.PlatformSpec.VSphere.VCenters != nil {
+			if len(infra.Spec.PlatformSpec.VSphere.VCenters) != 0 {
```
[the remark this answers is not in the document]
jcpowermac: Nope, I was being cautious, removed

pkg/cloud/vsphere/vsphere.go
```diff
@@ -89,3 +95,204 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+				// delete all existing [VirtualCenter] sections
+				// they will be recreated below
+				for _, section := range cfg.Sections() {
+					if strings.Contains(section.Name(), "VirtualCenter") {
+						cfg.DeleteSection(section.Name())
+					}
+				}
```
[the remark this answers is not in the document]
jcpowermac: I was thinking the infra object is the single source of truth and the 3cmo sets that
If that is wrong I will certainly change it

pkg/cloud/vsphere/vsphere.go
jcpowermac: Maybe you have a suggestion here. 
I need to know if the node networking struct has any values assigned. I couldn't find an easy/quick of checking all struct parameters so I figured loop over the map, if there are no ini section keys I know no values were set in the struct and just delete that section.

pkg/cloud/vsphere/vsphere.go
jcpowermac: fixed

jcpowermac:
fyi @gnufied

jcpowermac:
I will work on this tomorrow morning to get back in shape

pkg/cloud/vsphere/vsphere_test.go
```diff
@@ -59,3 +70,338 @@ func TestResourcesRenderingSmoke(t *testing.T) {
...
+secret-namespace = "kube-system"
+insecure-flag = "1"
+
+[Workspace]
+server = "vcenter.sddc-44-236-21-251.vmwarevmc.com"
+datacenter = "SDDC-Datacenter"
+default-datastore = "WorkloadDatastore"
+folder = "/SDDC-Datacenter/vm/jcallen"
+
+[VirtualCenter "vcenter.sddc-44-236-21-251.vmwarevmc.com"]
```
lobziik, line 260: Is this virtual center expected to be removed from resulted config?

jcpowermac: I changed the test to use more generic names. The `[VirtualCenter]` section should remain. The only section removed will be `[Workspace]`.

jcpowermac: @lobziik I double checked too, purposefully made the structs unequal to make sure the test worked. 
I probably should add some test that fail.

PR #210, opened 2022-12-16
OCPCLOUD-1807: Port to ginkgo v2

Fedosin:
This commit ports tests to ginkgo v2 framework.


lobziik (commented):
Could we enable coverage and junit report?
Also, seems ginkgo v1 is still vendored, is this needed?

JoelSpeed:
We will sort the coverage and junit in a follow up
/lgtm
/approve

PR #213, opened 2023-01-02
Enable lazy loading for rest mapper

Fedosin:
Instead of implementing a custom rest mapper that filters what groups we need, it's better to use the lazy loading mapper.


pkg/restmapper/lazyrestmapper.go
```diff
@@ -89,14 +168,14 @@ func fetchGroupVersionResources(d discovery.DiscoveryInterface, apiGroups []*met
...
-func filteredServerGroupsAndResources(d discovery.DiscoveryInterface, groupFilterPredicate GroupFilterPredicate) ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
+func (m *LazyRESTMapper) filteredServerGroupsAndResources(d discovery.DiscoveryInterface) ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
 	sgs, err := d.ServerGroups()
 	if sgs == nil {
 		return nil, nil, err
 	}
 	resultGroups := []*metav1.APIGroup{}
 	for i := range sgs.Groups {
-		if groupFilterPredicate(&sgs.Groups[i]) {
+		if m.isGroupKnown((&sgs.Groups[i]).Name) {
```
lobziik, line 178: I have a feeling, that for every new group will happen `len(KnownGroups)` number of requests, since it makes a request for every group down below.

pkg/restmapper/lazyrestmapper.go
```diff
@@ -10,40 +10,119 @@ import (
...
+		return nil, err
+	}
+
+	return &LazyRESTMapper{
+		Mapper: restmapper.NewDiscoveryRESTMapper([]*restmapper.APIGroupResources{}),
+		Client: dc,
+	}, nil
+}
+
+func (m *LazyRESTMapper) AddKnownGroupAndReload(groupName string) error {
```
lobziik, line 34: Could this be private?

Also, I would appreciate docstring. :)

lobziik (changes requested):
I like the idea to produce a new mapper, but i'm a bit concerned about potential unnecessary requests and the fact that LazyMapper exposes everything without apparent reasons.

pkg/restmapper/lazyrestmapper.go
```diff
@@ -151,3 +230,13 @@ func getFilteredAPIGroupResources(cl discovery.DiscoveryInterface, groupFilterPr
 	}
 	return result, nil
 }
+
+func (m *LazyRESTMapper) isGroupKnown(groupName string) bool {
+	for _, knownGroup := range m.KnownGroups {
```
lobziik, line 235: Just thought, perhaps `KnownGroups` might be a map? Just thought, no strict necessary here.

pkg/restmapper/lazyrestmapper.go
lobziik, line 34: Is this function thread-safe?

pkg/restmapper/lazyrestmapper.go
```diff
@@ -10,40 +10,124 @@ import (
...
-			),
-		)
-		if err != nil {
-			return nil, err
+// LazyRESTMapper is a RESTMapper that will lazily query the provided
+// client for discovery information to do REST mappings.
+type LazyRESTMapper struct {
+	mapper      meta.RESTMapper
+	client      *discovery.DiscoveryClient
+	knownGroups map[string]any
```
lobziik, line 20: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set

Might using this make code more readable? Just a thought, I'm ok having a map here.

pkg/restmapper/lazyrestmapper.go
```diff
@@ -10,40 +10,124 @@ import (
...
+	}, nil
+}
+
+func (m *LazyRESTMapper) addKnownGroupAndReload(groupName string) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	m.knownGroups[groupName] = struct{}{}
+
+	groupResources, err := getFilteredAPIGroupResources(m.client, m.knownGroups)
```
lobziik, line 46: Hm, is it possible to get `[]*restmapper.APIGroupResources` from the existing mapper, append the `APIGroupResources` for the single group there, and reinitialize the mapper? This might allow not to make requests for all groups in the `knownGroups` and make it just for a new one.

Or, alternatively, `[]*restmapper.APIGroupResources` which comes from `getFilteredAPIGroupResources` might be cached in the `LazyRESTMapper` somehow (raw idea: perhaps we could use `knownGroups` map for this, like `map[string]*restmapper.APIGroupResources`).

pkg/restmapper/lazyrestmapper.go
```diff
@@ -89,14 +173,14 @@ func fetchGroupVersionResources(d discovery.DiscoveryInterface, apiGroups []*met
 // Mainly replicate ServerGroupsAndResources function from the client-go. The difference is that this function takes
 // a function of the GroupFilterPredicate type as an argument for filtering out unwanted groups.
 // ref: https://github.com/kubernetes/kubernetes/blob/a84d877310ba5cf9237c8e8e3218229c202d3a1e/staging/src/k8s.io/client-go/discovery/discovery_client.go#L383
-func filteredServerGroupsAndResources(d discovery.DiscoveryInterface, groupFilterPredicate GroupFilterPredicate) ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
+func filteredServerGroupsAndResources(d discovery.DiscoveryInterface, knownGroups map[string]any) ([]*metav1.APIGroup, []*metav1.APIResourceList, error) {
```
lobziik, line 176: WDYT about renaming this? `groupsAndResourcesForGroups` or smth like this maybe

pkg/restmapper/lazyrestmapper.go
```diff
@@ -123,8 +207,8 @@ func filteredServerGroupsAndResources(d discovery.DiscoveryInterface, groupFilte
 
 // getFilteredAPIGroupResources uses the provided discovery client to gather
 // discovery information and populate a slice of APIGroupResources.
-func getFilteredAPIGroupResources(cl discovery.DiscoveryInterface, groupFilterPredicate GroupFilterPredicate) ([]*restmapper.APIGroupResources, error) {
-	gs, rs, err := filteredServerGroupsAndResources(cl, groupFilterPredicate)
+func getFilteredAPIGroupResources(cl discovery.DiscoveryInterface, knownGroups map[string]any) ([]*restmapper.APIGroupResources, error) {
```
lobziik, line 210: suggest renaming this, `getAPIGroupResources` would be better, IMHO

lobziik (commented):
I think it would be better to get rid of `Filtered` notion in function names. Previously it was there because this thing expected predicates for groups filtering, now it is not actual.

pkg/restmapper/lazyrestmapper.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		res, err = m.mapper.KindsFor(resource)
+	}
+
+	return res, err
+}
+
+// ResourceFor implements Mapper.ResourceFor.
+func (m *LazyRESTMapper) ResourceFor(input schema.GroupVersionResource) (schema.GroupVersionResource, error) {
```
elmiko, line 75: any reason not to name the `input` argument as `resource` like the other functions?

pkg/restmapper/lazyrestmapper.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		res, err = m.mapper.ResourceFor(input)
+	}
+
+	return res, err
+}
+
+// ResourcesFor implements Mapper.ResourcesFor.
+func (m *LazyRESTMapper) ResourcesFor(input schema.GroupVersionResource) ([]schema.GroupVersionResource, error) {
```
elmiko, line 89: similar question about `input` here as well

elmiko (commented):
this generally looks good to me and make sense, i just had a question about some of the argument names for consistency

pkg/restmapper/lazyrestmapper.go
elmiko: i noticed after posting this review that the function is finding the "resource", so making the argument name the same doesn't necessarily make as much sense, but i wonder if we should name `gvr` for all the functions, similar to how we have `gk` in the mapper functions.

PR #215, opened 2023-01-10
Change verbosity level for some logs

Fedosin:
To prevent showing too many less important messages in the log we set the verbosity level for them to 1.

Here is an [example](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-cloud-controller-manager-operator/213/pull-ci-openshift-cluster-cloud-controller-manager-operator-master-e2e-gcp-ovn-ccm-install/1612583627818799104/artifacts/e2e-gcp-ovn-ccm-install/gather-extra/artifacts/pods/openshift-cloud-controller-manager-operator_cluster-cloud-controller-manager-operator-5547c7d677-t68hk_config-sync-controllers.log) of such messages.

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -36,7 +36,7 @@ type CloudConfigReconciler struct {
 }
 
 func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	klog.Infof("Syncing cloud-conf ConfigMap")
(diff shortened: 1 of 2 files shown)


JoelSpeed (commented):
Perhaps rather than removing these, we can just increase the log level to say `V(1)` or `V(2)`?

lobziik (commented):
I would rather move these log lines into a higher verbosity level instead of complete removal.

Fedosin:
Okay, done! I set it to 1 for those calls.

PR #218, opened 2023-01-12
OCPCLOUD-1817: vSphere cloud-config yaml converter 

lobziik:
This PR is the follow-up for https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/206 which introduces vsphere cloud-config transformer for supporting config supplementing with zones and networking information.

An approach in this PR is slightly different and introduces conversion from deprecated INI config format to YAML along the way, which seem helps to simplify logic for networking/zonal info population a bit.

For working purposes, I just used tests from #206 for now. In further commits I'm going to refactor this a bit and remove dependency for the vsphere ccm.

cc @jcpowermac 

UPD: During team internal discussion we decided that it would be ok to have a dependency for vsphere ccm for tests at the moment. This will help us to ensure that config conversion code produces valid CCM configs.

TODO:

- [x] Docstrings and comments
~- [ ] remove dependency on vsphere ccm~
- [x] more tests


pkg/cloud/vsphere/vsphere_cloud_config/config.go
```diff
@@ -0,0 +1,44 @@
...
+
+	klog.V(3).Info("Try to parse vSphere config, yaml format first")
+	cfg, err := readCPIConfigYAML(config)
+	if err != nil {
+		klog.Warningf("Parsing yaml config failed, fallback to ini: %s", err)
+
+		cfg, err = readCPIConfigINI(config)
+		if err != nil {
+			klog.Errorf("ini config parsing failed: %s", err)
+			return nil, err
```
JoelSpeed, line 28: Do you want to wrap the error?

pkg/cloud/vsphere/vsphere_cloud_config/config.go
```diff
@@ -0,0 +1,44 @@
...
+// Unlike 'cloud-provider-vsphere' version of a similar function, this does ignore environment variables.
+func ReadConfig(config []byte) (*CPIConfig, error) {
+	if len(config) == 0 {
+		return nil, fmt.Errorf("vSphere config is empty")
+	}
+
+	klog.V(3).Info("Try to parse vSphere config, yaml format first")
+	cfg, err := readCPIConfigYAML(config)
+	if err != nil {
+		klog.Warningf("Parsing yaml config failed, fallback to ini: %s", err)
```
JoelSpeed, line 23: I think the idiom for errors is to use `%v` in logs and `%w` when in a `fmt.Errorf`
```suggestion
		klog.Warningf("Parsing yaml config failed, fallback to ini: %v", err)
```

pkg/cloud/vsphere/vsphere_cloud_config/config.go
```diff
@@ -0,0 +1,44 @@
...
+
+		cfg, err = readCPIConfigINI(config)
+		if err != nil {
+			klog.Errorf("ini config parsing failed: %s", err)
```
JoelSpeed, line 27: Idiom
```suggestion
			klog.Errorf("ini config parsing failed: %v", err)
```

pkg/cloud/vsphere/vsphere_cloud_config/config.go
```diff
@@ -0,0 +1,44 @@
...
+			return nil, err
+		}
+
+		klog.V(3).Info("ini config parsed successfully")
```
JoelSpeed, line 31: We don't have a counterpart for successful YAML parsing?

pkg/cloud/vsphere/vsphere_cloud_config/config_test.go
```diff
@@ -0,0 +1,182 @@
...
+port = -1
+ip-family = "ipv6"
+`
+
+func TestINIConfigConversion(t *testing.T) {
+
+	t.Run("basic config yaml conversion", func(t *testing.T) {
+		g := gmg.NewWithT(t)
+		configStruct, err := ReadConfig([]byte(basicConfigINI))
+		g.Expect(err).Should(gmg.Succeed())
```
JoelSpeed, line 107: Nit: This should read "expect err to not have occurred"
```suggestion
		g.Expect(err).ToNot(gmg.HaveOccurred())
```

pkg/cloud/vsphere/vsphere_cloud_config/config_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+		convertedConfig, err := MarshalConfig(configStruct)
+		g.Expect(err).Should(gmg.Succeed())
```
JoelSpeed, line 110: As above, error checks should be "to not have occurred", please fix the other occurrences too
```suggestion
		g.Expect(err).ToNot(gmg.HaveOccurred())
```

pkg/cloud/vsphere/vsphere_cloud_config/config_test.go
```diff
@@ -0,0 +1,182 @@
...
+		g.Expect(VC2.VCenterIP).To(gmg.Equal("10.0.0.2"))
+		g.Expect(VC2.IPFamilyPriority).To(gmg.Equal([]string{}))
+
+		VC3 := configStruct.Vcenter["10.0.0.3"]
+		g.Expect(len(VC3.Datacenters)).To(gmg.Equal(2))
+		g.Expect(VC3.Datacenters).To(gmg.BeComparableTo([]string{"DC5", "DC6"}))
+		g.Expect(VC3.IPFamilyPriority).To(gmg.Equal([]string{"ipv6"}))
+
+		_, err = MarshalConfig(configStruct)
+		g.Expect(err).Should(gmg.Succeed())
```
JoelSpeed, line 152: Not checking the expected output of marshal config on this one?

pkg/cloud/vsphere/vsphere_cloud_config/config_test.go
```diff
@@ -0,0 +1,182 @@
...
+			"empty", "", "vSphere config is empty",
+		},
+	}
+
+	for _, tc := range invalidConfigTestCases {
+		t.Run(fmt.Sprintf("test invalid config: %s", tc.name), func(t *testing.T) {
+			g := gmg.NewWithT(t)
+			_, err := ReadConfig([]byte(tc.input))
+			g.Expect(err).Should(gmg.HaveOccurred())
+			g.Expect(err.Error()).To(gmg.ContainSubstring(tc.errSubstring))
```
JoelSpeed, line 179: You can shorten this, use `MatchError` to check the string value of the error, it will fail if no error occurs or if the error message doesn't match the next matcher
```suggestion
			g.Expect(err).To(gmg.MatchError(gmg.ContainSubstring(tc.errSubstring)))
```

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+func CloudConfigTransformer(source string, infra *configv1.Infrastructure, _ *configv1.Network) (string, error) {
+	if infra.Status.PlatformStatus == nil ||
+		infra.Status.PlatformStatus.Type != configv1.VSpherePlatformType {
+		return "", fmt.Errorf("invalid platform, expected to be %s", configv1.VSpherePlatformType)
+	}
+
+	cpiCfg, err := ccmConfig.ReadConfig([]byte(source))
+
```
JoelSpeed, line 24: Empty new line?

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -0,0 +1,88 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+
+	ccmConfig "github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/vsphere/vsphere_cloud_config"
+)
+
+const (
+	regionLabelValue = "openshift-region"
+	zoneLabelValue   = "openshift-zone"
+)
```
JoelSpeed, line 15: I'd like some comments on these, explain what they are, why they are these values, why they can't be configured by a user etc, right now they look like magic consts without any context

pkg/cloud/vsphere/vsphere_config_transformer_test.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			g := gmg.NewWithT(t)
+			infraResouce := tc.infraBuilder.Build()
+			transformedConfig, err := CloudConfigTransformer(tc.inputConfig, infraResouce, makeDummyNetworkConfig())
+			if tc.errMsg != "" {
+				g.Expect(err).Should(gmg.HaveOccurred())
+				g.Expect(err.Error()).Should(gmg.ContainSubstring(tc.errMsg))
```
JoelSpeed, line 333: Can shorten this using `MatchError`

pkg/cloud/vsphere/vsphere_config_transformer_test.go
```diff
@@ -0,0 +1,355 @@
...
+				return
+			}
+
+			// Using CPI config reader from cloud-provider-vsphere
+			// to ensure that config transformation produces valid yaml config which
+			// will be readable and usable by the CCM then
+
+			wantConfig, err := ccm.ReadCPIConfig([]byte(tc.equivalentConfig))
+			if err != nil {
+				g.Expect(err).Should(gmg.Equal(nil))
```
JoelSpeed, line 343: That's not how we normally check errors?

pkg/cloud/vsphere/vsphere_config_transformer_test.go
```diff
@@ -0,0 +1,355 @@
...
+				return
+			}
+			gotConfig, err := ccm.ReadCPIConfig([]byte(transformedConfig))
+			if err != nil {
+				g.Expect(err).Should(gmg.Equal(nil))
+				return
+			}
+
+			g.Expect(gotConfig).Should(gmg.BeComparableTo(wantConfig))
```
JoelSpeed, line 352: Why `BeComparableTo` vs `Equal` here?

JoelSpeed (commented):
I've left a bunch of nits and a few questions but otherwise this LGTM in general, can @jcpowermac please take a look as well to make sure it's still doing what he expects regarding the conversion from infrastructure to config

pkg/cloud/vsphere/vsphere_config_transformer_test.go
lobziik: My main reasoning was a nice diff which`BeComparableTo` spits out.

Speaking about differences - under the hood `BeComparableTo` uses [go-cmp](https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal) which seems to be ok alternative to `DeepEqual` for tests.

pkg/cloud/vsphere/vsphere_cloud_config/config_test.go
lobziik: Nope, my intention was to ensure that the config struct will be marshaled without errors (sort of smoke test).

As I understand ordering would not be guaranteed for `vcenter` section and if I will compare it with some sort of expected string it might be flaky. 
At the moment of writing, I thought that just checking errors should be enough and extra effort for checking output here is not worth it.

pkg/cloud/vsphere/vsphere_cloud_config/config.go
lobziik: Wrapped, removed redundant log msg

pkg/cloud/vsphere/vsphere_config_transformer.go
lobziik: added, PTAL

pkg/cloud/vsphere/vsphere_cloud_config/config.go
```diff
@@ -0,0 +1,45 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+// ReadConfig parses vSphere cloud-config file and returns CPIConfig structure
+// Accepts both YAML and INI formats as input.
+// YAML format takes precedence, in case parsing YAML is not successful function falls back to the legacy INI format.
+// Unlike 'cloud-provider-vsphere' version of a similar function, this does ignore environment variables.
+func ReadConfig(config []byte) (*CPIConfig, error) {
+	if len(config) == 0 {
+		return nil, fmt.Errorf("vSphere config is empty")
```
JoelSpeed, line 17: Nit: Without a format string, should use `errors.New`
```suggestion
		return nil, errors.New("vSphere config is empty")
```

pkg/cloud/vsphere/vsphere_cloud_config/config_test.go
JoelSpeed: Please change this error line to "to not have occurred"

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -0,0 +1,92 @@
...
+// Such tags are meant to be on vSphere resources, such as clusters and datacenters to figure out and properly set up
+// K8s topology labels on node objects.
+// https://github.com/kubernetes/cloud-provider-vsphere/blob/release-1.25/docs/book/cloud_config.md#labels
+// https://github.com/openshift/enhancements/blob/f6b33eb0cd4ba060af71fee6192297cf6bc31e5a/enhancements/installer/vsphere-ipi-zonal.md#cloud-config
+const (
+	regionLabelValue = "openshift-region"
+	zoneLabelValue   = "openshift-zone"
+)
+
+func CloudConfigTransformer(source string, infra *configv1.Infrastructure, _ *configv1.Network) (string, error) {
```
JoelSpeed, line 22: Nit public functions should have a godoc

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -0,0 +1,92 @@
...
+
+// setNodes sets Nodes section in vsphere-cloud-provider config according passed VSpherePlatformNodeNetworking spec
+func setNodes(cfg *ccmConfig.CPIConfig, nodeNetworking *configv1.VSpherePlatformNodeNetworking) {
+	cfg.Nodes.ExternalVMNetworkName = nodeNetworking.External.Network
+	cfg.Nodes.ExternalNetworkSubnetCIDR = strings.Join(nodeNetworking.External.NetworkSubnetCIDR[:], ",")
+	cfg.Nodes.ExcludeExternalNetworkSubnetCIDR = strings.Join(nodeNetworking.External.ExcludeNetworkSubnetCIDR[:], ",")
+
+	cfg.Nodes.InternalVMNetworkName = nodeNetworking.Internal.Network
+	cfg.Nodes.InternalNetworkSubnetCIDR = strings.Join(nodeNetworking.Internal.NetworkSubnetCIDR[:], ",")
+	cfg.Nodes.ExcludeInternalNetworkSubnetCIDR = strings.Join(nodeNetworking.Internal.ExcludeNetworkSubnetCIDR[:], ",")
```
JoelSpeed, line 58: These are functionally equivalent, is it simpler to exclude the `[:]`?
```suggestion
	cfg.Nodes.ExternalNetworkSubnetCIDR = strings.Join(nodeNetworking.External.NetworkSubnetCIDR, ",")
	cfg.Nodes.ExcludeExternalNetworkSubnetCIDR = strings.Join(nodeNetworking.External.ExcludeNetworkSubnetCIDR, ",")

	cfg.Nodes.InternalVMNetworkName = nodeNetworking.Internal.Network
	cfg.Nodes.InternalNetworkSubnetCIDR = strings.Join(nodeNetworking.Internal.NetworkSubnetCIDR, ",")
	cfg.Nodes.ExcludeInternalNetworkSubnetCIDR = strings.Join(nodeNetworking.Internal.ExcludeNetworkSubnetCIDR, ",")
```

JoelSpeed (commented):
Few more minor comments

PR #221, opened 2023-01-23
[OCPCLOUD-1815]: Modernize tooling

odvarkadaniel:
This commit introduces modernization of the approach to update auxiliary tooling to be more consistent such as in Control Plane Machine Set repository. We align envtest, controller-gen and other tooling as well as in the CPMS repo.


pkg/controllers/cache.go
```diff
@@ -81,7 +81,7 @@ func (n *objectWatcher) watch(ctx context.Context, obj client.Object) error {
 
 	// Add an event handler that only allows events through for the correct object name
 	// Since the informer is namespace bound, this should limit the events from this event handler to a single resource.
-	informer.AddEventHandler(&eventToChannelHandler{
+	_, _ = informer.AddEventHandler(&eventToChannelHandler{
```
lobziik, line 84: i suppose `AddEventHandler` returns an error, can we propagate this instead suppressing?

go.mod
```diff
@@ -1,53 +1,44 @@
...
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d
 	github.com/go-logr/logr v1.2.3
 	github.com/golangci/golangci-lint v1.49.0
-	github.com/onsi/ginkgo/v2 v2.6.0
-	github.com/onsi/gomega v1.24.0
-	github.com/openshift/api v0.0.0-20230111143458-54592eea5539
-	github.com/openshift/library-go v0.0.0-20230112164258-24668b1349e6
-	github.com/spf13/cobra v1.5.0
+	github.com/onsi/ginkgo/v2 v2.7.0
+	github.com/onsi/gomega v1.24.2
```
lobziik, line 10: gomega 1.25 was already released

lobziik (changes requested):
Can we please exclude k8s dependencies bump from this pr?

Also, I wondering about XML reporting. The last time I experimented with it, it was broken (only one suite out of dozen was covered in this report). Could we check what will happen in case some test failing in a suite that is not covered by junit report.

odvarkadaniel:
> Can we please exclude k8s dependencies bump from this pr?
> 
> Also, I wondering about XML reporting. The last time I experimented with it, it was broken (only one suite out of dozen was covered in this report). Could we check what will happen in case some test failing in a suite that is not covered by junit report.

I can separate the k8s deps into another PR.

AFAIK, this was broken in ginkgov2 v2.6.x, but already solved in v2.7.0.

lobziik:
>AFAIK, this was broken in ginkgov2 v2.6.x, but already solved in v2.7.0.

IIRC this was coverage.

odvarkadaniel:
> > AFAIK, this was broken in ginkgov2 v2.6.x, but already solved in v2.7.0.
> 
> IIRC this was coverage.

I will investigate it.

pkg/controllers/cache.go
odvarkadaniel: This change will be covered in an upcoming PR for k8s packages update. This is not an issue with the old version k8s having deps with older version golangci-lint.

go.mod
odvarkadaniel: Do you prefer me to update it in this PR or have it in the PR for k8s packages update?

go.mod
```diff
@@ -1,53 +1,44 @@
...
-	github.com/stretchr/testify v1.8.0
+	github.com/stretchr/testify v1.8.1
 	gopkg.in/gcfg.v1 v1.2.3
 	gopkg.in/ini.v1 v1.66.6
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.25.2
-	k8s.io/apiextensions-apiserver v0.25.1
-	k8s.io/apimachinery v0.25.2
-	k8s.io/client-go v0.25.2
+	k8s.io/api v0.26.1
```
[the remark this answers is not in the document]
odvarkadaniel: Rebase this PR. Separate PR will come for the k8s bump.

go.mod
lobziik: Here will be ok i think

JoelSpeed:
/approve

I'm ok with this and i think increasing the test time is an ok addition, I suspect changing the tooling introduces some additional overhead within the test containers so CPU limitations are probably making these tests a little slower

lobziik (commented):
>Also, I wondering about XML reporting. The last time I experimented with it, it was broken (only one suite out of dozen was covered in this report). Could we check what will happen in case some test failing in a suite that is not covered by junit report.

Here is the ginkgo issue: https://github.com/onsi/ginkgo/issues/1115
I'm ok to merge it as is, but we need to track JUnit reporting issue somewhere.

Will hold this for now, @JoelSpeed feel free to unhold, if we do not want to block this due to complications around reporting.

/lgtm
/hold

JoelSpeed:
/hold cancel

I'd rather we get this merged and revisit later for the JUnit issue

PR #223, opened 2023-02-07
OCPBUGS-6658: Clear useManagedIdentityExtension if it's set

jsafrane:
Add `--disable-identity-extension-auth` cmdline option to set `"useManagedIdentityExtension"` to false, because the purpose of the injector is to inject username + password into `cloud.conf` and therefore we want to use that credentials when talking to the cloud. `"useManagedIdentityExtension: true"` would use identity of the Azure VM.

cc @openshift/storage @Fedosin @lobziik

### DIFF
--- a/cmd/azure-config-credentials-injector/main.go
+++ b/cmd/azure-config-credentials-injector/main.go
@@ -14,8 +14,9 @@ const (
 	clientIDEnvKey     = "AZURE_CLIENT_ID"
 	clientSecretEnvKey = "AZURE_CLIENT_SECRET"
 
-	clientIDCloudConfigKey     = "aadClientId"
-	clientSecretCloudConfigKey = "aadClientSecret"
+	clientIDCloudConfigKey               = "aadClientId"
+	clientSecretCloudConfigKey           = "aadClientSecret"
+	useManagedIdentityExtensionConfigKey = "useManagedIdentityExtension"
 )
 
 var (
@@ -26,8 +27,9 @@ var (
 	}
 
 	injectorOpts struct {
-		cloudConfigFilePath string
-		outputFilePath      string
+		cloudConfigFilePath          string
+		outputFilePath               string
+		disableIdentityExtensionAuth bool
 	}
 )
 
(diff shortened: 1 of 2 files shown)


lobziik:
/approve

looks good to me

Would be great if someone else will have a quick look and lgtmed. cc @JoelSpeed @elmiko

cmd/azure-config-credentials-injector/main.go
```diff
@@ -91,6 +94,16 @@ func readCloudConfig(path string) (map[string]interface{}, error) {
 func prepareCloudConfig(cloudConfig map[string]interface{}, clientId string, clientSecret string) ([]byte, error) {
 	cloudConfig[clientIDCloudConfigKey] = clientId
 	cloudConfig[clientSecretCloudConfigKey] = clientSecret
+	if value, found := cloudConfig[useManagedIdentityExtensionConfigKey]; found {
+		if injectorOpts.disableIdentityExtensionAuth {
+			fmt.Printf("%s cleared\n", useManagedIdentityExtensionConfigKey)
```
JoelSpeed, line 99: Is this debug? Or intended as a log?

cmd/azure-config-credentials-injector/main.go
```diff
@@ -91,6 +94,16 @@ func readCloudConfig(path string) (map[string]interface{}, error) {
...
+			cloudConfig[useManagedIdentityExtensionConfigKey] = false
+		} else {
+			if value == true {
+				fmt.Printf("Warning: %s is set to \"true\", injected credentials may not be used\n", useManagedIdentityExtensionConfigKey)
```
JoelSpeed, line 103: Likewise, is this meant as a log? Do we not have any logging libraries initialised here?

elmiko (commented):
looks good to me aside from Joel's questions

cmd/azure-config-credentials-injector/main.go
jsafrane: Both messages are indeed logs, I added `klog/v2` as the logger to the binary. It prints something like this now:

```
W0208 11:17:12.025252    7724 main.go:104] Warning: useManagedIdentityExtension is set to "true", injected credentials may not be used
```

JoelSpeed:
/lgtm

Thanks for updating the logging

PR #224, opened 2023-02-13
vSphere: If only one failuredomain do not config labels

jcpowermac:
Changes to the installer to GA zonal
always create at least one failuredomain.
If labels are set and the tag or tag
categories do not exist installation boostrap
fails. To support the legacy non-zonal install
labels will only be applied when the length of
failure domains is greater than one.


jcpowermac:
Tested in VMC w/installer changes: https://github.com/openshift/installer/pull/6770
Confirm this change resolves the issue.

jcpowermac:
cc: @rvanderp3 @lobziik

PR #227, opened 2023-02-15
OCPBUGS-5036: Restart pods if related configuration was changed

lobziik:
To ensure that operands are using the most recent configurations,
such as cloud configs or credentials, this PR introduces checks for related config content.

This PR  hash calculation for secrets and configmaps content if they are detected in deployment/daemonset pod template spec.

For deployments and daemonsets additional routine was added within the resourceApply procedure. For each reconciliation, it will calculate a hash for configmaps and secrets content which was referenced within PodTemplateSpec.


pkg/controllers/resourceapply/config_hash.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	for _, container := range spec.Spec.Containers {
+		collectRelatedConfigsFromContainer(&container, &sources)
+	}
+
+	return sources
+}
+
+func collectRelatedConfigsFromContainer(container *v1.Container, sources *configSources) {
```
elmiko, line 57: does this func need a comment as well?

pkg/controllers/resourceapply/config_hash_test.go
```diff
@@ -0,0 +1,271 @@
...
+				},
+			},
+			expectedSecrets:    []string{"envSecret", "envSecret2", "secret", "secret2"},
+			expectedConfigMaps: []string{"configmap", "configmap2", "envConfigMap", "envConfigMap2"},
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			g := gmg.NewWithT(t)
```
elmiko, line 186: i haven't used this function before, what does it do here?

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -129,13 +129,17 @@ func applyConfigMap(ctx context.Context, client coreclientv1.Client, recorder re
 
 func applyDeployment(ctx context.Context, client coreclientv1.Client, recorder record.EventRecorder, requiredOriginal *appsv1.Deployment) (bool, error) {
 	required := requiredOriginal.DeepCopy()
-	err := SetSpecHashAnnotation(&required.ObjectMeta, required.Spec)
-	if err != nil {
+	if err := annotatePodSpecWithRelatedConfigsHash(ctx, client, required.Namespace, &required.Spec.Template); err != nil {
+		klog.V(5).Infof("Can not check related configs and secrets: %w", err)
```
elmiko, line 133: i looked through the code and we seem to have logs at levels `1`, `2`, `3`, and `4`. i wonder if we should maybe use 3 or 4 for these?

imo, 3 looks like a good target for our failure logs (see pkg/cloud/vsphere/vsphere_cloud_config/config.go)

pkg/controllers/resourceapply/resourceapply.go
```diff
@@ -129,13 +129,17 @@ func applyConfigMap(ctx context.Context, client coreclientv1.Client, recorder re
...
+		recorder.Event(required, corev1.EventTypeWarning, "Can not check related configs", err.Error())
```
elmiko, line 134: when making events, the [general convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events) for the "reason" is to use a single CamelCase word, and then use something more free form for the "message" part.  in this case we might do something like:

```suggestion
		recorder.Event(required, corev1.EventTypeWarning, "FailedConfigurationCheck", fmt.Sprintf("Can not check related configs, error: %v", err.Error()))
```

and similar on the other events too

elmiko (commented):
code mostly makes sense to me, the hashing stuff can be brittle sometimes but the code and tests seem good.

i did have a question about log levels and a suggestion for the events.

pkg/controllers/resourceapply/config_hash_test.go
lobziik: Intializes gomega matcher for using things like `Expect`.

https://pkg.go.dev/github.com/onsi/gomega#NewWithT

pkg/controllers/resourceapply/resourceapply.go
lobziik: I updated loglevel, and also added info about the resource itself there.

https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/227/files#diff-8a4ec169e12f51d19bc1aae0acb142b300939b46e8337ee467621b8f09eff6c6R148

pkg/controllers/resourceapply/resourceapply.go
lobziik: Good suggestion. Replaced all reasons in this file with constants, added that as separate commit: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/227/commits/c2fdaa97c429ab2f05769a4bca20b1b287dcb4ca

elmiko (commented):
updates look nice to me, thanks!
/lgtm

lobziik:
/cc @eurijon
/cc @JoelSpeed 

Re-requesting review breaks review requests for some reason... :/

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -111,6 +111,10 @@ spec:
             items:
               - key: cloud.conf
                 path: cloud.conf
+        - name: secret-vccm
+          secret:
+            secretName: {{ .globalCredsSecretName }}
+            optional: true
```
JoelSpeed, line 117: What is vccm?

pierreprinetti:
FYI `e2e-openstack-ovn` is only failing for a known unrelated reason.

pkg/controllers/clusteroperator_controller.go
```diff
@@ -196,7 +196,9 @@ func (r *CloudOperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&source.Kind{Type: &operatorv1.KubeControllerManager{}},
 			handler.EnqueueRequestsFromMapFunc(toClusterOperator),
 			builder.WithPredicates(kcmPredicates())).
-		Watches(&source.Channel{Source: watcher.EventStream()}, handler.EnqueueRequestsFromMapFunc(toClusterOperator))
+		Watches(&source.Channel{Source: watcher.EventStream()}, handler.EnqueueRequestsFromMapFunc(toClusterOperator)).
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, handler.EnqueueRequestsFromMapFunc(toClusterOperator)).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(toClusterOperator))
```
JoelSpeed, line 201: So any configmap or secret change means we re-reconcile the cluster operator, which in turn leads us into the loop that manages the deployments?

pkg/controllers/resourceapply/config_hash.go
```diff
@@ -0,0 +1,136 @@
+package resourceapply
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
```
JoelSpeed, line 9: corev1 idiomatically, can we use the common name?

pkg/controllers/resourceapply/config_hash.go
```diff
@@ -0,0 +1,136 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const configsHashAnnotation = "operator.openshift.io/configs-hash"
```
JoelSpeed, line 17: I would make this `configHash` rather than `configsHash`
```suggestion
const configHashAnnotation = "operator.openshift.io/config-hash"
```

pkg/controllers/resourceapply/config_hash.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, container := range spec.Spec.Containers {
+		collectRelatedConfigsFromContainer(&container, &sources)
+	}
+
+	return sources
+}
+
+// collectRelatedConfigsFromContainer collects related configs names into passed configSources instance.
+// Looks into env and envVar of the passed container spec and populates configSources with configmaps and secrets names.
+func collectRelatedConfigsFromContainer(container *v1.Container, sources *configSources) {
```
JoelSpeed, line 59: I don't know that either of these inputs need to be pointers do they? We are only reading from the container and a configSources reference should still have the same pointers to the same maps as in the parent scope for the sets

pkg/controllers/resourceapply/config_hash.go
```diff
@@ -0,0 +1,136 @@
...
+	hashBytes := sha256.Sum256(hashSourceBytes)
+	return fmt.Sprintf("%x", hashBytes), nil
+}
+
+// annotatePodSpecWithRelatedConfigsHash annotates pod template spec with a hash of related config maps and secrets content.
+func annotatePodSpecWithRelatedConfigsHash(ctx context.Context, cl runtimeclient.Client, ns string, spec *v1.PodTemplateSpec) error {
+	sources := collectRelatedConfigSources(spec)
+	hash, err := calculateRelatedConfigsHash(ctx, cl, ns, sources)
+	if err != nil {
+		return err
```
JoelSpeed, line 129:
```suggestion
		return fmt.Errorf("error calculating configuration hash: %w", err)
```

pkg/controllers/resourceapply/config_hash_test.go
```diff
@@ -0,0 +1,271 @@
...
+			},
+		},
+		{
+			Name: "bar", VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{SecretName: "secret2"},
+			},
+		},
+		{
+			Name: "baz", VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{SecretName: "secret"},
```
JoelSpeed, line 53: Deliberately sharing a name with volume foo?

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
lobziik: vsphere cloud controller manager

pkg/controllers/clusteroperator_controller.go
lobziik: Yes. 

Actually, we do not care about a specific request content, reconcile loop is common.

Also, as far as I can say the only changes that matters are in `openshift-cloud-controller-manager` namespace. 
https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/cmd/cluster-cloud-controller-manager-operator/main.go#L116

pkg/controllers/resourceapply/config_hash.go
lobziik: >I don't know that either of these inputs need to be pointers do they?

I made these pointers to avoid copying basically. For `configSources` I agree, it doesn't matter. For containers, hm, I'm not so sure.

pkg/controllers/resourceapply/config_hash_test.go
lobziik: Yes

pkg/controllers/resourceapply/config_hash.go
lobziik: Looked into this again and recalled my thoughts a bit more:

`container` is a pointer to avoid unnecessary struct copying, and `sources` is mutating inside.


No strict necessity to have it as pointers, but I would like to leave it as is.

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
JoelSpeed: What's the motivation behind adding this new secret, I don't see quite how that's related to the goal of this PR? Is it that the vSphere CCM normally looks up the creds but we need a signal to the logic that this secret is being used?

pkg/controllers/resourceapply/config_hash.go
JoelSpeed: IMO it's simpler for these not to be pointers, the data we are copying is relatively small and adding pointers adds complexity, as developers we now have to think about it more, so I would say it's preferable to be references on both. Go has a good GC, I'm not worried about data copying

pkg/controllers/resourceapply/config_hash_test.go
JoelSpeed: I think a comment about why this is deliberately sharing the name would be useful

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
lobziik: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml#L63

ccm reads this secret, and need to reroll deployment if credentials changed. This is a sorta crutch to achieve this.

pkg/controllers/resourceapply/config_hash.go
lobziik: I do not agree that it adds complexity. 

First, this function is internal to the module and is not intended to be exposed.

Second, this code will be called for every reconciliation, and I do not want to have a `Container` copy there.

Third, pointer to `sources` + docstring indicates that `sources` will be mutated. Rely on the knowledge that there are sets inside and they are references - this adds complexity IMHO.

JoelSpeed: I'm ok to continue with it as you prefer, but if this creates confusion in the future, I reserve the right to say I told you so 😉 
Agree to disagree as it were

elmiko (commented):
changes look good to me, i am also slightly confused by the vsphere secret update. i see that vsphere needs it, just seems a little confusing as part of this pr.

/lgtm

pkg/controllers/resourceapply/config_hash.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+
+	if len(errList) > 0 {
+		return "", errors.NewAggregate(errList)
+	}
+
+	hashSourceBytes, err := json.Marshal(hashSource)
+	if err != nil {
+		return "", fmt.Errorf("unable to marshal dependant config content into JSON: %v", err)
```
JoelSpeed, line 117: Nit, should be `%w`

JoelSpeed (commented):
/approve

pierreprinetti:
Is this going to be backported?

PR #229, opened 2023-03-02
[release-4.10] OCPBUGS-6039: Try to limit groups for the REST mapper discovery

lobziik:
Manual cherry-pick of https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/216

Old version of gomega `HaveField` matcher does not support pointers to a struct. I bumped gomega version in separate commit. 

This gomega update affect only unittests and should not affect application itself.


elmiko (commented):
/lgtm

PR #230, opened 2023-03-02
OCPCLOUD-1779: Prevent operator to provision operands if platform is set to 'External'

lobziik:
Operator should not provision anything in case if 'External' platform is encountered.


elmiko (commented):
/approve
/lgtm

PR #233, opened 2023-03-08
OCPCLOUD-1779, OCPBUGS-8694: Set missed operator status in case of the 'External' platform type encountered

lobziik:
Follow up for
- https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/230

After #230 merged, we encountered that there is no operator status reported to cluseroperator object.
This PR adds missed status reporting.


elmiko (commented):
thanks @lobziik !

/lgtm

lobziik:
/override ci/prow/e2e-aws-ovn-ccm

This job exercising tech preview cluster, but AWS CCM was marked as GA in 4.14 and this job was removed from the list of required. Weirdly, prow still wants it to pass for some reason.

PR #238, opened 2023-03-22
WIP: Check operands presence in case if operator determined that provisioning is not allowed

lobziik:
If operator determined that provisioning is not allowed, operands should be removed.

This required for potential necessity to rollback back to the in-tree cloud provider on vSphere platform
due to potential CSI migration issues.

### DIFF
--- a/pkg/controllers/clusteroperator_controller.go
+++ b/pkg/controllers/clusteroperator_controller.go
@@ -25,8 +25,11 @@ import (
 	"github.com/openshift/library-go/pkg/cloudprovider"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	errutils "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


pkg/controllers/clusteroperator_controller.go
```diff
@@ -149,6 +170,47 @@ func (r *CloudOperatorReconciler) sync(ctx context.Context, config config.Operat
...
 
+// operandsProvisioned checks if resources from the passed slice are provisioned
+func (r *CloudOperatorReconciler) operandsProvisioned(ctx context.Context, resources []client.Object) (bool, error) {
+	var errList []error
+	// use delete with dry run for avoid occasional resource mutation in the passed slice
+	opts := &client.DeleteOptions{
+		DryRun: []string{metav1.DryRunAll},
+	}
+	for _, resource := range resources {
+		err := r.Client.Delete(ctx, resource, opts)
```
elmiko, line 181: should this be a `Get` ?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -110,7 +105,38 @@ func (r *CloudOperatorReconciler) Reconcile(ctx context.Context, _ ctrl.Request)
...
+		klog.Errorf("Unable to determine cluster state to check if provision is allowed: %v", err)
+		return ctrl.Result{}, err
+	} else if !allowedToProvision {
+		operandsProvisioned, err := r.operandsProvisioned(ctx, resources)
+		if err != nil {
+			klog.Errorf("Unable to determine if operands are already provisioned: %v", err)
+		}
+		if operandsProvisioned {
+			if err := r.deleteOperands(ctx, resources); err != nil {
+				if err := r.setStatusDegraded(ctx, err); err != nil {
```
elmiko, line 129: will we need to become un-degraded once all the resources are removed?

elmiko (commented):
i think this generally makes sense, couple of comments

pkg/controllers/clusteroperator_controller.go
lobziik: On next reconciliation this status will be set to available during the check in the `provisioningAllowed ` procedure

lobziik: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/238/files#diff-1e56759e83cb848f28cc71546c3151724a2bc26915e59e279ab8f702858a91d9R327

elmiko:
it sounds like we are not going to follow the reversion path, i am going to close this PR as i don't think we'll need it.

/close

PR #239, opened 2023-03-29
OCPBUGS-7898: add a check for nutanix cloud conf map

elmiko:
When upgrading from 4.12 to 4.13 on the nutanix platform, the cloud-conf configmap must be present to avoid a degradation. This change adds a check in the cloud_config_sync_controller reconcile to check the platform type and configmap existence, if it does not exist on nutanix then the upgradeable condition will be set to false.

ref: https://issues.redhat.com/browse/OCPBUGS-7898

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -55,6 +55,36 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, err
 	}
 
+	// Due to an upgrade issue from 4.12 -> 4.13 on Nutanix, a user will need to manually create the necessary
+	// cloud configuration ConfigMap. We need to check here and set the operator's upgradeable status to
+	// false if the configmap does not exist.
+	// We are checking here because Nutanix does not require a sync in this version and the controller will return early otherwise.
+	// See https://issues.redhat.com/browse/OCPBUGS-7898 for more information.
+	if cmNeeded, err := r.isProviderNutanixAndCloudConfigNeeded(ctx, infra.Status.PlatformStatus); cmNeeded {
+		// The ConfigMap does not exist, set upgradeable to false and return
+		reason := "MissingNutanixConfigMap"
+		message := "Cloud Config Controller is not upgradeable due to missing \"cloud-conf\" ConfigMap"
+		if err := r.setUpgradeableCondition(ctx, configv1.ConditionFalse, reason, message); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
+		}
+		return ctrl.Result{}, nil
+	} else if err != nil {
+		// An error occurred while checking, set degraged and return the error.
+		if err := r.setDegradedCondition(ctx); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
+		}
+		return ctrl.Result{}, err
+	} else if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.Type == configv1.NutanixPlatformType {
+		// we are on Nutanix platform and the ConfigMap exists, ensure that the operator is upgradeable
+		if err := r.setUpgradeableCondition(ctx, configv1.ConditionTrue, ReasonAsExpected, ""); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
+		}
+
+		// we will always return early on Nutanix, but as this patch is intended for 4.12 only it will be fine
+		// because Nutanix does not have a CCM in 4.12 and this operator will be replaced by the incoming 4.13 version.
+		return ctrl.Result{}, nil
+	}
+
 	syncNeeded, err := r.isCloudConfigSyncNeeded(infra.Status.PlatformStatus, infra.Spec.CloudConfig)
 	if err != nil {
 		if err := r.setDegradedCondition(ctx); err != nil {
@@ -307,4 +337,47 @@ func (r *CloudConfigReconciler) setDegradedCondition(ctx context.Context) error 
 	co.Status.Versions = []configv1.OperandVersion{{Name: operatorVersionKey, Version: r.ReleaseVersion}}
 	klog.Info("Cloud Config Controller is degraded")
 	return r.syncStatus(ctx, co, conds)
+}
+
+func (r *CloudConfigReconciler) setUpgradeableCondition(ctx context.Context, condition configv1.ConditionStatus, reason string, message string) error {
+	co, err := r.getOrCreateClusterOperator(ctx)
+	if err != nil {
+		return err
+	}
+
+	conds := []configv1.ClusterOperatorStatusCondition{
+		newClusterOperatorStatusCondition(configv1.OperatorUpgradeable, condition, reason, message),
+	}
+
+	co.Status.Versions = []configv1.OperandVersion{{Name: operatorVersionKey, Version: r.ReleaseVersion}}
+	if condition == configv1.ConditionFalse {
+		klog.Info("Cloud Config Controller is not upgradeable")
+	} else {
+		klog.Info("Cloud Config Controller is upgradeable")
+	}
+	return r.syncStatus(ctx, co, conds)
+}
+
+// isProviderNutanixAndCloudConfigNeeded will determine whether the proper cloud configuration ConfigMap is present
+// on the Nutanix platform. This function is being added to mitigate an upgrade issue where the user must create
+// the ConfigMap before the upgrade can progress.
+// See https://issues.redhat.com/browse/OCPBUGS-7898 for more information.
+func (r *CloudConfigReconciler) isProviderNutanixAndCloudConfigNeeded(ctx context.Context, platformStatus *configv1.PlatformStatus) (bool, error) {
+	if platformStatus == nil || platformStatus.Type != configv1.NutanixPlatformType {
+		return false, nil
+	}
+
+	cm := &corev1.ConfigMap{}
+	// the name of this ConfigMap is sourced from the Nutanix asset for the volume mount
+	// See: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/release-4.13/pkg/cloud/nutanix/assets/cloud-controller-manager-deployment.yaml#L107
+	if err := r.Get(ctx, client.ObjectKey{Name: syncedCloudConfigMapName, Namespace: r.ManagedNamespace}, cm); errors.IsNotFound(err) {
+		// confirmed that the ConfigMap does not exist.
+		klog.Warningf("ConfigMap \"cloud-conf\" is not found, and is required for upgrade on Nutanix platform")
+		return true, nil
+	} else if err != nil {
+		// got an unexpected error, report false and return the error
+		return false, err
+	}
+
+	return false, nil
 }
--- a/pkg/controllers/cloud_config_sync_controller_test.go
+++ b/pkg/controllers/cloud_config_sync_controller_test.go
@@ -484,3 +484,98 @@ var _ = Describe("Cloud config sync reconciler", func() {
 		}
 	})
 })
+
+var _ = Describe("Cloud config sync isProviderNutanixAndCloudConfigNeeded", func() {
+	// This is just testing the isProviderNutanixAndCloudConfigNeeded function
+	var reconciler *CloudConfigReconciler
+
+	ctx := context.Background()
+	targetNamespaceName := testManagedNamespace
+
+	BeforeEach(func() {
+		reconciler = &CloudConfigReconciler{
+			ClusterOperatorStatusClient: ClusterOperatorStatusClient{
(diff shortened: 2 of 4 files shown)


pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -308,3 +327,45 @@ func (r *CloudConfigReconciler) setDegradedCondition(ctx context.Context) error
...
+		return err
+	}
+
+	conds := []configv1.ClusterOperatorStatusCondition{
+		newClusterOperatorStatusCondition(cloudConfigControllerAvailableCondition, configv1.ConditionTrue, ReasonAsExpected,
+			"Cloud Config Controller works as expected"),
+		newClusterOperatorStatusCondition(cloudConfigControllerDegradedCondition, configv1.ConditionFalse, ReasonAsExpected,
+			"Cloud Config Controller works as expected"),
+		newClusterOperatorStatusCondition(configv1.OperatorUpgradeable, configv1.ConditionFalse, ReasonNotUpgradeable,
+			"Cloud Config Controller is not upgradeable"),
```
JoelSpeed, line 343: I wonder if we can add a string to the signature here so that we can decorate this message with a bit more useful information?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -308,3 +327,45 @@ func (r *CloudConfigReconciler) setDegradedCondition(ctx context.Context) error
...
+	klog.Info("Cloud Config Controller is not upgradeable")
+	return r.syncStatus(ctx, co, conds)
+}
+
+// isProviderNutanixAndCloudConfigNeeded will determine whether the proper cloud configuration ConfigMap is present
+// on the Nutanix platform. This function is being added to mitigate an upgrade issue where the user must create
+// the ConfigMap before the upgrade can progress.
+// See https://issues.redhat.com/browse/OCPBUGS-7898 for more information.
+func (r *CloudConfigReconciler) isProviderNutanixAndCloudConfigNeeded(ctx context.Context, platformStatus *configv1.PlatformStatus) (bool, error) {
+	if platformStatus != nil && platformStatus.Type == configv1.NutanixPlatformType {
```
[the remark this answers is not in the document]
elmiko: i can switch it like that, when i start writing functions i test to just put my expected return at the bottom and then start filling in the rest. it's more a stylistic artifact than anything.

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -484,3 +484,98 @@ var _ = Describe("Cloud config sync reconciler", func() {
...
+			Scheme: scheme.Scheme,
+		}
+	})
+
+	It("should return false if no PlatformStatus in infra resource presented", func() {
+		infraResource := makeInfrastructureResource(configv1.NutanixPlatformType)
+		Expect(cl.Create(ctx, infraResource)).To(Succeed())
+
+		needed, err := reconciler.isProviderNutanixAndCloudConfigNeeded(ctx, infraResource.Status.PlatformStatus)
+		Expect(err).To(BeNil())
```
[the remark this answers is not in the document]
elmiko: thank you, i could not for the life of me remember that lol. i copied the current code from another section of tests XD

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -484,3 +484,98 @@ var _ = Describe("Cloud config sync reconciler", func() {
...
+		Expect(needed).To(BeFalse())
```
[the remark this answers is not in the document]
elmiko: good catch

pkg/controllers/cloud_config_sync_controller.go
elmiko: hmm, making that text more specific doesn't quite feel right, what if the text directs users to check the logs?

pkg/controllers/cloud_config_sync_controller.go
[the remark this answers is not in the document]
elmiko: writing out the early return still requires a return at the end, i'll post the updated code but i think it looks more busy. probably just a style thing though.

pkg/controllers/cloud_config_sync_controller.go
JoelSpeed: If the text can be varied by the caller, I think it's ok, but checking logs also works

elmiko: part of thing is, this code will only exist for 4.12.z and only gets engaged for nutanix

if you are cool with adding a comment to look at the logs, i think that makes sense to me

wking:
I'm moved SPLAT-984 from `blocks` to `relates to` temporarily to avoid slowing down this  pull request on that.

/jira refresh

wking:
And I moved `OCPBUGS-11177` from  `Fixed` to `Done`.

/jira refresh

pkg/controllers/status.go
```diff
@@ -20,10 +20,11 @@ import (
...
 const (
-	ReasonAsExpected   = "AsExpected"
-	ReasonInitializing = "Initializing"
-	ReasonSyncing      = "SyncingResources"
-	ReasonSyncFailed   = "SyncingFailed"
+	ReasonAsExpected     = "AsExpected"
+	ReasonInitializing   = "Initializing"
+	ReasonSyncing        = "SyncingResources"
+	ReasonSyncFailed     = "SyncingFailed"
+	ReasonNotUpgradeable = "NotUpgradeable"
```
wking, line 27: nit: doesn't really matter for a one-off, but `Upgradeable=False` has the type/status already saying "not upgradeable".  Ideally `reason` adds some context like `MissingNutanixConfigMap` that's specific to the thing we're asking the admin to fix.

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -55,6 +55,25 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
+		}
+		return ctrl.Result{}, nil
+	} else if err != nil {
+		// An error occurred while checking, set degraged and return the error.
+		if err := r.setDegradedCondition(ctx); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
+		}
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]
elmiko: this is going to sound weird, but there is another function running in the main controller's Reconcile that will set the operator as available on the next reconcile. so it should drop through here, and then get picked up by the main controller..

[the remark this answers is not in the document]
elmiko: although, now that i think about it, we will need that function to preserve the upgradeable bits.

pkg/controllers/status.go
elmiko: thanks, i wasn't sure about this one. i'll make it more descriptive/specific.

pkg/controllers/cloud_config_sync_controller.go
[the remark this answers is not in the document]
elmiko: i added some extra code to the main controller where we set available status to preserve the upgradeable flag. hopefully that will work for everyone.

elmiko:
updated to address comments, also added some logic to preserve the upgradeable flag in the main controller.

pkg/controllers/status.go
```diff
@@ -115,12 +115,24 @@ func (r *ClusterOperatorStatusClient) setStatusAvailable(ctx context.Context) er
 		return err
 	}
 
+	// we want to preserve the upgradeable flag in case the sync controller has marked us as not upgradeable
+	upgradeable := configv1.ConditionTrue
+	upgReason := ReasonAsExpected
+	upgMessage := ""
```
wking, line 121: nit: would be nice to follow [this pattern][1] and include some encouraging words here like `No concerns about updating.`, so users don't have to look at the machine-oriented `reason` to understand what's going on.  Although the `status` is pretty clear, so no worries if you actively prefer to leave `message` empty.

[1]: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/188d382d413cb1bd2a37b2b7dc17d0d58430634e/pkg/controllers/cloud_config_sync_controller.go#L283-L287

elmiko: i was more just following the pattern of the previous code

elmiko:
updated with a few refinements in the setStatusAvailable function

wking (approved):
/lgtm

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -308,3 +338,46 @@ func (r *CloudConfigReconciler) setDegradedCondition(ctx context.Context) error
 	klog.Info("Cloud Config Controller is degraded")
 	return r.syncStatus(ctx, co, conds)
 }
+
+func (r *CloudConfigReconciler) setUpgradeableCondition(ctx context.Context, condition configv1.ConditionStatus, reason string, message string) error {
```
lobziik, line 342: nit: in future would be nice to extract this functions to separate file, along with others like set degraded and place it next to each other.

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -308,3 +338,46 @@ func (r *CloudConfigReconciler) setDegradedCondition(ctx context.Context) error
...
+// See https://issues.redhat.com/browse/OCPBUGS-7898 for more information.
+func (r *CloudConfigReconciler) isProviderNutanixAndCloudConfigNeeded(ctx context.Context, platformStatus *configv1.PlatformStatus) (bool, error) {
+	if platformStatus == nil || platformStatus.Type != configv1.NutanixPlatformType {
+		return false, nil
+	}
+
+	cm := &corev1.ConfigMap{}
+	// the name of this ConfigMap is sourced from the Nutanix asset for the volume mount
+	// See: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/release-4.13/pkg/cloud/nutanix/assets/cloud-controller-manager-deployment.yaml#L107
+	if err := r.Get(ctx, client.ObjectKey{Name: syncedCloudConfigMapName, Namespace: r.ManagedNamespace}, cm); errors.IsNotFound(err) {
```
lobziik, line 373: Hm, will this cause issues on first reconciliation when cloud config is not yet synced? Code which calls it runs before the config sync logic. Or customer should be place config into ccm namespace directly?

lobziik (commented):
Honestly, I'm not sure how it will play with other config sync logic...

I'm bit afraid about what happened with logic after the nutanix check if user placed cloud config into ccm namespace, but there will be no references in infra resource.

pkg/controllers/cloud_config_sync_controller.go
elmiko: i had a similar thought, but i am going to make a jira card about cleaning up these functions we have split across 2 different controller files.

pkg/controllers/cloud_config_sync_controller.go
elmiko: i don't think so because this is looking in the same namespace as the operator. if it doesn't find the configmap it will mark upgradeable=false, but on the next reconcile if the cm exists it will go to true.

elmiko:
> I'm bit afraid about what happened with logic after the nutanix check if user placed cloud config into ccm namespace, but there will be no references in infra resource.

i think it will be ok because there isn't a nutanix ccm for 4.12, so this logic will only exist until the operator is replaced by the 4.13 version. then the new ccm and operator will take over.

elmiko:
@sunzhaohua2 ptal for cherry-pick-approval

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -308,3 +338,46 @@ func (r *CloudConfigReconciler) setDegradedCondition(ctx context.Context) error
...
 }
+
+func (r *CloudConfigReconciler) setUpgradeableCondition(ctx context.Context, condition configv1.ConditionStatus, reason string, message string) error {
+	co, err := r.getOrCreateClusterOperator(ctx)
+	if err != nil {
+		return err
+	}
+
+	conds := []configv1.ClusterOperatorStatusCondition{
+		newClusterOperatorStatusCondition(configv1.OperatorUpgradeable, configv1.ConditionFalse, reason, message),
```
yanhua121, line 349: I think it should put the input argument "condition" in the conds, instead of the hard-coded configv1.ConditionFalse. This code line should be: 

newClusterOperatorStatusCondition(configv1.OperatorUpgradeable, condition, reason, message),

elmiko: yes, good spot @yanhua121 . i will update

elmiko:
updated to fix issue with `setUpgradeableCondition`

petr-muller:
/payload-job periodic-ci-openshift-release-master-nightly-4.12-e2e-nutanix-ovn

elmiko:
i think i see the issue that is causing the installation failures. the main available conditions are not being added. i need to think about how to rework this.

elmiko:
updated to make sure we set the available status even when we are not processing a CCM for the platform.

PR #240, opened 2023-04-03
OCPBUGS-11284: Add beta topology labels flag to Azure cloud node manager

JoelSpeed:
This enables the feature flag created in https://github.com/openshift/cloud-provider-azure/pull/62

We need this to preserve the behaviour of applying beta topology labels which have not yet been removed by openshift


JoelSpeed:
Azure looks good
/approve

PR #245, opened 2023-04-19
CCO-324,CCO-325: add support for workload identity

RomanBednar:
In scope of [Azure workload identity feature](https://docs.google.com/document/d/18y0pq0Jq0AgvCymb-4dwso_pwEKbUeXyLFRDW_JwgBA/edit#) we need to extend injector to parse new configuration values.

### DIFF
--- a/cmd/azure-config-credentials-injector/main.go
+++ b/cmd/azure-config-credentials-injector/main.go
@@ -52,12 +52,12 @@ func mergeCloudConfig(_ *cobra.Command, args []string) error {
 		return err
 	}
 
-	azureClientId, found := os.LookupEnv(clientIDEnvKey)
+	azureClientId, found := mustLookupEnvValue(clientIDEnvKey)
 	if !found {
 		return fmt.Errorf("%s env variable should be set up", clientIDEnvKey)
 	}
 
-	azureClientSecret, found := os.LookupEnv(clientSecretEnvKey)
+	azureClientSecret, secretFound = mustLookupEnvValue(clientSecretEnvKey)
 	if !found {
 		return fmt.Errorf("%s env variable should be set up", clientSecretEnvKey)
 	}
@@ -119,4 +119,12 @@ func writeCloudConfig(path string, preparedConfig []byte) error {
 		return err
 	}
 	return nil
+}
+
+func mustLookupEnvValue(key string) (string, bool) {
+	value, found := os.LookupEnv(key)
+	if !found || len(value) == 0 {
+		return "", false
+	}
+	return value, true
 }
--- a/cmd/azure-config-credentials-injector/main.go
+++ b/cmd/azure-config-credentials-injector/main.go
@@ -11,12 +11,18 @@ import (
 )
 
 const (
-	clientIDEnvKey     = "AZURE_CLIENT_ID"
-	clientSecretEnvKey = "AZURE_CLIENT_SECRET"
+	clientIDEnvKey       = "AZURE_CLIENT_ID"
+	clientSecretEnvKey   = "AZURE_CLIENT_SECRET"
+	tenantIDEnvKey       = "AZURE_TENANT_ID"
+	federatedTokenEnvKey = "AZURE_FEDERATED_TOKEN_FILE"
 
 	clientIDCloudConfigKey               = "aadClientId"
 	clientSecretCloudConfigKey           = "aadClientSecret"
 	useManagedIdentityExtensionConfigKey = "useManagedIdentityExtension"
+
+	tenantIdConfigKey                              = "tenantId"
+	aadFederatedTokenFileConfigKey                 = "aadFederatedTokenFile"
+	useFederatedWorkloadIdentityExtensionConfigKey = "useFederatedWorkloadIdentityExtension"
 )
 
 var (
@@ -29,6 +35,7 @@ var (
 	injectorOpts struct {
 		cloudConfigFilePath          string
 		outputFilePath               string
+		enableWorkloadIdentity       string
 		disableIdentityExtensionAuth bool
 	}
 )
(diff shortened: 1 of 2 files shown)


RomanBednar:
@abutcher can you please take a look as well?

abutcher:
> So I'm wondering, do we still want to prioritize client secret like this? Can we even expect to get a secret from CCO where we have both? And if we do should we rather just "pass through" all the values and let driver/cloud provider pick the auth method?


CCO should never provide a secret which has both client secret and workload identity fields set. If priority is a concern I think it would be okay to prioritize workload identity if both fields are present; or perhaps if the federatedTokenFile is set then it should be assumed that workload identity is requested.

RomanBednar:
> CCO should never provide a secret which has both client secret and workload identity fields set. If priority is a concern I think it would be okay to prioritize workload identity if both fields are present; or perhaps if the federatedTokenFile is set then it should be assumed that workload identity is requested.

@abutcher Thanks for the details, assuming workload identity method sounds reasonable to me. I understand both methods should never be set by CCO, but we can have a check here, just in case. I've updated the PR again - it's ready for review now.

elmiko (commented):
i think this is mostly making sense to me, although i do not know the azure credentials system at a deep level. i do have a question about one of the log messages.

cmd/azure-config-credentials-injector/main.go
```diff
@@ -57,17 +63,26 @@ func mergeCloudConfig(_ *cobra.Command, args []string) error {
...
-	if !found {
-		return fmt.Errorf("%s env variable should be set up", clientSecretEnvKey)
+	// Handle error cases for authentication methods. We should configure exactly one method.
+	tenantId, federatedTokenFile, err := lookupWorkloadIdentityEnv(tenantIDEnvKey, federatedTokenEnvKey)
+	azureClientSecret, secretFound := os.LookupEnv(clientSecretEnvKey)
+	if err == nil && secretFound {
+		// Workload identity is available, but client secret is provided too.
+		// If we don't fail here we would use client secret unnecessarily, also we can assume that user intended to use workload identity because all required values are provided.
+		return fmt.Errorf("%s env variable is set along with env variables %s and %s which indicate workload identity auth should be used, this should never happen\n"+
+			"Please consider reporting a bug to CCO team: https://github.com/openshift/cloud-credential-operator/issues", clientSecretEnvKey, tenantIDEnvKey, federatedTokenEnvKey)
```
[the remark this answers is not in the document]
RomanBednar: You're right, all bug reports should probably land in `issues.redhat.com`, thank you. 

> i think this is mostly making sense to me, although i do not know the azure credentials system at a deep level. i do have a question about one of the log messages.

Here we're mostly just passing through the values we get from CCO, most of the interesting stuff is done [there](https://github.com/openshift/cloud-credential-operator/pull/523) so that's why I wanted @abutcher to review this PR as well.

cmd/azure-config-credentials-injector/main.go
```diff
@@ -57,17 +63,26 @@ func mergeCloudConfig(_ *cobra.Command, args []string) error {
...
+	azureClientSecret, secretFound := os.LookupEnv(clientSecretEnvKey)
+	if err == nil && secretFound {
+		// Workload identity is available, but client secret is provided too.
+		// If we don't fail here we would use client secret unnecessarily, also we can assume that user intended to use workload identity because all required values are provided.
+		return fmt.Errorf("%s env variable is set along with env variables %s and %s which indicate workload identity auth should be used, this should never happen\n"+
+			"Please consider reporting a bug to CCO team: https://issues.redhat.com", clientSecretEnvKey, tenantIDEnvKey, federatedTokenEnvKey)
+	}
+	if err != nil && !secretFound {
+		// Workload identity failed, and we don't have a secret - can't continue.
+		return fmt.Errorf("user+secret method failed: %s is missing; workload identity method failed: %v", clientSecretEnvKey, err)
```
elmiko, line 77: would it be possible to have `err != nil && secretFound` ?

just wondering if we need to cover all cases

cmd/azure-config-credentials-injector/main.go
```diff
@@ -106,6 +121,17 @@ func prepareCloudConfig(cloudConfig map[string]interface{}, clientId string, cli
...
 
+	// Azure workload identity should take precedence, if available.
+	if len(tenantId) != 0 && len(federatedTokenFile) != 0 {
+		cloudConfig[tenantIdConfigKey] = tenantId
+		cloudConfig[aadFederatedTokenFileConfigKey] = federatedTokenFile
+		cloudConfig[useFederatedWorkloadIdentityExtensionConfigKey] = true
+	} else {
+		// Workload identity failed, but we have a secret.
+		klog.V(4).Info("%s env variable is set, client secret authentication will be used", clientSecretEnvKey)
+		cloudConfig[clientSecretCloudConfigKey] = clientSecret
```
elmiko, line 132: do we need to check `len(clientSecret) > 0` ?

elmiko (commented):
just a couple more questions

cmd/azure-config-credentials-injector/main.go
RomanBednar: With the conditions here we rule out the failure cases which are:

1) we have both secret and workload identity 
2) we have neither

So what remains is that we either have secret or we have workload identity. Later we resolve this in `prepareCloudConfig()` where we take one or the other, prioritizing workload identity. So all cases are covered.

So sure we can have the conditions explicitly here, but then I believe we'd have to refactor `prepareCloudConfig()` so it does not do any decisions (which it already does with `useManagedIdentityExtensionConfigKey`). That would make the code more readable perhaps by having the logic at one place - I was tempted to go that route, but we have to consider the time invested to refactoring and value added.

cmd/azure-config-credentials-injector/main.go
RomanBednar: Good point, I'm not sure if we want to pass through `""` - I don't think CCO should let that through. But it does make sense to check for this. But if we do that with secret we should do that with all other values to be consistent. See my latest commit.

RomanBednar:
@elmiko This PR needs one more, hopefully last review. I had to change this to drive the feature using a gate instead of assumptions from variables, the rest should be the same.

PR #249, opened 2023-04-26
OCPBUGS-13547: [OCPCLOUD-2034] Update feature gate observation to use featuregateaccess

JoelSpeed:
This updates the way we look at feature gates to use the new featuregateaccess interface from library-go. This means the decision is now made centrally and not by vendored code in each repository, which will greatly improve the UX for promoting a feature in this repo.

/hold for library-go pin


cmd/cluster-cloud-controller-manager-operator/main.go
```diff
@@ -137,15 +145,62 @@ func main() {
...
+	}
+
+	recorder := events.NewKubeRecorder(kubeClient.CoreV1().Events(*managedNamespace), "cloud-controller-manager-operator", controllerRef)
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		desiredVersion, missingVersion,
+		configInformers.Config().V1().ClusterVersions(), configInformers.Config().V1().FeatureGates(),
+		recorder,
+	)
+
+	featureGateAccessor.SetChangeHandler(func(featureChange featuregates.FeatureChange) {
```
deads2k, line 178: you sure it does?

JoelSpeed: We've been watching and reacting to the feature gate changes since we started this operator, if it couldn't handle the change (eg from nothing to TPNU) then we would have hit issues before no?

JoelSpeed:
/hold

Since this comes so early in the bootstrap process, it relies on the install time feature gate rendering, will need to merge the installer PR before we can merge

JoelSpeed:
/hold cancel

I believe prior issues are now resolved

elmiko (commented):
i'm not super familiar with the featuregate access stuff, but this pr makes sense to me
/lgtm

PR #252, opened 2023-05-22
OCPCLOUD-2052: Rebase to 1.27

odvarkadaniel:
This PR rebases the repo to 1.27 k8s packages.


pkg/controllers/cache.go
```diff
@@ -108,8 +108,10 @@ type eventToChannelHandler struct {
 	name       string
 }
 
-func (e *eventToChannelHandler) OnAdd(obj interface{}) {
-	e.queueEventForObject(nil, obj)
+func (e *eventToChannelHandler) OnAdd(obj interface{}, isInInitialList bool) {
+	if !isInInitialList {
```
JoelSpeed, line 112: Where was this change inspired from? I'm wondering if this could issue for certain operators that expect to always reconcile at first

pkg/controllers/clusteroperator_controller.go
```diff
@@ -189,18 +188,18 @@ func (r *CloudOperatorReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			handler.EnqueueRequestsFromMapFunc(toClusterOperator),
 			builder.WithPredicates(featureGatePredicates())).
-		Watches(&source.Kind{Type: &operatorv1.KubeControllerManager{}},
+		Watches(&operatorv1.KubeControllerManager{},
 			handler.EnqueueRequestsFromMapFunc(toClusterOperator),
 			builder.WithPredicates(kcmPredicates())).
-		Watches(&source.Channel{Source: watcher.EventStream()}, handler.EnqueueRequestsFromMapFunc(toClusterOperator)).
-		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, handler.EnqueueRequestsFromMapFunc(toClusterOperator)).
-		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(toClusterOperator))
+		//Watches(&source.Channel{Source: watcher.EventStream()}, handler.EnqueueRequestsFromMapFunc(toClusterOperator)).
```
JoelSpeed, line 200: Commented out?

pkg/restmapper/restmapper.go
```diff
@@ -10,38 +11,25 @@ import (
...
 			return nil, err
 		}
 
-		return drm, nil
+		groupResources, err := getFilteredAPIGroupResources(dc, groupFilterPredicate)
+		if err != nil {
+			return nil, err
+		}
+
+		return restmapper.NewDiscoveryRESTMapper(groupResources), nil
```
JoelSpeed, line 32: This is because Mikhails updated discovery mapper upstream is merged now?

pkg/restmapper/restmapper_test.go
```diff
@@ -60,11 +60,13 @@ func TestPartialRestMapperProvider(t *testing.T) {
 
 		// Create two different REST mappers with different passed group filter predicates
 		allGroupsRestMapperProvider := NewPartialRestMapperProvider(AllGroups)
-		allGroupsRestMapper, err := allGroupsRestMapperProvider(restCfg)
+		httpClient, err := rest.HTTPClientFor(restCfg)
+		g.Expect(err).To(gmg.BeNil())
```
JoelSpeed, line 64: Idiomatically this would be
```suggestion
		g.Expect(err).ToNot(gmg.HaveOccurred())
```

pkg/controllers/clusteroperator_controller_test.go
```diff
@@ -198,7 +198,7 @@ var _ = Describe("toClusterOperator mapping is targeting requests to 'cloud-cont
 				Name: clusterOperatorName,
 			},
 		}}
-		Expect(toClusterOperator(object)).To(Equal(requests))
+		Expect(toClusterOperator(context.TODO(), object)).To(Equal(requests))
```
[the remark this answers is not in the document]
JoelSpeed: We normally have `var ctx = context.Background()` in the suite, can you add one here, https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/854f5041e39f4174b6552db6f7b4bf8fba2755fe/pkg/controllers/suite_test.go#L59-L61 and then use that instead?

pkg/controllers/cache.go
odvarkadaniel: The `ResourceEventHandler` interface got changed, but as you are mentioning, using that `!isInInitialList` logic doesn't make sense here.

pkg/controllers/clusteroperator_controller.go
odvarkadaniel: Tbh, I don't really know what to do with this channel. Don't you know about a way I could keep the same behavior with the channel somehow being an Object?

JoelSpeed: You can change it to use [this](https://github.com/kubernetes-sigs/controller-runtime/blob/e7b94074ad3814832bbd5f254bb9c5c16a286a35/pkg/builder/controller.go#L174-L187), `WatchesRawSource`, I'm going to trust that we knew what we were doing when we added this event stream

odvarkadaniel: Great, thanks!

odvarkadaniel:
All the remarks mentioned above are addressed in the latest commit.

pkg/controllers/suite_test.go
```diff
@@ -58,6 +58,7 @@ const (
 
 var cfg *rest.Config
 var cl client.Client
+var ctx = context.Background()
```
JoelSpeed, line 61: Bonus points for being alphabetically sorted 👍

JoelSpeed (commented):
/lgtm
/approve

Lets see what the tests think

JoelSpeed:
/hold

@odvarkadaniel I think there's something that needs investigating here, see the failures in the unit test
```
There were failures detected in the following suites:
                              azure ./pkg/cloud/azure [Compilation failure]
                            powervs ./pkg/cloud/powervs [Compilation failure]
                        controllers ./pkg/controllers [Compilation failure]
                         azurestack ./pkg/cloud/azurestack [Compilation failure]
                      resourceapply ./pkg/controllers/resourceapply [Compilation failure]
  azure-config-credentials-injector ./cmd/azure-config-credentials-injector [Compilation failure]
                                aws ./pkg/cloud/aws [Compilation failure]
                          openstack ./pkg/cloud/openstack [Suite did not run because the timeout elapsed]
                              cloud ./pkg/cloud [Compilation failure]
                             common ./pkg/cloud/common [Compilation failure]
                         restmapper ./pkg/restmapper [Suite did not run because the timeout elapsed]
                                ibm ./pkg/cloud/ibm [Compilation failure]
                            vsphere ./pkg/cloud/vsphere [Suite did not run because the timeout elapsed]
                             config ./pkg/config [Compilation failure]
                            alibaba ./pkg/cloud/alibaba [Compilation failure]
                            nutanix ./pkg/cloud/nutanix [Compilation failure]
                                gcp ./pkg/cloud/gcp [Compilation failure]
```

odvarkadaniel:
@JoelSpeed, it seems to be ok now. can u relabel?

JoelSpeed:
CCMO is degraded in the E2E tests now, will need further investigation

odvarkadaniel:
Cluster installs are failing for all providers but aws for some reason. I don't really see why, will look into this tomorrow again.
The issue with unit tests is also pretty random, experienced it a couple of times already while working on PR, though only with CI runs - local runs work fine.

cmd/config-sync-controllers/main.go
```diff
@@ -95,9 +96,13 @@ func main() {
...
 	syncPeriod := 10 * time.Minute
-	cacheBuilder := cache.MultiNamespacedCacheBuilder([]string{
-		*managedNamespace, controllers.OpenshiftConfigNamespace, controllers.OpenshiftManagedConfigNamespace,
-	})
+
+	opts := cache.Options{}
+	opts.Namespaces = []string{*managedNamespace, controllers.OpenshiftConfigNamespace, controllers.OpenshiftManagedConfigNamespace}
+	cacheBuilder := func(*rest.Config, cache.Options) (cache.Cache, error) {
+		return cache.New(restConfig, opts)
+	}
```
JoelSpeed, line 104: Wonder if, now that the multi namespace stuff is included directly, if we should just embed the cache opts into the ctrl.Options directly and let it build out a default cache using these options, suspect that would be equivalent now

odvarkadaniel: I thought this at first as well, but ctrl.Options.Namespace is just a string. We would need it to be []string to work. Wdyt?

cmd/config-sync-controllers/main.go
```diff
@@ -95,9 +95,10 @@ func main() {
 	})
 
 	syncPeriod := 10 * time.Minute
-	cacheBuilder := cache.MultiNamespacedCacheBuilder([]string{
-		*managedNamespace, controllers.OpenshiftConfigNamespace, controllers.OpenshiftManagedConfigNamespace,
-	})
+
+	opts := cache.Options{}
+	opts.Namespaces = []string{*managedNamespace, controllers.OpenshiftConfigNamespace, controllers.OpenshiftManagedConfigNamespace}
```
JoelSpeed, line 100: You could inline all of this below on line 121, lets see if it works first
```
Cache: cache.Options{
  Namespaces: []string{...},
},
```

odvarkadaniel: Will do.

JoelSpeed:
/lgtm

Lets see what the tests think now

JoelSpeed:
/hold cancel

Stale hold

PR #254, opened 2023-05-25
OCPBUGS-13862: add separate upgradeable condition for the sync controller

elmiko:
this change adds a separate condition for the sync controller so that primary controller can more accurately detect when the sync controller wants to mark a blocked upgrade. Separating the upgradeable condition in this manner prevents the primary controller from becoming blocked if it goes non-upgradeable due to its own issues or degraded status.

added a unit test to check that the interaction between the config sync controller and the primary controller is working the way we want.

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -26,8 +26,9 @@ const (
 	defaultConfigKey = "cloud.conf"
 
 	// Controller conditions for the Cluster Operator resource
-	cloudConfigControllerAvailableCondition = "CloudConfigControllerAvailable"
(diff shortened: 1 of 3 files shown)


elmiko:
i think it's worth noting that the upgrade to the unit tests does not specifically test the intermittent degraded condition we are seeing on some clusters, but it seems like i will need to add a much larger test suite to cover that as well. if folks want, i can grow the unit tests more but my concern here is that we are adding stuff to a release branch without bringing it forward to master.

cc @JoelSpeed @sdodson

sdodson:
Setting `jira/valid-bug` and `bugzilla/valid-bug` labels as this code doesn't exist in 4.13+

elmiko:
test failures don't seem to be related to this change, even the upgrade test is waiting on something else.

/test e2e-azure-ovn-upgrade

PR #256, opened 2023-05-25
OPNET-319: Enable CloudDualStackNodeIPs feature gate

mkowalski:
In order to allow passing dual-stack IPs to the `--node-ip` when using cloud provider, we are enabling the CloudDualStackNodeIPs feature gate.

This affects only behaviour of deployments that want to pass 2 IP addresses as `--node-ip` together with `--cloud-provider`. Before K8s 1.27 this was not possible (you can either have dual-stack Node IPs or use a cloud provider, but not both).

Upstream KEP kubernetes/enhancements#3705
adds ability to pass exactly 2 IPs as `--node-ip` also when using cloud provider, but given its alpha stage, feature gate is required here.

Relates-to: OPNET-13
Relates-to: OPNET-230
Fixes: OPNET-319


pkg/cloud/openstack/assets/deployment.yaml
```diff
@@ -74,7 +74,8 @@ spec:
             --leader-elect-lease-duration=137s \
             --leader-elect-renew-deadline=107s \
             --leader-elect-retry-period=26s \
-            --leader-elect-resource-namespace=openshift-cloud-controller-manager
+            --leader-elect-resource-namespace=openshift-cloud-controller-manager \
+            --feature-gates=CloudDualStackNodeIPs=true
```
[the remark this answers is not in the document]
mkowalski: Given that we have the feature gate already in o/api and `/etc/kubernetes/kubelet.conf` has it correctly, is there some boilerplate that would help me doing it in an elegant way, or are we the first ones here to ever try something like that?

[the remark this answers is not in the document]
JoelSpeed: The reconciler already has [access to the feature gates](https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/785ce6738296d66dcf452078337b1c800f0a7fde/cmd/cluster-cloud-controller-manager-operator/main.go#L205), so you need to iterate through the `KnownFeatures()` on that accessor and if they are `Enabled()`, set the value `<gate>=true` or `<gate>=false` if they are not enabled.

That should build you a slice of strings. We already have templating inside each clouds methods so we just need to update that now to include this list of feature gates

mkowalski:
/retitle OPNET-319: Pass feature gates to vSphere and OpenStack providers

pkg/cloud/openstack/assets/deployment.yaml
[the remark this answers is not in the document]
mkowalski: Ok, I did rewrite it a bit. I would appreciate having a look as some stuff may be my arbitrary choice and not necessarily how you'd see things

pkg/util/featuregates.go
```diff
@@ -0,0 +1,33 @@
+package util
+
+import (
+	"strings"
+
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+)
+
+func GetEnabledDisabledFeatures(features featuregates.FeatureGate) ([]string, []string) {
```
JoelSpeed, line 9: Would appreciate some godoc describing this function and the other function

JoelSpeed (commented):
That's much better yes! Exactly what I was imagining. We should probably pass through to all CCMs in the end but this will get us across the line for now

pkg/util/featuregates.go
mkowalski: Of course, I will push as soon as the initial test results arrive (don't want to terminate them now only to push comments)

MaysaMacedo:
Since we didn't get logs from the Builds, I tried one  install on 4.14 with only this PR and got the following on CCM:

`error: invalid argument "AlibabaPlatform=true,CloudDualStackNodeIPs=true,OpenShiftPodSecurityAdmission=true,AWSSecurityTokenService=false,AdmissionWebhookMatchConditions=false,AzureWorkloadIdentity=false,BuildCSIVolumes=false,CSIDriverSharedResource=false,DynamicResourceAllocation=false,EventedPLEG=false,ExternalCloudProvider=false,ExternalCloudProviderAzure=false,ExternalCloudProviderExternal=false,ExternalCloudProviderGCP=false,GCPLabelsTags=false,GatewayAPI=false,InsightsConfigAPI=false,MachineAPIProviderOpenStack=false,MaxUnavailableStatefulSet=false,NodeSwap=false,PDBUnhealthyPodEvictionPolicy=false,PrivateHostedZoneAWS=false,RetroactiveDefaultStorageClass=false,SigstoreImageVerification=false,VSphereStaticIPs=false" for "--feature-gates" flag: unrecognized feature gate: AlibabaPlatform`

Perhaps we shouldn't include all feature gates?

mkowalski:
> Perhaps we shouldn't include all feature gates?

According to https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/ there is no `AlibabaCloud` feature gate anymore. I am not sure why we keep it but I will simply try to now only pass the enabled gates. Hope it helps

MaysaMacedo:
@mkowalski we have ` AlibabaPlatform=true`, so we will likely hit this again

JoelSpeed:
> Perhaps we shouldn't include all feature gates?

Hmm, other controllers just log a warning for the ones that they don't recognise. Which CCM was erroring here? OpenStack?

JoelSpeed:
Looks like we may need to filter the list based on https://github.com/kubernetes/controller-manager/blob/master/pkg/features/kube_features.go

If we use [this function](https://github.com/kubernetes/controller-manager/blob/f45e59bf25372fef0370a6e2677286888ddb36e8/pkg/features/kube_features.go#L57-L59) we can then iterate through the known features and filter out any that aren't relevant for CCM

MaysaMacedo:
> > Perhaps we shouldn't include all feature gates?
> 
> Hmm, other controllers just log a warning for the ones that they don't recognise. Which CCM was erroring here? OpenStack?

Yup, I tested with OpenStack

mkowalski:
> Looks like we may need to filter the list based on https://github.com/kubernetes/controller-manager/blob/master/pkg/features/kube_features.go
> 
> If we use [this function](https://github.com/kubernetes/controller-manager/blob/f45e59bf25372fef0370a6e2677286888ddb36e8/pkg/features/kube_features.go#L57-L59) we can then iterate through the known features and filter out any that aren't relevant for CCM

After some attempts I cannot simply use `k8s.io/controller-manager/features because it operates on top of `MutableFeatureGate` interface which is different from `openshift/library-go/pkg/operator/configobserver/featuregates`; but it got me to this idea that we could keep our own set of white-listed feature gates in `openshift/api/config` and then simply consume it whenever needed, like e.g. when I call `GetEnabledDisabledFeatures`.

If this is an approach we can take, then https://github.com/openshift/api/pull/1514 is the PR in o/api and here in CCM I would do something as simple as

```
func GetEnabledDisabledFeatures(features featuregates.FeatureGate) ([]string, []string) {
[...]
	for _, feature := range features.KnownFeatures() {
		for _, allowed := range v1.CloudPublicFeatureGates {
			if feature == v1.FeatureGateName(allowed) {
				if features.Enabled(feature) {
					enabled = append(enabled, string(feature))
				} else {
					disabled = append(disabled, string(feature))
				}
			}
		}
	}
```

How does it sound?

JoelSpeed:
Keeping a static list of feature gates is likely to rot over time, I'd much rather we use the feature gates, it shouldn't be too hard, you'll just need to create your own mutable feature gate. Something like below should work
```
import (
	"fmt"

	"k8s.io/component-base/featuregate"
)

func main() {
	upstreamFeatureGate := featuregate.NewFeatureGate()

	if err := cloudfeatures.SetupCurrentKubernetesSpecificFeatureGates(upstreamFeatureGate); err != nil {
		panic(err)
	}
	fmt.Printf("KnownFeatures: %v", upstreamFeatureGate.KnownFeatures())
}
```
Once you have those known features you can use that as a filter for the features coming from the `FeatureGate` in OpenShift

pkg/config/config_test.go
```diff
@@ -214,11 +216,19 @@ func TestComposeConfig(t *testing.T) {
...
+			[]configv1.FeatureGateName{"ChocobombBlueberry", "ChocobombBanana"},
+		),
 		expectConfig: OperatorConfig{
 			ManagedNamespace: defaultManagementNamespace,
 			ImagesReference:  defaultImagesReference,
 			PlatformStatus:   &configv1.PlatformStatus{Type: configv1.OpenStackPlatformType},
 			IsSingleReplica:  true,
+			// We only see CloudDualStackNodeIPs returned here because openshift/api/config defines
+			// white-listed features that are allowed to be used by cloud providers. Anything that
+			// is not defined there won't be passed to the cloud provider.
```
JoelSpeed, line 230: Is this comment in need of update?

pkg/util/featuregates.go
```diff
@@ -0,0 +1,56 @@
...
+// in the kubernetes ecosystem.
+func GetEnabledDisabledFeatures(features featuregates.FeatureGate) ([]string, []string) {
+	var enabled []string
+	var disabled []string
+
+	upstreamFeatureGate := upstreamfeature.NewFeatureGate()
+	if err := cloudfeatures.SetupCurrentKubernetesSpecificFeatureGates(upstreamFeatureGate); err != nil {
+		klog.Errorf("can not get feature gates: %v", err)
+		return nil, nil
+	}
```
JoelSpeed, line 26: This should return an error and block when it can't get the feature gates, we shouldn't proceed in this case

pkg/util/featuregates.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	upstreamfeature "k8s.io/component-base/featuregate"
+	cloudfeatures "k8s.io/controller-manager/pkg/features"
+)
+
+// GetEnabledDisabledFeatures returns two slices that contain all the known feature gates
+// and separates them by their enabled/disabled state. It filters features so that we return
+// only those that are to be consumed by the cloud provider instead of all that exist
+// in the kubernetes ecosystem.
+func GetEnabledDisabledFeatures(features featuregates.FeatureGate) ([]string, []string) {
```
JoelSpeed, line 18: Because you're now using this in main, it will filter all the CCM feature gates to be only those upstream. This prevents us using any feature gates that are CCMO specific. I think the filtering logic needs to be moved close to where we inject the feature gates into the CCM configuration

pkg/util/featuregates.go
mkowalski: ACK, will do

pkg/config/config_test.go
mkowalski: Yes, once we settle on the implementation I will fix all docstrings

pkg/util/featuregates.go
JoelSpeed: There are two points where feature gates are used. Firstly, they are logged in the main.go file, here, the feature gates should be the full openshift set. The second is injected into the assets for the different providers. These second ones are the ones that need to be filtered out.
The CCMO itself doesn't care about extra feature gates which means it doesn't need them filtered early like you currently have it

pkg/util/featuregates.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// GetUpstreamCloudFeatureGates returns a list of feature gates that are allowed to be used in the
+// context of cloud provider.
+func GetUpstreamCloudFeatureGates() ([]string, error) {
+	var result []string
+	upstreamFeatureGate := upstreamfeature.NewFeatureGate()
+	err := cloudfeatures.SetupCurrentKubernetesSpecificFeatureGates(upstreamFeatureGate)
+	if err != nil {
+		return nil, err
```
JoelSpeed, line 57: Please wrap this to add additional context
```suggestion
		return nil, fmt.Errorf("could not set feature gates: %w", err)
```

pkg/config/config.go
```diff
@@ -87,13 +91,26 @@ func ComposeConfig(infrastructure *configv1.Infrastructure, clusterProxy *config
 		return OperatorConfig{}, err
 	}
 
+	featureGatesString := ""
+	upstreamGates, err := util.GetUpstreamCloudFeatureGates()
+	if err != nil {
+		klog.Errorf("Unable to get upstream feature gates: %s", err)
+		return OperatorConfig{}, err
```
JoelSpeed, line 98: Please wrap the error adding additional context before passing back

JoelSpeed (commented):
Couple of nits, but otherwise LGTM if it's working now

pkg/util/featuregates.go
mkowalski: Yeah that's right. So my concern is still valid (i.e. what to do with custom-developed feature gates for cloud provider), but as we don't have any instance today (the only flag that we need to pass is upstream one) I defer it to be a problem of future us (hopefully never)

pkg/util/featuregates.go
mkowalski: Makes sense

pkg/config/config.go
mkowalski: In this context we don't wrap errors, look at L80+ of the same file. Do you still want to wrap it here (and then in all the 3 instances of returning `err` in this function, or nowhere?

JoelSpeed: We should be wrapping the errors everywhere, it's a legacy mistake that we aren't in some places

JoelSpeed:
Is there an issue with the unit tests given this change? Looks like they've been running for nearly three hours?

Code wise everything looks good if we can get the tests passing

mkowalski:
They pass locally, I suspect something with the CI and that's why I keep retesting

mkowalski:
The Pod's log says

```
make generate
make[1]: Entering directory '/go/src/github.com/openshift/cluster-cloud-controller-manager-operator'
go run /go/src/github.com/openshift/cluster-cloud-controller-manager-operator/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
make[1]: Leaving directory '/go/src/github.com/openshift/cluster-cloud-controller-manager-operator'
./hack/verify-diff.sh
make manifests
make[1]: Entering directory '/go/src/github.com/openshift/cluster-cloud-controller-manager-operator'
make[1]: 'manifests' is up to date.
make[1]: Leaving directory '/go/src/github.com/openshift/cluster-cloud-controller-manager-operator'
./hack/verify-diff.sh
KUBEBUILDER_ASSETS="/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/bin/k8s/1.27.1-linux-amd64" ./hack/ci-test.sh
go run ./hack/../vendor/github.com/onsi/ginkgo/v2/ginkgo -v --randomize-all --randomize-suites --keep-going --race --trace --timeout=30m --junit-report=junit_cluster-cloud-controller-manager-operator.xml --cover --coverprofile=test-unit-coverage.out --output-dir=/logs/artifacts ./...
=== RUN   TestPartialRestMapperProvider
    restmapper_test.go:15: Setup envtest
[controller-runtime] log.SetLogger(...) was never called, logs will not be displayed:
goroutine 10 [running]:
runtime/debug.Stack()
	/usr/lib/golang/src/runtime/debug/stack.go:24 +0x72
sigs.k8s.io/controller-runtime/pkg/log.eventuallyFulfillRoot()
	/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/vendor/sigs.k8s.io/controller-runtime/pkg/log/log.go:59 +0xdb
sigs.k8s.io/controller-runtime/pkg/log.(*delegatingLogSink).Error(0xc00049c300, {0x25f2fa0, 0xc000232320}, {0x22f8de8, 0x20}, {0xc0003de040, 0x2, 0x2})
	/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/vendor/sigs.k8s.io/controller-runtime/pkg/log/deleg.go:139 +0x78
github.com/go-logr/logr.Logger.Error({{0x260c838?, 0xc00049c300?}, 0x0?}, {0x25f2fa0, 0xc000232320}, {0x22f8de8, 0x20}, {0xc0003de040, 0x2, 0x2})
	/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/vendor/github.com/go-logr/logr/logr.go:299 +0xcb
sigs.k8s.io/controller-runtime/pkg/envtest.(*Environment).startControlPlane(0xc0000e1680)
	/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/vendor/sigs.k8s.io/controller-runtime/pkg/envtest/server.go:331 +0xa5
sigs.k8s.io/controller-runtime/pkg/envtest.(*Environment).Start(0xc0000e1680)
	/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/vendor/sigs.k8s.io/controller-runtime/pkg/envtest/server.go:261 +0xd05
github.com/openshift/cluster-cloud-controller-manager-operator/pkg/restmapper.setupEnvtest(0xc000102b60)
	/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/pkg/restmapper/restmapper_test.go:18 +0xfd
github.com/openshift/cluster-cloud-controller-manager-operator/pkg/restmapper.TestPartialRestMapperProvider(0xc000102b60)
	/go/src/github.com/openshift/cluster-cloud-controller-manager-operator/pkg/restmapper/restmapper_test.go:30 +0x48
testing.tRunner(0xc000102b60, 0x23c7ae8)
	/usr/lib/golang/src/testing/testing.go:1576 +0x217
created by testing.(*T).Run
	/usr/lib/golang/src/testing/testing.go:1629 +0x806
```

so I need to have a deeper look. I still don't believe it's my change, but it is what it is and for some reason tests are not finishing

mkowalski:
/test unit

Locally the same suite succeeds

```
=== RUN   TestPartialRestMapperProvider
    restmapper_test.go:15: Setup envtest
=== RUN   TestPartialRestMapperProvider/getFilteredAPIGroupResources_should_return_APIGroupResources_based_on_passed_predicates
=== RUN   TestPartialRestMapperProvider/NewPartialRestMapperProvider_should_return_different_RESTMapperProviders_based_on_passed_predicates
=== NAME  TestPartialRestMapperProvider
    restmapper_test.go:23: Stop envtest
--- PASS: TestPartialRestMapperProvider (3.53s)
    --- PASS: TestPartialRestMapperProvider/getFilteredAPIGroupResources_should_return_APIGroupResources_based_on_passed_predicates (0.02s)
    --- PASS: TestPartialRestMapperProvider/NewPartialRestMapperProvider_should_return_different_RESTMapperProviders_based_on_passed_predicates (0.02s)
PASS
```

mkowalski:
/test e2e-vsphere-ovn
/test e2e-azure-ovn-upgrade

They didn't fail for a reasonable reason, so let's retry

PR #257, opened 2023-06-01
OCPBUGS-14425: Alibaba platforms should not be upgreadable

JoelSpeed:
Alibaba clusters are TechPreviewNoUpgrade and should not be upgradeable.
This introduces an upgrade blocker via CCMO to prevent alibaba clusters upgrading.

The condition management in this operator sucks and needs refactoring, for now, I've made an overrides concept so that the top level operator can pass these overrides down. We should come up with a better way of doing this.

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -289,7 +289,7 @@ func (r *CloudConfigReconciler) setAvailableCondition(ctx context.Context) error
 
 	co.Status.Versions = []configv1.OperandVersion{{Name: operatorVersionKey, Version: r.ReleaseVersion}}
 	klog.V(1).Info("Cloud Config Controller is available")
-	return r.syncStatus(ctx, co, conds)
+	return r.syncStatus(ctx, co, conds, nil)
 }
 
 func (r *CloudConfigReconciler) setDegradedCondition(ctx context.Context) error {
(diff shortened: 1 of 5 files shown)


JoelSpeed:
/payload-job periodic-ci-openshift-release-master-nightly-4.14-e2e-alibaba-ovn

JoelSpeed:
Can see in the Alibaba payload that it added the condition correctly, however, the test suite is failing because the upgradable condition is set to false, looks like we will need to add an exception to origin to allow for this

elmiko:
> IMO it would be good if the team could sit down and review how we are setting conditions in this operator, there's 3 controllers at the moment and the code is a mess, we don't really have time to do that immediately IMO but we do have a card on the backlog that @elmiko created to review how these work and improve them in the long run

+1, i have _thoughts_ about this. basically, i think we just need to have separate conditions for each controller and then let the primary controller set the canonical operator conditions that are used by the wider openshift mechanisms. at the least, that will allow us to untangle the interwoven nature of what is happening currently.

JoelSpeed:
/payload-job periodic-ci-openshift-release-master-nightly-4.14-e2e-alibaba-ovn

The test should be fixed now

JoelSpeed:
Start all core operators failure is fixed, other issues in the failure seem to be unrelated

damdo (commented):
/lgtm

elmiko (commented):
/approve

PR #258, opened 2023-06-26
OCPBUGS-14696: Update leader election configuration for CCM to match KCM

JoelSpeed:
We are seeing long periods of downtime of the cloud controller manager during upgrades. This means that the load balancer  membership is not being updated when nodes are going down for maintenance during this period. Since this affects disruption metrics for workloads, I'm proposing we make the leader election match the kube controller manager to allow the load balancer membership to be updated more aggressively.

### DIFF
--- a/pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
+++ b/pkg/cloud/alibaba/assets/cloud-controller-manager-deployment.yaml
@@ -66,9 +66,9 @@ spec:
               exec /bin/alibaba-cloud-controller-manager \
               --allow-untagged-cloud=true \
               --leader-elect=true \
-              --leader-elect-lease-duration=137s \
-              --leader-elect-renew-deadline=107s \
-              --leader-elect-retry-period=26s \
+              --leader-elect-lease-duration=15s \
+              --leader-elect-renew-deadline=12s \
+              --leader-elect-retry-period=3s \
               --leader-elect-resource-namespace=openshift-cloud-controller-manager \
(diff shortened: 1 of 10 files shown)


JoelSpeed:
/payload-job periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-aws-ovn-upgrade

JoelSpeed:
/payload-job periodic-ci-openshift-release-master-nightly-4.14-e2e-aws-sdn-upgrade periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-aws-ovn-upgrade

JoelSpeed:
Reviewing yesterdays payload jobs, the AWS disruption for service load balancers was low, only 1 or 2 seconds in each job. I think that's a good result so I think this is a worthwhile change. Should improve reaction times to unready nodes needing to be removed from the load balancer.

JoelSpeed:
/hold

I would like to double check with API folks about the impact of making this more aggressive. I think it's important to avoid workload disruption so hopefully it's justifiable

JoelSpeed:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-aws-ovn-upgrade 10

JoelSpeed:
/hold 

After more research, I don't think this is going to solve the issue highlighted in the bug

JoelSpeed:
Managed to solve the bug without this, I think we should pursue adding release on cancel upstream and then this isn't required anyway

PR #259, opened 2023-07-06
OCPBUGS-15154: Prune provider permissions post move to use-service-account-credentials

JoelSpeed:
We started using `--use-service-account-credentials` which means that the service account for the cloud-controller-manager now only has the responsibility of the leader election lease and some other small bits. Used audit2rbac from a couple of platform CI runs to prune down to what I think is the required set. Will need to test across all platforms we support to validate that this is correct


JoelSpeed:
/test e2e-nutanix-ovn

We don't appear to have an IBM job to hand

/payload-job periodic-ci-openshift-release-master-nightly-4.14-e2e-ibmcloud-csi

JoelSpeed:
With the latest push and https://github.com/openshift/cloud-provider-nutanix/pull/17 I think we have this working across all platforms, will get the nutanix parts pre-merge verified

elmiko:
this is making sense to me. i see the comment about the vsphere tests, happy to add labels when you are ready @JoelSpeed 

the only thing that occurs to me during review is if it is possible to factor out commonalities in the various apply role functions. i'm guessing not, but figured i'd ask.

JoelSpeed:
> the only thing that occurs to me during review is if it is possible to factor out commonalities in the various apply role functions. i'm guessing not, but figured i'd ask.

I don't think so since everything is typed by that point, and the types are incompatible

JoelSpeed:
/hold cancel

This is now tested and QE'd across all platforms and is ready to go

/approve

PR #262, opened 2023-07-12
OpenStack: Remove `use-octavia` from config

dulek:
That option got removed in kubernetes/cloud-provider-openstack#1882. We don't need to set it anymore and this commit makes sure it's gone from the configuration.

### DIFF
--- a/pkg/cloud/openstack/openstack.go
+++ b/pkg/cloud/openstack/openstack.go
@@ -61,14 +61,10 @@ func getTemplateValues(images *imagesReference, operatorConfig config.OperatorCo
 }
 
 // createLoadBalancerSection creates a loadBalancer section populated with
-// the use-octavia key and depending on the network type also the enabled key.
+// the enabled key depending on the network type.
 // It returns any error that happens.
 func createLoadBalancerSection(cfg *ini.File, network *configv1.Network) error {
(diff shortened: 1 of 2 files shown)


JoelSpeed:
Looks like some origin test needs to be updated before this can be merged, complaining of a test failure because this setting isn't present

dulek:
> Looks like some origin test needs to be updated before this can be merged, complaining of a test failure because this setting isn't present

True. openstack-test#115 takes care of that.

dulek:
/retest

I'm pretty sure this compiles, so failures on that stage seems wrong.

dulek:
Tests pass, this is good to go after openshift/openstack-test#115 got merged.

dulek:
> /lgtm
> 
> @dulek I believe we will also need to remove from [docs](https://github.com/search?q=repo%3Aopenshift%2Fopenshift-docs%20use-octavia&type=code) and [CI](https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/openstack/enable-octavia/ipi-conf-openstack-enable-octavia-commands.sh#L30).

https://github.com/openshift/installer/pull/7335
https://github.com/openshift/release/pull/41335
https://github.com/openshift/openshift-docs/pull/62439

PR #264, opened 2023-07-13
OpenStack: Enable manage-security-groups by default

dulek:
`manage-security-groups` enables us to be more granular when allowing LB access to the cluster nodes. When enabled, CPO will create SGs dynamically for each of the LoadBalancer Services and will make sure only required NodePorts are opened on the nodes.

This also enables `loadBalancerSourceRanges` support for LBs created using OVN Octavia Provider.

This commit makes sure the option is enabled by default. Admin can override it by setting it to false manually.

### DIFF
--- a/pkg/cloud/openstack/openstack.go
+++ b/pkg/cloud/openstack/openstack.go
@@ -78,6 +78,12 @@ func createLoadBalancerSection(cfg *ini.File, network *configv1.Network) error {
 	if err != nil {
 		return fmt.Errorf("failed to disable shared LBs: %w", err)
 	}
+
+	// Enable manage-security-groups by default to expose NodePorts dynamically.
+	_, err = loadBalancer.NewKey("manage-security-groups", "true")
+	if err != nil {
+		return fmt.Errorf("failed to enable managing LB members SGs: %w", err)
+	}
 	return nil
 }
 
@@ -97,6 +103,7 @@ func updateLoadBalancerSection(loadBalancer *ini.Section, network *configv1.Netw
 			enabledKey.SetValue("false")
 		}
 	}
+
 	// Disable shared LBs by default if not overriden already
 	_, err := loadBalancer.GetKey("max-shared-lb")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/cloud/openstack/openstack.go
```diff
@@ -104,6 +108,15 @@ func updateLoadBalancerSection(loadBalancer *ini.Section, network *configv1.Netw
 			enabledKey.SetValue("false")
 		}
 	}
+	// Enable manage-security-groups by default if not overriden already
+	_, err = loadBalancer.GetKey("manage-security-groups")
```
JoelSpeed, line 112: Is the intention that, if it's already set, we don't care what it is, so won't overwrite? We don't enforce this to be true?

dulek: This is indeed my intention to allow overriding it manually. I can think of some use cases when it makes sense:

* User doesn't care, then having the full range of NodePorts opened on master/worker SG puts less stress on Neutron.
* User has a Neutron plugin that is bad at SGs (happened in the past) and prefers to limit the number of them.
* Opening an SG takes some time (for CPO to create it, for Neutron to actually enforce it) and that's unacceptable for user's workload.

I intend to create an extensive release note to go with this PR along with instructions how to override the new behavior if need be.

dulek:
/retest

I'm going to try again, I see oomkilled in the unit tests job logs.

dulek:
/hold cancel 
/retest

https://github.com/openshift/installer/pull/7378 is in.

dulek:
4h to run unit tests, WTH?
/test unit

dulek:
This looks good now!

/cc @MaysaMacedo @mandre

MaysaMacedo:
@dulek 

The PR description mentions:

> This also enables loadBalancerSourceRanges support for LBs created using OVN Octavia Provider.

does this comes with the `manage-security-groups` setting?

elmiko:
sorry @dulek , we are having some resource issue with our ci configuration. i've update our ci config, but we are still seeing a few intermittent failures. glad to see this passing though.

dulek:
> @dulek
> 
> The PR description mentions:
> 
> > This also enables loadBalancerSourceRanges support for LBs created using OVN Octavia Provider.
> 
> does this comes with the `manage-security-groups` setting?

Yes, see https://github.com/kubernetes/cloud-provider-openstack/pull/2291 and https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/openstack/loadbalancer.go#L2246-L2252.

PR #266, opened 2023-07-21
OpenStack: Extend owners

dulek:
Seems like we were missing some team members from approvers. This commit adds the whole team.


damdo (commented):
/lgtm

PR #267, opened 2023-07-24
OCPBUGS-16690: Allow all NoSchedule taints for Azure CNM DaemonSet

JoelSpeed:
Customers have the ability to taint nodes via MachineSets, when they do so, they can prevent the cloud-node-manager from being created on worker nodes. If they do this, then the providerID and uninitialized taint are not set and not removed. This means that pods cannot schedule on the nodes and they become wedged.

To enable wider usage of the taints feature via MachineSets, this PR moves the toleration to allow any NoSchedule taint. This should be sufficient for most use cases. I've deliberately not included NoExecute taints since I think it is valid to not run on a NoExecute for the node manager.

### DIFF
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/common"
+	. "github.com/onsi/gomega"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/stretchr/testify/assert"
@@ -14,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/common"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/config"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/util/testingutils"
 )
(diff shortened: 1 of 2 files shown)


pkg/cloud/cloud_test.go
```diff
@@ -377,10 +378,21 @@ func checkResourceTolerations(t *testing.T, podSpec corev1.PodSpec) {
...
+		Effect:   corev1.TaintEffectNoSchedule,
+	}
 
 	tolerations := podSpec.Tolerations
-	assert.Contains(t, tolerations, uninitializedTaint, "PodSpec should tolerate the uninitialized taint")
-	assert.Contains(t, tolerations, notReadyTaint, "PodSpec should tolerate the not-ready taint")
+
+	g := NewWithT(t)
+	g.Expect(tolerations).To(SatisfyAny(
+		SatisfyAll(
```
elmiko, line 390: i'm just curious about the logic here, why the nested `SatisfyAll` inside the `SatisfyAny`? (i would have expected a `SatisfyAll` with all the tolerations)

elmiko (commented):
the reasoning for this makes sense and i tend to agree about the `NoExecute` taint, i just have a question:

pkg/cloud/cloud_test.go
JoelSpeed: We previously required to have the Uninitialized and Unready taints tolerated, both of them together. The logic here now is saying that you must have `(Uninitiailized && Unready) || All`, or at least, that's what I've tried to do. Note the NoSchedule toleration is outside of the `SatisfyAll`

elmiko: ok, that  makes sense to me. it's slightly confusing to read, but i get what you are saying, thanks!

PR #269, opened 2023-07-31
OCPBUGS-17101: update x/net dependency to 0.7.0

elmiko:
this change updates the transitive dependency for the x/net module to address issues from CVE-2022-41723.


elmiko:
looks like the unit test is timing out now, i'm not sure what is happening. going to do some digging.

elmiko:
@JoelSpeed for some reason the approved label didn't take

PR #270, opened 2023-08-01
WIP DO NOT MERGE

elmiko:
testing unit stuff


elmiko:
looking at the release config for this job i see we are specifying `10m` for cpu requests and i'm starting to wonder if that's too low for our envtest suite.

elmiko:
unit test seems like it's working on 4.13 now

PR #274, opened 2023-08-10
OCPCLOUD-2013: Move Azure Credentials Request to custom role

JoelSpeed:
This PR moves the Azure credentials request to use fine grained permissions rather than a generic role.

i've manually tested this and can bring up an Azure workload identity cluster with this manually. CI is covering our existing cluster bring up.


abutcher:
Ah, merge conflict. The CR now has `serviceAccountNames` which will be required in rebase for the test to work.

PR #276, opened 2023-08-31
OCPBUGS-17652: apply necessary RBAC for the alibaba cloud controller manager

rvanderp3:
What changed? 
Created additional clusterrole/role and clusterrole/role binding to grant the necessary privileges to the Alibaba CCM.

Why?
A problem was found which prevented the cloud controller manager from initializing nodes and thus preventing a cluster from bootstrapping.
~~~sh
I0818 02:32:45.382509       1 leaderelection.go:248] attempting to acquire leader lease openshift-cloud-controller-manager/ccm...
E0818 02:32:45.384196       1 leaderelection.go:330] error retrieving resource lock openshift-cloud-controller-manager/ccm: endpoints "ccm" is forbidden: User "system:serviceaccount:openshift-cloud-controller-manager:cloud-controller-manager" cannot get resource "endpoints" in API group "" in the namespace "openshift-cloud-controller-manager" 
~~~

Testing
A new [presubmit](https://github.com/openshift/release/pull/42886) was created which allowed testing of this PR.  The problem was reproduced in `release` and it was verified that the installation completed in this PR.


jianli-wei (approved):
/label qe-approved

JoelSpeed (commented):
Does the alibaba provider use service account credentials for each controller? Other providers do and therefore implicitly get the permissions from other, existing preset RBAC

For those that then need additional permissions on top of that, we are adding them into the assets directly, see vsphere, https://github.com/openshift/cluster-cloud-controller-manager-operator/tree/master/pkg/cloud/vsphere/assets, for example.

Can we make sure to follow the same pattern here

rvanderp3:
> Does the alibaba provider use service account credentials for each controller? Other providers do and therefore implicitly get the permissions from other, existing preset RBAC
> 
I am not sure about whether or not each controller uses service account credentials but the cloud provider was crashlooping soon after it started due to missing privileges for the `cloud-controller-manager` service account. 

> For those that then need additional permissions on top of that, we are adding them into the assets directly, see vsphere, https://github.com/openshift/cluster-cloud-controller-manager-operator/tree/master/pkg/cloud/vsphere/assets, for example.
> 
> Can we make sure to follow the same pattern here
Yep, just made that change.

jianli-wei:
@JoelSpeed @rvanderp3 As pre-merge testing passed, wonder if we could get the PR merged as soon, because the corresponding bug is a test-blocker? Thanks!

JoelSpeed:
> @JoelSpeed it looks like changing the cluster role [might not be possible](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-cloud-controller-manager-operator/276/pull-ci-openshift-cluster-cloud-controller-manager-operator-master-e2e-alibaba-ovn/1697615647028023296#1:build-log.txt%3A174-186) from the assets directory. Would you mind taking a look to see if I may have missed something?

Any new permissions you are adding must also be added to the CCMO clusterrole so that it isn't attempting privilege escalation. We did this for [vsphere](https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/786dda44224610b4dd78f2b953a90686e90b68fd/manifests/0000_26_cloud-controller-manager-operator_02_rbac_operator.yaml#L76-L105)

rvanderp3:
> @JoelSpeed @rvanderp3 As pre-merge testing passed, wonder if we could get the PR merged as soon, because the corresponding bug is a test-blocker? Thanks!

there is still some work to do here and I'm taking a look at it today

pkg/cloud/alibaba/assets/cloud-controller-manager-clusterrole.yaml
```diff
@@ -0,0 +1,66 @@
...
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumes
+      - services
+      - secrets
```
JoelSpeed, line 14: We recently removed the need for other CCMs to use a clusterrole with overly permissive secrets, is this definitely required by the alibaba CCM or could this use a role scoped to the required namespaces?

pkg/cloud/alibaba/assets/cloud-controller-manager-clusterrole.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cloud-controller-manager
```
JoelSpeed, line 4: We already have a cloud-controller-manager role don't we? You'll need to give this a different name and create a clusterrolebinding as well

JoelSpeed (commented):
Is the Alibaba CCM actually using service account credentials for each controller as intended? If so, a lot of the permissions it's being granted here should already be there. The base CCM should have permissions for leases etc already

I would expect an additional clusterrole and binding to contain only those that are needed in addition to the basic set

pkg/cloud/alibaba/assets/cloud-controller-manager-clusterrole.yaml
rvanderp3: i'll investigate this

pkg/cloud/alibaba/assets/cloud-controller-manager-clusterrole.yaml
rvanderp3: we do, it didn't occur to me there was a name collision here.  I'll work on this.

jianli-wei:
@rvanderp3 @JoelSpeed Any updates? The bug led to all the 4.14 Prow CI jobs failing, hopefully we could get this merged as soon. Please advise, thanks in advance!

rvanderp3:
@jianli-wei I've not had a chance to incorporate Joel's feedback yet.  I'll try to get back to this this week.

jianli-wei:
> @jianli-wei I've not had a chance to incorporate Joel's feedback yet. I'll try to get back to this this week.

@rvanderp3 Any updates? Thanks!

rvanderp3:
> > @jianli-wei I've not had a chance to incorporate Joel's feedback yet. I'll try to get back to this this week.
> 
> @rvanderp3 Any updates? Thanks!

Not yet unfortunately, I'll work on this as soon as I can.

gujingit:
@rvanderp3 The version v1.9.3.376-g5c84e19-aliyun-379-g8c532d2e6  is out-of-date. You can use [v2.7.0](https://github.com/kubernetes/cloud-provider-alibaba-cloud/tags) or build a new version from the master branch, which can run with the openshift default ccm role.

gujingit:
@JoelSpeed Alibaba CCM use the default openshift cloud-controller-manager role, and the role has bee pruned in the commit #b895144aa6e5b94cefaa5b95a0fb2ad7ac0b49f8. The clusterrole policy alibaba ccm needs is as blew

rvanderp3:
Thanks @gujingit! i'll take a look at this.

rvanderp3:
converted to draft while changes are being tested to prevent CI churn

rvanderp3:
made progress today, will continue work tomorrow.

rvanderp3:
@jianli-wei would you mind performing pre-submit testing on this?

jianli-wei:
> @jianli-wei would you mind performing pre-submit testing on this?

Sure, let me retry today. Thanks!

jianli-wei:
@rvanderp3 Pre-merge testing with the latest commits passed, see bug [comment](https://issues.redhat.com/browse/OCPBUGS-17652?focusedId=23258916&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23258916). Do you have any other concern?  And is it possible to get the PR merged as soon, and back-port to 4.14 soon? Thanks!

rvanderp3:
Failures in e2e-alibaba-ovn do not appear to be related to this PR.  I reviewed the job with others on the SPLAT team and it appears the tests are failing due to the amount of time it takes to reconcile the load balancer during the e2e tests.

pkg/cloud/alibaba/assets/alibaba-cloud-controller-manager-rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: alibaba-cloud-controller-manager:alibaba-cloud-controller-manager
+  namespace: openshift-cloud-controller-manager
+roleRef:
+  kind: Role
+  name: alibaba-cloud-controller-manager
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    namespace: openshift-cloud-controller-manager
+    name: cloud-controller-manager
```
JoelSpeed, line 13: Nit missing new line at end of file

JoelSpeed (commented):
Few nits, but otherwise if it's working, then I'm happy to merge it

rvanderp3:
> Few nits, but otherwise if it's working, then I'm happy to merge it

Thanks Joel, I just made the updates and it should be ready to go.

JoelSpeed:
Are we confident that the pre-merge testing here is sufficient and that the E2E failing is acceptable?

rvanderp3:
> Are we confident that the pre-merge testing here is sufficient and that the E2E failing is acceptable?

I believe so.  I'll file a bug to follow up on the test failure.  We've not been able to even install for months.

rvanderp3:
> > Are we confident that the pre-merge testing here is sufficient and that the E2E failing is acceptable?
> 
> I believe so. I'll file a bug to follow up on the test failure. We've not been able to even install for quite some time now.

I hit comment a little too soon .. I meant to follow that up with that it appears that it is taking an extended period of time for the load balancer to reconcile.  I'll have a bug on that shortly.

PR #279, opened 2023-09-18
OCPBUGS-18641: Set dual-stack IPFamilyPriority for vSphere

mkowalski:
We have discovered that in dual-stack setups NodeAddresses field of
the instance metadata contains only IPv4 addresses for VMs that do have
both IPv4 and IPv6 addresses assigned (and detected by the VM agent).

It has been traced back to the function responsible for populating this
metadata field. We found out that for our configuration we always filter
only IPv4 addresses even if running in dual-stack. Reason for that is
that `IPFamilyPriority` has a value of `ipv4` even when running in a
dual-stack setup.

This causes an issue because this instance metadata is cross-checked
with addresses provided by the kubelet as part of the
`alpha.kubernetes.io/provided-node-ip` annotation. Without correct value
of `IPFamilyPriority` we are thus removing all the IPv6 addresses.

This PR takes advantage of the Service Networks configured by the user in
install-config and the fact that we only allow 2 networks to be configured
if the setup is dual-stack.

Fixes: OCPBUGS-18641

### DIFF
--- a/pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
+++ b/pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
@@ -66,6 +66,8 @@ spec:
               value: {{ .globalCredsSecretNamespace }}
             - name: VSPHERE_SECRET_NAME
               value: {{ .globalCredsSecretName }}
+            - name: ENABLE_ALPHA_DUAL_STACK
+              value: "true"
           resources:
             requests:
               cpu: 200m
--- a/pkg/cloud/vsphere/vsphere_config_transformer.go
+++ b/pkg/cloud/vsphere/vsphere_config_transformer.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"k8s.io/utils/net"
 
 	ccmConfig "github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/vsphere/vsphere_cloud_config"
 )
@@ -26,7 +27,7 @@ const (
 // Currently, CloudConfigTransformer is responsible to populate vcenters, labels, and node networking parameters from
 // the Infrastructure resource.
 // Also, this function converts legacy deprecated INI configuration format to a YAML-based one.
-func CloudConfigTransformer(source string, infra *configv1.Infrastructure, _ *configv1.Network) (string, error) {
+func CloudConfigTransformer(source string, infra *configv1.Infrastructure, network *configv1.Network) (string, error) {
 	if infra.Status.PlatformStatus == nil ||
 		infra.Status.PlatformStatus.Type != configv1.VSpherePlatformType {
 		return "", fmt.Errorf("invalid platform, expected to be %s", configv1.VSpherePlatformType)
@@ -42,6 +43,7 @@ func CloudConfigTransformer(source string, infra *configv1.Infrastructure, _ *co
 	// https://github.com/openshift/enhancements/blob/f6b33eb0cd4ba060af71fee6192297cf6bc31e5a/enhancements/installer/vsphere-ipi-zonal.md
 	// https://github.com/openshift/api/pull/1278
 	if infra.Spec.PlatformSpec.VSphere != nil {
+		setDualStack(cpiCfg, infra.Status.PlatformStatus.VSphere, &infra.Spec.PlatformSpec.VSphere.NodeNetworking, network)
 		setNodes(cpiCfg, &infra.Spec.PlatformSpec.VSphere.NodeNetworking)
 		setVirtualCenters(cpiCfg, infra.Spec.PlatformSpec.VSphere)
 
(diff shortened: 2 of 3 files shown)


mkowalski:
/test e2e-vsphere-ovn

It's sig-storage that fails, let's just retest

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
```diff
@@ -66,6 +66,8 @@ spec:
               value: {{ .globalCredsSecretNamespace }}
             - name: VSPHERE_SECRET_NAME
               value: {{ .globalCredsSecretName }}
+            - name: ENABLE_ALPHA_DUAL_STACK
+              value: "true"
```
JoelSpeed, line 70: Is this forcing on an alpha feature gate for all vSphere clusters?

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -99,3 +101,41 @@ func setVirtualCenters(cfg *ccmConfig.CPIConfig, vSphereSpec *configv1.VSpherePl
...
+// (i.e. API VIP and Ingress VIP for IPI installations and fd69::2 which is internal to OVN-K8s).
+//
+// Ref.: https://issues.redhat.com/browse/OCPBUGS-18641
+func setDualStack(cfg *ccmConfig.CPIConfig, status *configv1.VSpherePlatformStatus, nodeNetworking *configv1.VSpherePlatformNodeNetworking, network *configv1.Network) {
+	if network != nil && len(network.Spec.ServiceNetwork) == 2 {
+		if net.IsIPv4CIDRString(network.Spec.ServiceNetwork[0]) {
+			cfg.Global.IPFamilyPriority = []string{"ipv4", "ipv6"}
+		} else {
+			cfg.Global.IPFamilyPriority = []string{"ipv6", "ipv4"}
+		}
```
JoelSpeed, line 124: Is there a possibility that both entries in the list are ipv4?

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
mkowalski: This is not really a feature gate, at least not in the sense how we implemented `CloudDualStackNodeIPs` in https://kubernetes.io/docs/concepts/services-networking/dual-stack/

This is some VMware internal(?) code that requires an existence of env variable (without even looking at its value), look at https://github.com/openshift/cloud-provider-vsphere/blob/master/pkg/cloudprovider/vsphere/cloud.go#L253-L255 to see how it's handled. Not really the way how every other component handles `CloudDualStackNodeIPs` (even the name does not match).

From the moment I've seen it I got confused about how we should handle this. The only thing that made me feel safe is that this validator I have linked is the only place in the whole codebase where this is used.

Any suggestion?

pkg/cloud/vsphere/vsphere_config_transformer.go
mkowalski: No, having `["ipv4,"ipv4"]` would behave exactly the same as `["ipv4"]`

JoelSpeed: My point was more, `ServiceNetwork` could be two IPv4 addresses right? Why aren't we checking both entries?

pkg/cloud/vsphere/assets/cloud-controller-manager-deployment.yaml
JoelSpeed: Looking at the code, it seems relatively minor, I guess we can leave it as is

pkg/cloud/vsphere/vsphere_config_transformer.go
mkowalski: No, that's not allowed. If you have 2 service networks, then one has to be v4 and the other one v6

mkowalski: In fact well, let me confirm this with o/installer... It may have changed and the validations are way more complex now

mkowalski: https://github.com/openshift/installer/blob/master/pkg/types/validation/installconfig.go#L310-L313 <-- so if you are installing single-stack v4 cluster, you may have only 1 service network entry

JoelSpeed: Great, can you just add a comment inline to explain that if there are two service networks, the installer already validated that one is ipv4 and the other is ipv6, so we don't lose this context?

mkowalski: Of course, done

JoelSpeed (commented):
/approve
/lgtm

JoelSpeed:
/hold
/approve cancel

CI failures indicate that the CSI driver is probably not ok with this change, @mkowalski can you please take a look

mkowalski:
As discussed, seems like wider failure what makes me think it's not my code what introduced it (luckily this time)

Ref.: https://redhat-internal.slack.com/archives/C01CQA76KMX/p1695228004392599

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -99,3 +101,47 @@ func setVirtualCenters(cfg *ccmConfig.CPIConfig, vSphereSpec *configv1.VSpherePl
...
+					addr = addr + "/32"
+				} else {
+					addr = addr + "/128"
+				}
+				nodeNetworking.External.ExcludeNetworkSubnetCIDR = append(nodeNetworking.External.ExcludeNetworkSubnetCIDR, addr)
+				nodeNetworking.Internal.ExcludeNetworkSubnetCIDR = append(nodeNetworking.Internal.ExcludeNetworkSubnetCIDR, addr)
+			}
+		}
+
+		nodeNetworking.External.ExcludeNetworkSubnetCIDR = append(nodeNetworking.External.ExcludeNetworkSubnetCIDR, "fd69::2/128")
```
rbbratta, line 144: maybe a named constant for `fd69::2/128` so we can crossref with ovn-kubernetes.

mkowalski: I'd totally agree if we had a common/shared library that ovn-k8s would consume. In the state like it is now, I'd say address is an address and it's not a big diff if there is variable or not (especially that I use it only in a single place)

rbbratta: we avoid magic numbers?  and it is used twice.   but maybe we add a comment in OVN-k8s as well.

https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L145

can customer override `V6HostMasqueradeIP`?

rbbratta: Yep, it's configurable  https://github.com/ovn-org/ovn-kubernetes/pull/3594

rbbratta: https://github.com/openshift/cluster-network-operator/pull/1807

rbbratta: Err, we can pull from the API?  https://github.com/openshift/api/pull/1410

mkowalski: I'm looking into this. It's not nice because they didn't add it as Network CRD but only in the Operator Config CRD so I'm not sure if I can access it easily from inside the 3cmo

mkowalski: To clarify - until https://github.com/openshift/cluster-network-operator/pull/1807 merges, this value is **not configurable**. I am now discussing with OVN-K8s folks the rationale, but at this moment this value is the only one that we will see in the wild.

Having said this, I wonder if we can push this change as it is now and then when OVN-K8s folks are ready with https://github.com/openshift/cluster-network-operator/pull/1807 (hopefully with revisited API that stores the custom subnet in the Network CR) improve it for 4.15

rbbratta: Yep, I agree we should merge this to unblock dual-stack.

mkowalski:
https://github.com/openshift/kubernetes/pull/1714 is already in the merge pool. Hopefully we get a nightly today and by tomorrow this one here is green

rbbratta:
`launch 4.14,openshift/cluster-cloud-controller-manager-operator#279 vsphere,dualstack`

failed, due to lack of nodes

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-vsphere-modern/1706620554640887808

```
message="Error creating: pods \"cluster-version-operator-7d7d9c9b99-\" is forbidden: autoscaling.openshift.io/ManagementCPUsOverride the cluster does not have any nodes"
```

retrying.

mkowalski:
We had a successful cluster from this PR back in September 19th, https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-vsphere-modern/1704172710533271552

rbbratta:
I had successful cluster yesterday, so maybe quota or something.

PR #287, opened 2023-10-11
OCPBUGS-18841: Additional permissions for internal load balancer on STS

JoelSpeed:
Managed to reproduce the issue QE were seeing by using an STS cluster, adding these two new permissions resolved the problem and the internal service got the IP it needed


damdo (commented):
/lgtm

PR #291, opened 2023-10-26
OCPBUGS-20213: Add azure cloud config transformer

elmiko:
This change creates a CloudConfigTransformer for Azure Stack cloud
provider. The transformer will unmarshal json from the cloud config into
a `Config` struct from the azure cloud provider and then update the the
`VMType` field if it is unset. This is being added because Azure has
updated the default vm type to value that cannot be added to the basic
load balancers which are associated with Azure Stack.

The change also adds a test suite to accompany the transformer.

I've added a dependency on the upstream cloud-provider-azure to get the Config object so that we can do a proper marshal/unmarshal operation in the transformer. This seemed necessary because I observed some odd type conversions that were happening when trying to marshal/unmarshal using `map[string]interface{}` types.


pkg/cloud/azurestack/azurestack.go
```diff
@@ -2,17 +2,21 @@ package azurestack
...
 	appsv1 "k8s.io/api/apps/v1"
+	azure "sigs.k8s.io/cloud-provider-azure/pkg/provider"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/common"
 	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/config"
 )
 
 const providerName = "azurestack"
+const vmTypeStandard = "standard"
```
JoelSpeed, line 19: Not exported in the Azure API somewhere?

pkg/cloud/azurestack/azurestack.go
```diff
@@ -86,3 +90,38 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+	// see OCPBUGS-20213 for more information.
+	if cfg.VMType == "" {
+		cfg.VMType = vmTypeStandard
+	}
+
+	if cfgbytes, err := json.Marshal(cfg); err != nil {
+		return "", fmt.Errorf("failed to marshal the cloud.conf: %w", err)
+	} else {
+		return string(cfgbytes), nil
+	}
```
JoelSpeed, line 126: Nit, I'm sure there's a linter that would suggest removing this else 🤔 
```suggestion
	}
	
	return string(cfgbytes), nil
```

pkg/cloud/azurestack/azurestack_test.go
```diff
@@ -62,3 +72,89 @@ func TestResourcesRenderingSmoke(t *testing.T) {
...
+			g.Expect(err).NotTo(HaveOccurred(), "Marshal of source data should succeed")
+
+			actual, err := CloudConfigTransformer(string(src), tc.infra, nil)
+			if tc.errMsg != "" {
+				g.Expect(err).Should(MatchError(tc.errMsg))
+				g.Expect(actual).Should(Equal(""))
+			} else {
+				var observed azure.Config
+				err = json.Unmarshal([]byte(actual), &observed)
+				g.Expect(err).NotTo(HaveOccurred(), "Unmarshal of observed data should succeed")
```
JoelSpeed, line 154: Nit, could inline this
```suggestion
				g.Expect(json.Unmarshal([]byte(actual), &observed)).To(Succeed(), "Unmarshal of observed data should succeed")
```

pkg/cloud/azurestack/azurestack.go
elmiko: but then i also need to take `cfgbytes` out of the `if` clause assignement right?

pkg/cloud/azurestack/azurestack.go
elmiko: i couldn't find one, sadly

pkg/cloud/azurestack/azurestack.go
JoelSpeed: Yeah you would need to, I think that's more idiomatic from my understanding

pkg/cloud/azurestack/azurestack.go
elmiko: with fresh eyes, i have found it

elmiko:
updated to address nits

PR #292, opened 2023-11-03
OSASINFRA-3284: Add Events to OpenStack's RBAC

dulek:
OpenStack keeps a separate SA definition for its CCM with a permission to patch Services. kubernetes/cloud-provider-openstack#2383 extends the needs to also be able to create events and this patch solves that.


elmiko (commented):
/approve
/lgtm

PR #294, opened 2023-11-14
OSASINFRA-3289: OpenStack: Remove Kuryr bits

dulek:
Kuryr is being removed from 4.15, this commit removes the Kuryr bits from the OpenStack's CCM config transformer.


mandre (commented):
/lgtm
The test failure is unrelated.

PR #299, opened 2023-11-15
OCPBUGS-23308: Set IPFamilyPriority and ExcludeNetworkSubnetCIDR for vSphere IPv6-only

mkowalski:
In 428d639f0272c17b33058b994bd43104525d0205 we have added a support for correctly setting `IPFamilyPriority` and `ExcludeNetworkSubnetCIDR` when using dual-stack configuration. That commit however missed the configuration for IPv6-only setups and as a result those clusters are running with `IPv4` as a value of `IPFamilyPriority`.

This PR changes the logic so that we now cover settings for both IPv6-only and dual-stack configurations. There is no need for explicit handling of IPv4-only clusters as those work correctly today.

Fixes: OCPBUGS-23308
Relates-to: OCPBUGS-18641


mkowalski:
e2e-vsphere-ovn passed so there is no regression in IPv4 world. Other failed platforms do not use the modified code so their failures are not relevant

PR #300, opened 2023-11-21
OCPBUGS-23996: trust bundle CA controller: add owner annotation for ccm-trusted-ca configmap

vrutkovs:
This adds annotation which identifies the component used to file issues with this certificate. The annotation is verified by [the test](https://github.com/openshift/origin/blob/master/test/extended/operators/certs.go#L130), so that all certificates would be registered and have owners assigned. More metadata would be added later.

See https://github.com/openshift/enhancements/pull/1502 for proposed workflow with TLS registry


JoelSpeed:
@vrutkovs Can you add a PR description that explains why this is happening. I have no idea what this annotation is

vrutkovs:
Done

PR #310, opened 2023-12-18
OCPBUGS-5825: Adds GCP ClusterRole and ClusterRoleBinding

theobarberbany:
This change adds back the ClusterRole and ClusterRoleBinding removed in this PR [OCPBUGS-5825](https://github.com/openshift/cluster-kube-controller-manager-operator/pull/778).

This is because it is still required at present.

### DIFF
--- a/pkg/cloud/gcp/assets/gcp-cloud-controller-manager-clusterrolebinding.yaml
+++ b/pkg/cloud/gcp/assets/gcp-cloud-controller-manager-clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: gcp-cloud-controller-manager:cloud-provider
+roleRef:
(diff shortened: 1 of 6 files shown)


pkg/cloud/gcp/assets/gcp-cloud-controller-manager-clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: gcp-cloud-controller-manager:gcp-cloud-controller-manager
```
JoelSpeed, line 4: I think this is typically named after the service account, so maybe this should be
```suggestion
  name: gcp-cloud-controller-manager:cloud-provider
```

JoelSpeed:
/override ci/prow/e2e-aws-ovn

I'm confident that this change isn't going to affect AWS, but, the change is blocking GCP CI from going green, so I think in this case it's desirable to override

PR #316, opened 2023-12-20
OCPBUGS-25483: Adds CloudConfigTransformer for Azure

theobarberbany:
This adds a CloudConfigTransformer for Azure that will set the VMType to 'standard' if unset.

It also adds the functionality from the cluster-config-operator azure CloudConfigTransformer, as we can no longer rely on it by introducing this transformer.

See #291, OCPBUGS-25483 and OCPBUGS-20213 for more information.

### DIFF
--- a/pkg/cloud/azure/azure.go
+++ b/pkg/cloud/azure/azure.go
@@ -4,10 +4,12 @@ import (
 	"embed"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/asaskevich/govalidator"
 	configv1 "github.com/openshift/api/config/v1"
 	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	azureconsts "sigs.k8s.io/cloud-provider-azure/pkg/consts"
@@ -26,6 +28,24 @@ var (
 		{ReferenceObject: &appsv1.Deployment{}, EmbedFsPath: "assets/cloud-controller-manager-deployment.yaml"},
 		{ReferenceObject: &appsv1.DaemonSet{}, EmbedFsPath: "assets/cloud-node-manager-daemonset.yaml"},
 	}
+)
+
+var (
+	validAzureCloudNames = map[configv1.AzureCloudEnvironment]bool{
+		configv1.AzurePublicCloud:       true,
+		configv1.AzureUSGovernmentCloud: true,
+		configv1.AzureChinaCloud:        true,
+		configv1.AzureGermanCloud:       true,
+		configv1.AzureStackCloud:        true,
+	}
+
+	validAzureCloudNameValues = func() []string {
+		v := make([]string, 0, len(validAzureCloudNames))
+		for n := range validAzureCloudNames {
+			v = append(v, string(n))
+		}
+		return v
+	}()
 )
 
 type imagesReference struct {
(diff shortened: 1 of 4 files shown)


pkg/cloud/cloud.go
```diff
@@ -53,7 +53,8 @@ func GetCloudConfigTransformer(platformStatus *configv1.PlatformStatus) (cloudCo
 		if azurestack.IsAzureStackHub(platformStatus) {
 			return azurestack.CloudConfigTransformer, true, nil
 		}
-		return nil, true, nil
+		// We need to use the azure CloudConfigTransformer to fix OCPBUGS-25483
```
JoelSpeed, line 56: This comment is unlikely to age well, and git blame is probably going to be easier for us to use to track the history of this.

theobarberbany: Right - so ok to just remove it?

JoelSpeed: Yeah, I expect in the future this transformer will handle fixes for multiple things, not just this one bug, so it will probably be wrong pretty quickly

pkg/cloud/azure/azure.go
```diff
@@ -23,6 +30,24 @@ var (
...
 )
 
+var (
+	validAzureCloudNames = map[configv1.AzureCloudEnvironment]bool{
+		configv1.AzurePublicCloud:       true,
+		configv1.AzureUSGovernmentCloud: true,
+		configv1.AzureChinaCloud:        true,
+		configv1.AzureGermanCloud:       true,
+		configv1.AzureStackCloud:        true,
+	}
```
JoelSpeed, line 40: I'm guessing this was copy/pasted? Nit, `map[T]struct{}` is more efficient than `map[T]bool`

pkg/cloud/azure/azure.go
```diff
@@ -85,3 +110,71 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
 	}
 	return assets, nil
 }
+
+func IsAzure(infra *configv1.Infrastructure) bool {
+	if infra.Status.PlatformStatus != nil {
+		if infra.Status.PlatformStatus.Type == configv1.AzurePlatformType &&
+			(infra.Status.PlatformStatus.Azure.CloudName != configv1.AzureStackCloud) {
```
JoelSpeed, line 117: Why do we exclude ASH from this? Might want a comment to explain that ASH is handled separately in its own transformer

pkg/cloud/azure/azure.go
```diff
@@ -85,3 +110,71 @@ func NewProviderAssets(config config.OperatorConfig) (common.CloudProviderAssets
...
+			return "",
+				fmt.Errorf(`invalid user-provided cloud.conf: \"cloud\" field in user-provided
+				cloud.conf conflicts with infrastructure object`)
+		}
+	}
+
+	cfg.Cloud = string(cloud)
+
+	// TODO: Remove when you work this out (before merging)
+	// Why is cfg.Cloud not typed: type AzureCloudEnvironment string
```
JoelSpeed, line 166: Is the cfg.Cloud an upstream struct definition or something we have locally? If it's an upstream definition, I guess there isn't much we can do

pkg/cloud/azure/azure_test.go
```diff
@@ -79,3 +91,173 @@ func TestResourcesRenderingSmoke(t *testing.T) {
...
+		// The matcher should assert err.Error == errMsg - but is not for some
+		// reason? According to:
+		// https://onsi.github.io/gomega/#matcherrorexpected-interface the matcher
+		// asserts that ACTUAL.Error() == EXPECTED
+		//	{
+		//		name:   "Azure throws an error if the infra has an invalid cloud",
+		//		source: azure.Config{},
+		//		infra:  makeInfrastructureResource(configv1.AzurePlatformType, "AzureAnotherCloud"),
+		//		errMsg: "status.platformStatus.azure.cloudName: Unsupported value: \"AzureAnotherCloud\": supported values: \"AzurePublicCloud\", \"AzureUSGovernmentCloud\", \"AzureChinaCloud\", \"AzureGermanCloud\", \"AzureStackCloud\""
+		//	},
```
JoelSpeed, line 210: Was this copied or something you've written? Can you provide the failure output for when this is not commented out?

pkg/cloud/azure/azure_test.go
```diff
@@ -79,3 +91,173 @@ func TestResourcesRenderingSmoke(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			src, err := json.Marshal(tc.source)
+			g.Expect(err).NotTo(HaveOccurred(), "Marshal of source data should succeed")
+
+			actual, err := CloudConfigTransformer(string(src), tc.infra, nil)
+			if tc.errMsg != "" {
+				// TODO: Remove once the above error case is working correctly
+				fmt.Printf("\n\n err.Error():%s:\n\n", err.Error())
```
JoelSpeed, line 253: Reminder to remove before merge

pkg/cloud/azure/azure.go
theobarberbany: Yep, it is copy/pasted from the existing functionality in the CCO. 

I didn't know that! Out of interest, what makes `map[T]struct{}` faster? Is it significantly faster for small sets, or only large ones?

JoelSpeed: Memory allocation, `struct{}` allocates less memory than `bool`, so the idiom in Go is to use `struct{}` when you need a set semantic

pkg/cloud/azure/azure.go
theobarberbany: Ah it's upstream. Cheers! :D

pkg/cloud/azure/azure_test.go
theobarberbany: It's something I've written. 

I think it is because we return a `field.NotSupported(` error. 

However, this should be matchable with the contents of it's `err.Error()`, hence my confusion.

EDIT: uncommenting and running `make test` locally this seems to be behaving as expected now 😵‍💫

theobarberbany: Wait no it's erroring again 😖 

```
=== RUN   TestCloudConfigTransformer/Azure_throws_an_error_if_the_infra_has_an_invalid_cloud


 err.Error():status.platformStatus.azure.cloudName: Unsupported value: "AzureAnotherCloud": supported values: "AzureUSGovernmentCloud", "AzureChinaCloud", "AzureGermanCloud", "AzureStackCloud", "AzurePublicCloud":

    azure_test.go:254:
        Expected
            <*field.Error | 0xc000069880>:
            status.platformStatus.azure.cloudName: Unsupported value: "AzureAnotherCloud": supported values: "AzureUSGovernmentCloud", "AzureChinaCloud", "AzureGermanCloud", "AzureStackCloud", "AzurePublicCloud"
            {
                Type: "FieldValueNotSupported",
                Field: "status.platformStatus.azure.cloudName",
                BadValue: <v1.AzureCloudEnvironment>"AzureAnotherCloud",
                Detail: "supported values: \"AzureUSGovernmentCloud\", \"AzureChinaCloud\", \"AzureGermanCloud\", \"AzureStackCloud\", \"AzurePublicCloud\"",
            }
        to match error
            <string>: status.platformStatus.azure.cloudName: Unsupported value: "AzureAnotherCloud": supported values: "AzurePublicCloud", "AzureUSGovernmentCloud", "AzureChinaCloud", "AzureGermanCloud", "AzureStackCloud"
```

PR #322, opened 2024-01-10
OCPBUGS-26494: PowerVS: handle 4.16 changes

hamzy:
1) Change the --bind-address argument to 127.0.0.1
2) Remove updating the pod IP address

### DIFF
--- a/pkg/cloud/powervs/assets/deployment.yaml
+++ b/pkg/cloud/powervs/assets/deployment.yaml
@@ -59,10 +59,6 @@ spec:
         image: {{ .images.CloudControllerManager }}
         imagePullPolicy: IfNotPresent
         env:
-          - name: POD_IP_ADDRESS
-            valueFrom:
-              fieldRef:
-                fieldPath: status.podIP
           - name: VPCCTL_CLOUD_CONFIG
             value: /etc/ibm/cloud.conf
           - name: ENABLE_VPC_PUBLIC_ENDPOINT
@@ -77,7 +73,7 @@ spec:
               source /etc/kubernetes/apiserver-url.env
             fi
             exec /bin/ibm-cloud-controller-manager \
-            --bind-address=$(POD_IP_ADDRESS) \
+            --bind-address=127.0.0.1 \
             --use-service-account-credentials=true \
             --configure-cloud-routes=false \
             --cloud-provider=ibm \


mkumatag (approved):
/lgtm

pkg/cloud/powervs/assets/deployment.yaml
```diff
@@ -77,7 +73,7 @@ spec:
               source /etc/kubernetes/apiserver-url.env
             fi
             exec /bin/ibm-cloud-controller-manager \
-            --bind-address=$(POD_IP_ADDRESS) \
+            --bind-address=$(127.0.0.1) \
```
JoelSpeed, line 76: This is a literal not a variable
```suggestion
            --bind-address=127.0.0.1 \
```

hamzy: stupid mistake!

JoelSpeed:
Change looks ok but I'd like to see an optional E2E job on the repo that proves this works

hamzy:
> Change looks ok but I'd like to see an optional E2E job on the repo that proves this works

I have hand edited a deploying cluster on the original changes and they worked for me.

hamzy:
I will also add that we should eventually work to get that testing done.  However, the PowerVS CI is not stable enough at the moment to require that testing before this PR merges.

JoelSpeed:
> I will also add that we should eventually work to get that testing done. However, the PowerVS CI is not stable enough at the moment to require that testing before this PR merges.

Not even stable enough to prove that a cluster can at least install? Given we know this prevents installs?

hamzy:
> > I will also add that we should eventually work to get that testing done. However, the PowerVS CI is not stable enough at the moment to require that testing before this PR merges.
> 
> Not even stable enough to prove that a cluster can at least install? Given we know this prevents installs?

Well it is more stable to get past the cluster installation step (I would guess > 80 or 90%).  It is getting past running E2E testing where passing every test case consistently is problematic.

hamzy:
I guess the optional passing requirement would at least not block other arches.

JoelSpeed:
What I'm interested in for the purpose of platforms like PowerVS is being able to check the CCM logs on CI runs and see if the CCM is doing what we expect it to do. Which I expect you can probably do with the current state of PowerVS CI

PR #327, opened 2024-01-26
NO-JIRA:Use expressive Gomega matchers with Eventually

nrb:
The matchers used with `Eventually` checks boiled down to a boolean
comparison, which is often unhelpful when debugging. This change uses
more detailed Gomega matchers in order to aid developers making changes.

This explicitly does _not_ try to consolidated the inconsistent import usages of Gomega within this repository.

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller_test.go
+++ b/pkg/controllers/cloud_config_sync_controller_test.go
@@ -224,12 +224,16 @@ var _ = Describe("Cloud config sync controller", func() {
 		co := &configv1.ClusterOperator{}
 		err := cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, co)
 		if err == nil || !apierrors.IsNotFound(err) {
-			Eventually(func() bool {
-				err := cl.Delete(context.Background(), co)
-				return err == nil || apierrors.IsNotFound(err)
-			}).Should(BeTrue())
+			Eventually(func() error {
+				return cl.Delete(context.Background(), co)
+			}).Should(SatisfyAny(
+				Not(HaveOccurred()),
+				MatchError(apierrors.IsNotFound, "IsNotFound"),
+			))
 		}
-		Eventually(apierrors.IsNotFound(cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, co))).Should(BeTrue())
+		Eventually(func() error {
+			return cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, co)
+		}).Should(MatchError(apierrors.IsNotFound, "IsNotFound"))
 
 		By("Cleanup resources")
 		deleteOptions := &client.DeleteOptions{
@@ -240,9 +244,9 @@ var _ = Describe("Cloud config sync controller", func() {
 		Expect(cl.List(ctx, allCMs)).To(Succeed())
 		for _, cm := range allCMs.Items {
 			Expect(cl.Delete(ctx, cm.DeepCopy(), deleteOptions)).To(Succeed())
-			Eventually(
-				apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{})),
-			).Should(BeTrue())
+			Eventually(func() error {
+				return cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{})
+			}).Should(MatchError(apierrors.IsNotFound, "IsNotFound"))
 		}
 
 		infraCloudConfig = nil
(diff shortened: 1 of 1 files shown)


pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -224,10 +224,12 @@ var _ = Describe("Cloud config sync controller", func() {
...
 		if err == nil || !apierrors.IsNotFound(err) {
-			Eventually(func() bool {
-				err := cl.Delete(context.Background(), co)
-				return err == nil || apierrors.IsNotFound(err)
-			}).Should(BeTrue())
+			Eventually(func() error {
+				return cl.Delete(context.Background(), co)
+			}).Should(SatisfyAny(
+				Not(HaveOccurred()),
+				MatchError(apierrors.IsNotFound, "IsNotFound"),
```
JoelSpeed, line 231: Not seen this used like this before, but cool, checked the docs and it works, nice find

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -241,8 +243,8 @@ var _ = Describe("Cloud config sync controller", func() {
 		for _, cm := range allCMs.Items {
 			Expect(cl.Delete(ctx, cm.DeepCopy(), deleteOptions)).To(Succeed())
 			Eventually(
-				apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{})),
-			).Should(BeTrue())
+				cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{}),
```
JoelSpeed, line 246: This is a gotcha of using eventually statements, what you're doing here is calling `cl.Get` while constructing the `Eventually`, which means `Eventually` gets a single result, which never changes.

The original is also broken, but you have two choices here to fix it:
```suggestion
				func() error { return cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{})},
```
Or use [komega](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest/komega) as we do in CPMS
```suggestion
				komega.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{}),
```
You'll need to either create a new instance of komega or set the global komega client before you can do this.

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -251,16 +253,15 @@ var _ = Describe("Cloud config sync controller", func() {
 		infra := makeInfrastructureResource(configv1.AzurePlatformType)
 		Expect(cl.Delete(ctx, infra)).To(Succeed())
 		Eventually(
-			apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(infra), infra)),
-		).Should(BeTrue())
+			cl.Get(ctx, client.ObjectKeyFromObject(infra), infra),
```
JoelSpeed, line 256: Same gotcha as above

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -251,16 +253,15 @@ var _ = Describe("Cloud config sync controller", func() {
...
 
 		networkResource := makeNetworkResource()
 		Expect(cl.Delete(ctx, networkResource)).To(Succeed())
 
-		Eventually(func() bool {
-			err := cl.Get(ctx, client.ObjectKeyFromObject(networkResource), networkResource)
-			return apierrors.IsNotFound(err)
-		}).Should(BeTrue(), "Expected not found error")
+		Eventually(
+			cl.Get(ctx, client.ObjectKeyFromObject(networkResource), networkResource),
```
JoelSpeed, line 263: Same gotcha as above

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -463,10 +462,12 @@ var _ = Describe("Cloud config sync reconciler", func() {
 		co := &configv1.ClusterOperator{}
 		err := cl.Get(context.Background(), client.ObjectKey{Name: clusterOperatorName}, co)
 		if err == nil || !apierrors.IsNotFound(err) {
-			Eventually(func() bool {
-				err := cl.Delete(context.Background(), co)
-				return err == nil || apierrors.IsNotFound(err)
-			}).Should(BeTrue())
+			Eventually(
+				cl.Delete(context.Background(), co),
```
JoelSpeed, line 466: Wrapping gotcha as above

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -478,24 +479,23 @@ var _ = Describe("Cloud config sync reconciler", func() {
 		// omitted error intentionally, 404 might be there for some cases
 		cl.Delete(ctx, infra) //nolint:errcheck
 		Eventually(
-			apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(infra), infra)),
-		).Should(BeTrue())
+			cl.Get(ctx, client.ObjectKeyFromObject(infra), infra),
```
JoelSpeed, line 482: Wrapping gotcha

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -478,24 +479,23 @@ var _ = Describe("Cloud config sync reconciler", func() {
...
 
 		networkResource := makeNetworkResource()
 		Expect(cl.Delete(ctx, networkResource)).To(Succeed())
 
-		Eventually(func() bool {
-			err := cl.Get(ctx, client.ObjectKeyFromObject(networkResource), networkResource)
-			return apierrors.IsNotFound(err)
-		}).Should(BeTrue(), "Expected not found error")
+		Eventually(
+			cl.Get(ctx, client.ObjectKeyFromObject(networkResource), networkResource),
```
JoelSpeed, line 489: Wrapping gotcha

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -478,24 +479,23 @@ var _ = Describe("Cloud config sync reconciler", func() {
...
+		).Should(MatchError(apierrors.IsNotFound, "IsNotFound"))
 
 		allCMs := &corev1.ConfigMapList{}
 		Expect(cl.List(ctx, allCMs)).To(Succeed())
 		for _, cm := range allCMs.Items {
 			Expect(cl.Delete(ctx, cm.DeepCopy(), deleteOptions)).To(Succeed())
 			Eventually(
-				apierrors.IsNotFound(cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{})),
-			).Should(BeTrue())
+				cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{}),
```
JoelSpeed, line 497: Wrapping gotcha

pkg/controllers/cloud_config_sync_controller_test.go
nrb: Might be new. I'm a little surprised there's not a matcher already defined that does this same thing, but I wasn't able to find one.

PR #329, opened 2024-02-05
OCPBUGS-13106: Increase GCP Concurrent Service Syncs to 10

JoelSpeed:
This should increase the reactiveness of the GCP service controller, meaning that creation and deletion of load balancers is quicker.

/hold

I would like to review the service controller implementation within GCP and run payload testing before we merge this

### DIFF
--- a/pkg/cloud/gcp/assets/cloud-controller-manager.yaml
+++ b/pkg/cloud/gcp/assets/cloud-controller-manager.yaml
@@ -85,6 +85,7 @@ spec:
                 --v=3 \
                 --cloud-config=$(CLOUD_CONFIG) \
                 --cloud-provider=gce \
+                --concurrent-service-syncs=10 \
                 --controllers=*,-nodeipam \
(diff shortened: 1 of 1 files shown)


JoelSpeed:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.16-upgrade-from-stable-4.15-e2e-gcp-ovn-upgrade 10

JoelSpeed:
Given feedback from QE, and looking at the test results here, I think this is safe to proceed with. Will want to monitor GCP in the short term to make sure we don't see any long term issues
/approve
/hold cancel

elmiko:
this looks ok to me, i think my main concern would be if the cpu usage grows on extremely large clusters. if we want to test more, i wonder if we could try this out in a cluster with say 100 nodes, or maybe run something that creates a lot of services so that we could see if it is getting paused or affects performance.

200m seems like a decent amount of cpu, i'm not sure what we would up the request to.

/lgtm

PR #332, opened 2024-02-12
OCPBUGS-29220: Set AZURE_FEDERATED_TOKEN_FILE on CCM container

JoelSpeed:
The Auth provider in the CCM has changed in 4.16 and is now relying on a different field within the Azure config, `AADClientCertPath`. It is not clear to me if this is intentional or not.

The field in the config can be overridden by setting this environment variable. It appears that this will be an easier long term fix while we ascertain if the upstream change should be reverted.


JoelSpeed:
/retest-required

Manual OIDC job has successfully installed a cluster, so the patch is working

elmiko (commented):
/approve
/lgtm

PR #334, opened 2024-02-28
OCPCLOUD-2514: Remove reliance on feature gates for whether to deploy CCMs

JoelSpeed:
We no longer need to observe the cloud provider external feature gates to determine if the CCMs should be applied or not. Now, only the platform type is observed.
This PR cleans up the references I could find to using a feature gate for external cloud provider enablement.


pkg/controllers/clusteroperator_controller_test.go
```diff
@@ -228,10 +228,7 @@ var _ = Describe("Component sync controller", func() {
 	var operands []client.Object
 	var watcher mockedWatcher
 
-	externalFeatureGateAccessor := featuregates.NewHardcodedFeatureGateAccess(
-		[]configv1.FeatureGateName{configv1.FeatureGateExternalCloudProvider},
-		nil,
-	)
+	featureGateAccessor := featuregates.NewHardcodedFeatureGateAccess(nil, nil)
```
elmiko, line 231: just a question, looking at the tests it seems like we can factor out this variable. what am i missing?

elmiko (commented):
happy to approve this, just a quick question

pkg/controllers/clusteroperator_controller_test.go
JoelSpeed: The clusteroperator still uses a featuregateaccess underneath to render feature gates for the operand resources, so I had left it in, in case we need to update the cases, but, I think you're right, that would probably be tested elsewhere so we maybe don't need it in the test cases here

pkg/controllers/clusteroperator_controller_test.go
```diff
@@ -494,24 +483,10 @@ var _ = Describe("Component sync controller", func() {
 					Type: configv1.KubevirtPlatformType,
 				},
 			},
-			featureGate:       externalFeatureGateAccessor,
 			kcmStatus:         kcmStatus,
 			coStatus:          coStatus,
 			expectProvisioned: false,
 		}),
-		Entry("Should provision resources for AWS if external FeatureGate is not present", testCase{
```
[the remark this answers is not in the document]
elmiko: yay for cleanup!

elmiko (commented):
/approve
/lgtm

elmiko:
saw this error on the kcmo update as well, not sure what they mean though:

```
[sig-arch] Managed cluster [It] should ensure pods use downstream images from our release image with proper ImagePullPolicy [apigroup:config.openshift.io] [Suite:openshift/conformance/parallel]
  github.com/openshift/origin/test/extended/operators/images.go:22

    [FAILED] unable to decode release payload with error: invalid character 'w' looking for beginning of value
```

that doesn't seem related to this change

PR #335, opened 2024-03-01
OCPCLOUD-2515: remove CloudControllerOwner condition logic

elmiko:
This change removes the dependency on the CloudControllerOwner condition as it is no longer needed on OpenShift version 4.15 and later due to the external cloud controllers being the default with no rollback to in-tree. This change removes the CloudControllerOwner related conditions on the ClusterOperator if it is present.

### DIFF
--- a/pkg/controllers/status.go
+++ b/pkg/controllers/status.go
@@ -129,21 +129,30 @@ func (r *ClusterOperatorStatusClient) setStatusAvailable(ctx context.Context, ov
 	return r.syncStatus(ctx, co, conds, overrides)
 }
 
-// setCloudControllerOwnerCondition sets the CloudControllerOwner condition to True, with the given reason and message.
-func (r *CloudOperatorReconciler) setCloudControllerOwnerCondition(ctx context.Context) error {
+// clearCloudControllerOwnerCondition clears the CloudControllerOwner condition. This condition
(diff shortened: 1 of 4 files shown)


pkg/controllers/status.go
```diff
@@ -129,21 +129,30 @@ func (r *ClusterOperatorStatusClient) setStatusAvailable(ctx context.Context, ov
...
+		return nil
 	}
 
+	// if we get here, that means the condition exists and we want to remove it
+	v1helpers.RemoveStatusCondition(&co.Status.Conditions, cloudControllerOwnershipCondition)
 	co.Status.Versions = []configv1.OperandVersion{{Name: operatorVersionKey, Version: r.ReleaseVersion}}
-	klog.V(2).Info("Setting condition CloudControllerOwner to True")
-	return r.syncStatus(ctx, co, conds, nil)
+	klog.V(2).Info("Removing CloudControllerOwner condition")
+	return r.syncStatus(ctx, co, nil, nil)
```
JoelSpeed, line 155: Why does the third parameter here become nil?

elmiko: looking at the implementation, that third parameter is a list of conditions that will be applied to the conditions that are already on the ClusterOperator. it basically adds nothing (aside from extra cycles) to pass in the same list of conditions that are already on the ClusterOperator.

unless i misread the code, which is possible XD

PR #336, opened 2024-03-07
Bug OCPBUGS-29605: Allow to patch events in OpenStack RBAC

MaysaMacedo:
With new events being adding to cloud-provider-openstack it's now required to also allow patching events.


MaysaMacedo:
/cc @dulek 

This is required for https://github.com/kubernetes/cloud-provider-openstack/pull/2557

PR #337, opened 2024-04-03
OCPBUGS-13680: Pass --cluster-name to OpenStack CCM

dulek:
By an terrible oversight, since the in-tree, OpenStack cloud provider was run without passing `--cluster-name` argument, meaning that it just used the default of "kubernetes" as the cluster name when constructing the names of the LB resources. Obviously this led to conflicts when there were multiple OCP clusters in a single OpenStack tenant and it's still a challenge for example for QE running tests on a single cloud.

This issue couldn't be fixed easily, because we haven't had an upgrade story solved. Now kubernetes/cloud-provider-openstack#2552 is an attempt to make CCM rename the LBs when the cluster-name is reconfigured. This should allow us to finally solve the name conflict issues.

This commit makes sure that CCMO deploys OpenStack CCM passing `--cluster-name`.

### DIFF
--- a/pkg/cloud/openstack/assets/deployment.yaml
+++ b/pkg/cloud/openstack/assets/deployment.yaml
@@ -54,6 +54,8 @@ spec:
           env:
             - name: CLOUD_CONFIG
               value: /etc/openstack/config/cloud.conf
+            - name: OCP_INFRASTRUCTURE_NAME
+              value: {{ .infrastructureName }}
           command:
           - /bin/bash
           - -c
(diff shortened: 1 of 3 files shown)


dulek:
/hold

Needs to wait for the upstream PR to be merged and synced into downstream first.

dulek:
@stephenfin, @gryf - please take a look. It's not merging yet, but I'd like you to be aware of this.

dulek:
The CPO change is merged downstream in cloud-provider-openstack#276, so this is good to go.

/hold cancel

PR #340, opened 2024-04-17
WRKLDS-1015: tolerate node-role.kubernetes.io/control-plane:NoExecute

knelasevero:
Required to be able to advise customers to follow workflow described in https://github.com/openshift/enhancements/pull/1583. If they want to taint control-plane nodes to avoid any other thing to run there (while enforcing it further with validating admission policies), control plane workloads need to tolerate that taint.


knelasevero:
Not sure if there is some other place where the taint needs to be tolerated

knelasevero:
> Please update the [test suite](https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/master/pkg/cloud/cloud_test.go) to ensure that all pods tolerate this taint (we already have examples of this).
> 
> What is the status of the EP? It's not merged yet and looks like it still needs substantial review? When are you hoping to merge the EP and the feature in general?

The EP started as a feature that we would provide as a new admission plugin/controller, but after review what we are instead providing is basically already available out of the box (well, at least tech preview with ValidatingAdmisionPolicies with CEL). So the KEP is now mostly up to have a reference of this workflow that we are/will recommend customers to follow (through documentation).

I will get the EP merged soon.

knelasevero:
Will update the tests

PR #342, opened 2024-05-10
OCPBUGS-33405: update azure and ash tolerations on node manager

elmiko:
In our documentation, we instruct users to add a NoExecute taint for the node-role.kubernetes.io/infra key. This is not tolerated by the azure node managers and can cause a race condition when new nodes are added that can prevent the node manager from running on a new node.

This change updates the tolerations for azure and ash to allow the infra taint, and it also makes the ash permissions more tolerant (same as azure) to prevent future failures.


JoelSpeed (commented):
/approve
/lgtm

PR #344, opened 2024-05-13
NO-JIRA: Ensure termination message policy in tests

JoelSpeed:
This is a new policy in OpenShift, and, while we are compliant, I would like to keep it that way. This test should prevent regression


elmiko:
i can't imagine there is a problem building the image, but it seems like the e2e-vsphere-ovn failed on creation. and while i'm typing this more are failing. is there a build issue?

damdo:
@elmiko it looks like a sigkill

PR #348, opened 2024-05-30
OCPBUGS-34689: Nutanix CCM pods Cipher Suites vulnerability issue

yanhua121:
https://issues.redhat.com/browse/OCPBUGS-34689
Customer running Openshift on Nutanix AHV and their Tenable Security Scan reported the following vulnerability on the Nutanix Cloud Controller Manager Deployment. 
We can resolve it by adding the "--tls-cipher-suites" option to the cli in the nutanix-ccm deployment manifest.

### DIFF
--- a/pkg/cloud/nutanix/assets/cloud-controller-manager-deployment.yaml
+++ b/pkg/cloud/nutanix/assets/cloud-controller-manager-deployment.yaml
@@ -97,7 +97,8 @@ spec:
                 --leader-elect-lease-duration=137s \
                 --leader-elect-renew-deadline=107s \
                 --leader-elect-retry-period=26s \
-                --leader-elect-resource-namespace=openshift-cloud-controller-manager
+                --leader-elect-resource-namespace=openshift-cloud-controller-manager \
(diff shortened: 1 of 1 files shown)


yanhua121:
@elmiko I don't know why the tests failed, even the lint and vet tests. My local run of the lint/vet tests passed. This PR only changed the nutanix-ccm-manager deployment yaml file.

elmiko:
i just gave a quick look and i wonder if some of these tests are failing due to an out of memory kill. the lint test looks like it has an issue with an unused import, but otherwise the tests seem to fail during the build step.

elmiko:
not sure why the linter would fail now, this change does not seem related:

```
 pkg/controllers/resourceapply/config_hash_test.go:14:2: "k8s.io/api/apps/v1" imported as appsv1 and not used (typecheck)
	appsv1 "k8s.io/api/apps/v1"
	^ 
```

elmiko:
i just looked at the file referenced by the failing lint test, and the test is wrong, that import is used.

elmiko:
the ci config update seems to have helped, we'll see what happens with the e2e tests

yanhua121:
> the ci config update seems to have helped, we'll see what happens with the e2e tests

@elmiko Thanks for resolving the test env. issue. Could you review the PR, it only has the option "--tls-cipher-suites" added to the nutanix-ccm-manager deployment yaml.

elmiko:
i was waiting to see the tests pass before reviewing

PR #349, opened 2024-05-31
DO NOT MERGE

elmiko:
testing out some failures in ci


elmiko:
the linter is wrong, not sure why

elmiko:
think we have this solved in the release config

PR #355, opened 2024-06-18
OCPBUGS-35798: PowerVS: Remove bind address

hamzy:
Remove the --bind-address to localhost as it makes the liveness test fail with the following error:

Liveness probe error: Get "https://192.168.169.11:10258/healthz": dial tcp 192.168.169.11:10258: connect: connection refused

### DIFF
--- a/pkg/cloud/powervs/assets/deployment.yaml
+++ b/pkg/cloud/powervs/assets/deployment.yaml
@@ -73,7 +73,6 @@ spec:
               source /etc/kubernetes/apiserver-url.env
             fi
             exec /bin/ibm-cloud-controller-manager \
-            --bind-address=127.0.0.1 \
             --use-service-account-credentials=true \
             --configure-cloud-routes=false \
             --cloud-provider=ibm \


mkumatag:
/cc @Karthik-K-N 

ptal

mjturek:
we'll also need this in 4.16

pkg/cloud/powervs/assets/deployment.yaml
```diff
@@ -73,7 +73,6 @@ spec:
               source /etc/kubernetes/apiserver-url.env
             fi
             exec /bin/ibm-cloud-controller-manager \
-            --bind-address=127.0.0.1 \
```
Karthik-K-N, line 76: @hamzy, Trying to understand what made it to fail now or was it failing from long time? also we almost do same thing like IBMCloud platform, did you happen to check whats the status there as I could see the same flag being used there as well.

hamzy: @Karthik-K-N The cluster completes creation.  It does not pass CI testing.  I am not involved in IBMCloud platform development or testing.  This change only affects PowerVS.

mjturek: this has been failing most of the 4.16 cycle. @cjschaef has IBM Cloud seen this at all?

hamzy: Let's not hold up for that though.

Karthik-K-N: I would better suggest to know @cjschaef opinion. Since both PowerVS and IBMCloud use same CCM code, It would be better to be in sync.

hamzy: @Karthik-K-N How long are we going to block PowerVS CI on no response?

cjschaef: We have experienced something similar during OCP Conformance (MonitorTests) related to liveness probes.
https://issues.redhat.com/browse/OCPBUGS-30267

The proposed change does not affect IBM Cloud VPC, since it is scoped to PowerVS, but I would be interested to see if this resolves the issue we have seen on 4.15+ with the MonitorTest results. Are those the same failures you mention with the CI?

hamzy: Yes, this should fix the MonitorTest failure.
@Karthik-K-N can we finally progress?!

Karthik-K-N (commented):
/lgtm

elmiko (commented):
/approve

PR #360, opened 2024-07-12
OCPBUGS-41936: IBMCloud: Modify liveness probe for IBM Cloud CCM to use loopback address

jared-hayes-dev:
per testing + discussion we chose to modify the liveness probe to use the loopback address over removing the bind address argument (security concerns with the former)


cjschaef:
`e2e-ibmcloud-ovn` failure was `[sig-arch] events should not repeat pathologically for ns/openshift-oauth-apiserver` which appears to tie back to https://issues.redhat.com/browse/OCPBUGS-30267 (underlying Infrastructure limitations).

I think otherwise, this has addressed the CCM liveness probe errors seen in MonitorTests.

elmiko (commented):
apologies for missing this, i think the changes make sense to me. let's see if the ibmcloud tests pass then i'm happy to approve

/lgtm

elmiko:
it appears that the failures in the test runs are related to the csi driver not having the proper permissions, and what looks like a failure based on historical data. i do not believe either are related to this change so i am going to approve this.

/approve
/lgtm

elmiko:
@jared-hayes-dev @jeffnowicki do we have a jira ticket for this?

elmiko:
this is needed by https://github.com/openshift/cloud-provider-ibm/pull/71

cjschaef:
/retitle OCPBUGS-41936: IBMCloud: Modify liveness probe for IBM Cloud CCM to use loopback address

PR #361, opened 2024-08-15
NO-JIRA: Update envtest configuration to new index format

JoelSpeed:
Demonstration of using new envtest-releases.yaml file in the API repo to source envtest utilities.


theobarberbany:
assuming that all the tests pass

/lgtm

PR #362, opened 2024-08-19
OCPBUGS-16728: Add admission policy to deny changing an AWS LB type on an existing service

JoelSpeed:
The AWS CCM does not account for changes to the type of the load balancer. The recommended and supportable approach is for users to delete and recreate the service, rather than changing the annotation live.

This PR adds policy to prevent services from adding/removing/changing the annotation to control which kind of load balancer the service represents, such that the annotation must be set on create for NLB load balancers.

This should prevent any leaking of NLB/CLB resources when users mistakenly change the load balancer type.


pkg/cloud/aws/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,19 @@
...
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["UPDATE"]
+      resources:   ["services"]
+  validations:
+    - expression: |
+        (has(object.objectMeta.annotations) && 'service.beta.kubernetes.io/aws-load-balancer-type' in object.objectMeta.annotations) ==
+          (has(oldObject.objectMeta.annotations) && 'service.beta.kubernetes.io/aws-load-balancer-type' in oldObject.objectMeta.annotations) &&
+          object.objectMeta.annotations['service.beta.kubernetes.io/aws-load-balancer-type'] == oldObject.objectMeta.annotations['service.beta.kubernetes.io/aws-load-balancer-type']
```
theobarberbany, line 17: I'm not 100% following how this works, so this is more for my understanding - shout if I'm wrong: 

We assert that our old object and our new object have both got the annotation matching `service.beta.kubernetes.io/aws-load-balancer-type`

then we check the value of the annotation remains equal? i.e does not change

```
object.objectMeta.annotations['service.beta.kubernetes.io/aws-load-balancer-type'] == oldObject.objectMeta.annotations['service.beta.kubernetes.io/aws-load-balancer-type']
```
If this annotation were being changed, we would fail the validation - so we  would fail that update, disasllowing changing it's value.

Or am I missing something?

JoelSpeed: > We assert that our old object and our new object have both got the annotation matching service.beta.kubernetes.io/aws-load-balancer-type. We need this because otherwise the controller would not be able to set the value initially?

You're close but missing one nuance, we check that whether "they have the annotation or not" is equal. If one has the annotation and the other does not, then the user changed something, and we aren't allowing that here.

theobarberbany: Ahh gotcha! Makes sense!

sunzhaohua2:
/verified by @sunzhaohua2
Tested three cases with this pr, add/update/remove aws-load-balancer-type annotation is not allowed, check aws console, no resource leak.
- Create one service without the annotation, and try to add it.
- Create one service with the annotation and try to change the value from nlb to clb(or clb to nlb)
- Create one service with the annotation and try to remove the annotation


```
$ oc edit svc lb-service-unsecure
# services "lb-service-unsecure" was not valid:
# * : ValidatingAdmissionPolicy 'openshift-cloud-controller-manager-cloud-provider-aws' with binding 'openshift-cloud-controller-manager-cloud-provider-aws' denied request: The annotation 'service.beta.kubernetes.io/aws-load-balancer-type' may not be added, removed, or have its value changed. Changing the type of an existing load balancer is not supported.
```
Already added test case https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-75886

PR #368, opened 2024-10-01
OCPCLOUD-2737: Bump k8s to 1.31

racheljpg:
Hello! This is a PR to bump the k8s dependencies to 1.31.0. Thank you!


Makefile
```diff
@@ -3,7 +3,7 @@
 IMG ?= controller:latest
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.30.3
```
JoelSpeed, line 6: Try 1.31.1 please, I don't think we published 1.31.0

go.mod
```diff
@@ -14,21 +14,21 @@ require (
...
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.30.1
-	k8s.io/apiextensions-apiserver v0.30.1
-	k8s.io/apimachinery v0.30.1
-	k8s.io/client-go v0.30.1
+	k8s.io/api v0.31.1
+	k8s.io/apiextensions-apiserver v0.31.1
+	k8s.io/apimachinery v0.31.1
+	k8s.io/client-go v0.31.1
 	k8s.io/cloud-provider-vsphere v1.30.1
```
JoelSpeed, line 27: Does this also need to be updated?

go.mod
```diff
@@ -14,21 +14,21 @@ require (
...
-	k8s.io/controller-manager v0.30.1
-	k8s.io/klog/v2 v2.120.1
-	k8s.io/utils v0.0.0-20240310230437-4693a0247e57
+	k8s.io/component-base v0.31.1
+	k8s.io/controller-manager v0.31.1
+	k8s.io/klog/v2 v2.130.1
+	k8s.io/utils v0.0.0-20240711033017-18e509b52bc8
 	sigs.k8s.io/cloud-provider-azure v1.30.3
 	sigs.k8s.io/cloud-provider-azure/pkg/azclient v0.0.25
 	sigs.k8s.io/controller-runtime v0.18.4
```
JoelSpeed, line 34: What about these, I think these probably warrant an update too

go.mod
```diff
@@ -1,39 +1,37 @@
 module github.com/openshift/cluster-cloud-controller-manager-operator
 
-go 1.22.0
-
-toolchain go1.22.2
+go 1.23.0
```
JoelSpeed, line 3: Must stay as 1.22 for now

go.mod
```diff
@@ -1,39 +1,37 @@
...
 	github.com/go-logr/logr v1.4.2
 	github.com/golangci/golangci-lint v1.54.2
-	github.com/onsi/ginkgo/v2 v2.19.0
-	github.com/onsi/gomega v1.33.1
+	github.com/onsi/ginkgo/v2 v2.20.1
+	github.com/onsi/gomega v1.34.2
 	github.com/openshift/api v0.0.0-20240610131338-a837bee24500
 	github.com/openshift/client-go v0.0.0-20240528061634-b054aa794d87
 	github.com/openshift/cluster-api-actuator-pkg/testutils v0.0.0-20240607201500-81075cf8e11a
 	github.com/openshift/library-go v0.0.0-20240607134135-aed018c215a1
```
JoelSpeed, line 14: OpenShift libraries also want updating?

go.mod
```diff
@@ -1,39 +1,37 @@
...
-	k8s.io/controller-manager v0.30.1
-	k8s.io/klog/v2 v2.120.1
-	k8s.io/utils v0.0.0-20240310230437-4693a0247e57
+	k8s.io/api v0.31.2
+	k8s.io/apiextensions-apiserver v0.31.2
+	k8s.io/apimachinery v0.32.0-alpha.3
+	k8s.io/client-go v0.31.2
+	k8s.io/cloud-provider-vsphere v1.31.1
+	k8s.io/component-base v0.31.2
+	k8s.io/controller-manager v0.31.2
```
JoelSpeed, line 27: These should all use the same version, but are spread across three presently?

go.mod
```diff
@@ -1,39 +1,37 @@
...
+	k8s.io/klog/v2 v2.130.1
+	k8s.io/utils v0.0.0-20240711033017-18e509b52bc8
 	sigs.k8s.io/cloud-provider-azure v1.30.3
 	sigs.k8s.io/cloud-provider-azure/pkg/azclient v0.0.25
```
JoelSpeed, line 31: Any updates for these two?

JoelSpeed (commented):
Your recent updates ended up in the vendoring commit which is making this difficult to review, please split go.mod and vendor updates to separate commits

go.mod
racheljpg: Says they both require go >= 1.23.1

go.mod
```diff
@@ -1,37 +1,37 @@
 module github.com/openshift/cluster-cloud-controller-manager-operator
 
-go 1.22.0
+go 1.22.5
 
-toolchain go1.22.2
+toolchain go1.23.0
```
JoelSpeed, line 5: This will break CI I think, can you manually revert just this line?

racheljpg: Yes my bad, I'm not sure what I've updated that's done this!

JoelSpeed:
The failed vendor check appears to be asking you to remove the toolchain directive altogether, can you take a look at that?

shellyyang1989:
Hi @racheljpg Thanks for the effort! I see that the E2E CI jobs for the PR are currently failing. When you have a chance, could you take a look at the issues and address them?

JoelSpeed:
There's an issue with https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/42446fc73c4e806d165d4bbf5979f2be854e9faf/cmd/config-sync-controllers/main.go#L134-L158, as the two controllers are registering with the same name. Lets make sure each controller's `SetupWithManager` registers the controller with an appropriate name

racheljpg:
Hey @JoelSpeed, I've registered names for them using the "Named" function from controller-runtime, is this how you imagined it should be done? Is there a better way?

JoelSpeed:
> Hey @JoelSpeed, I've registered names for them using the "Named" function from controller-runtime, is this how you imagined it should be done? Is there a better way?

Seems like the appropriate approach, thanks

I _think_ CI should be happy now

/lgtm
/approve

racheljpg:
@JoelSpeed, they've been running a while, so I hope you're right! I was just wondering if maybe I should have done something in main.go also/instead, but we'll see

PR #369, opened 2024-10-11
OCPBUGS-43164: set goimport version to v0.24.0

elmiko:
This change is being proposed because the release-4.16 branch uses golang version 1.21 and the versions of goimports greater than 0.24.0 require golang version 1.22+.

### DIFF
--- a/hack/goimports.sh
+++ b/hack/goimports.sh
@@ -7,7 +7,7 @@ REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/..")
 
 function runGoimports() {
   local GOIMPORTS_PATH=$REPO_ROOT/bin/goimports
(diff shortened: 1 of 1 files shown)


elmiko:
this is needed by #366

JoelSpeed:
This will need an appropriate chain of bug backports to merge, or staff eng to override

Should we pin, rather than using latest, all the way through our releases?

elmiko:
i think we need to start pinning, i don't think we should backport this change but we should start pinning it. i have made the bugs for the backport chain.

elmiko:
had a discussion about this internally and we think the best solution is to migrate this repository to utilize the goimports from golangci-lint, which it is already including.

/hold

elmiko:
i am going to create a new PR that will fix this more thoroughly, and then cherry-pick the backport.

PR #376, opened 2024-11-13
OCPBUGS-44126: update azure permissions

elmiko:
this change adds some of the linked action permissions that are needed to ensure proper operation.


shellyyang1989:
cc @JoelSpeed @theobarberbany @nrb for review and visibility. Thanks.

PR #379, opened 2025-01-22
OCPBUGS-48481: annotation validation policy

miyadav:
if the annotation `cloud.google.com/network-tier` was used in service resource and it is left without value or incorrect value(say xyz) , the service goes to `Pending` status with only warning 
`  Warning  SyncLoadBalancerFailed  1s (x3 over 17s)  service-controller  Error syncing load balancer: failed to ensure load balancer: unsupported network tier: "xyz"`
 , to avoid that , added the VAP .

### DIFF
--- a/pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
+++ b/pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
@@ -1,0 +1,28 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: azure-load-balancer-tcp-idle-timeout-annotation-validation-policy
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["services"]
+  variables:
+    - name: hasIdleTimeout
+      expression: has(object.metadata.annotations) && 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' in object.metadata.annotations
+    - name: idleTimeoutValue
+      expression: object.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
+    - name: hasOldIdleTimeout
+      expression: oldObject != null && oldObject.metadata.?annotations != null && 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' in oldObject.metadata.annotations
+    - name: oldIdleTimeoutValue
+      expression: oldObject.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
+    - name: valueHasChanged
+      expression: variables.hasIdleTimeout && (!variables.hasOldIdleTimeout || variables.oldIdleTimeoutValue != variables.idleTimeoutValue) 
+  validations:
+    - expression: "!variables.hasIdleTimeout || !variables.valueHasChanged || variables.idleTimeoutValue.matches('^[0-9]+$')"
+      message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be an integer"
+    - expression: "!variables.hasIdleTimeout || !variables.valueHasChanged || (int(variables.idleTimeoutValue) >= 4 && int(variables.idleTimeoutValue) <= 100)"
+      message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be no less than 4 and no more than 100"
+
--- a/pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
+++ b/pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
@@ -1,4 +1,4 @@
-apiVersion: admissionregistration.k8s.io/v1
+apiVersion: admissionregistration.k8s.io/v1beta1
 kind: ValidatingAdmissionPolicy
 metadata:
   name: azure-load-balancer-tcp-idle-timeout-annotation-validation-policy
--- a/pkg/cloud/gcp/assets/validating-admission-policy.yaml
+++ b/pkg/cloud/gcp/assets/validating-admission-policy.yaml
@@ -1,0 +1,26 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: network-tier-annotation-validation-policy
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["services"]
+  variables:
+    - name: hasNetworkTier
+      expression: has(object.metadata.annotations) && 'cloud.google.com/network-tier' in object.metadata.annotations
+    - name: networkTierValue
+      expression: object.metadata.?annotations['cloud.google.com/network-tier'].orValue("")
+    - name: hasOldNetworkTier
+      expression: oldObject != null && oldObject.metadata.?annotations != null && 'cloud.google.com/network-tier' in oldObject.metadata.annotations
+    - name: oldNetworkTierValue
+      expression: oldObject.metadata.?annotations['cloud.google.com/network-tier'].orValue("")
+    - name: valueHasChanged
+      expression: variables.hasNetworkTier && (!variables.hasOldNetworkTier || variables.oldNetworkTierValue != variables.networkTierValue)
+  validations:
+    - expression: |
+        !variables.valueHasChanged || variables.networkTierValue in ['Standard', 'Premium']
+      message: "The annotation 'cloud.google.com/network-tier', if specified, must be either 'Standard' or 'Premium'."
--- a/pkg/cloud/azure/assets/validating-admission-service-annotation-policy-binding.yaml
+++ b/pkg/cloud/azure/assets/validating-admission-service-annotation-policy-binding.yaml
@@ -1,0 +1,7 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: azure-load-balancer-tcp-idle-timeout-validation-annotation-binding
+spec:
+  policyName: azure-load-balancer-tcp-idle-timeout-annotation-validation-policy
+  validationActions: ["Deny"]
--- a/pkg/cloud/azure/assets/validating-admission-service-annotation-policy-binding.yaml
+++ b/pkg/cloud/azure/assets/validating-admission-service-annotation-policy-binding.yaml
@@ -1,4 +1,4 @@
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingAdmissionPolicyBinding
 metadata:
   name: azure-load-balancer-tcp-idle-timeout-validation-annotation-binding
--- a/pkg/cloud/azure/azure.go
+++ b/pkg/cloud/azure/azure.go
@@ -36,6 +36,8 @@ var (
 		{ReferenceObject: &rbacv1.ClusterRoleBinding{}, EmbedFsPath: "assets/azure-cloud-controller-manager-clusterrolebinding.yaml"},
 		{ReferenceObject: &admissionregistrationv1.ValidatingAdmissionPolicy{}, EmbedFsPath: "assets/validating-admission-policy.yaml"},
 		{ReferenceObject: &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}, EmbedFsPath: "assets/validating-admission-policy-binding.yaml"},
+		{ReferenceObject: &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}, EmbedFsPath: "assets/validating-admission-service-annotation-policy-binding.yaml"},
+		{ReferenceObject: &admissionregistrationv1.ValidatingAdmissionPolicy{}, EmbedFsPath: "assets/validating-admission-service-annotation-policy.yaml"},
 	}
 )
 
--- a/pkg/cloud/azure/azure_test.go
+++ b/pkg/cloud/azure/azure_test.go
@@ -88,7 +88,7 @@ func TestResourcesRenderingSmoke(t *testing.T) {
 			}
 
 			resources := assets.GetRenderedResources()
-			assert.Len(t, resources, 8)
+			assert.Len(t, resources, 10)
 		})
 	}
 }
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -129,22 +129,26 @@ func TestGetResources(t *testing.T) {
 	}, {
 		name:                  "GCP resources returned as expected",
 		testPlatform:          platformsMap[string(configv1.GCPPlatformType)],
-		expectedResourceCount: 4,
+		expectedResourceCount: 6,
 		expectedResourcesKindName: []string{
 			"Deployment/gcp-cloud-controller-manager",
 			"PodDisruptionBudget/gcp-cloud-controller-manager",
 			"ClusterRole/gcp-cloud-controller-manager",
 			"ClusterRoleBinding/gcp-cloud-controller-manager:cloud-provider",
+			"ValidatingAdmissionPolicyBinding/network-tier-annotation-binding",
+			"ValidatingAdmissionPolicy/network-tier-annotation-validation-policy",
 		},
 	}, {
 		name:                  "GCP resources returned as expected with single node cluster",
 		testPlatform:          platformsMap[string(configv1.GCPPlatformType)],
-		expectedResourceCount: 3,
+		expectedResourceCount: 5,
 		singleReplica:         true,
 		expectedResourcesKindName: []string{
 			"Deployment/gcp-cloud-controller-manager",
 			"ClusterRole/gcp-cloud-controller-manager",
 			"ClusterRoleBinding/gcp-cloud-controller-manager:cloud-provider",
+			"ValidatingAdmissionPolicyBinding/network-tier-annotation-binding",
+			"ValidatingAdmissionPolicy/network-tier-annotation-validation-policy",
 		},
 	}, {
 		name:                  "Azure resources returned as expected",
@@ -159,6 +163,8 @@ func TestGetResources(t *testing.T) {
 			"ClusterRoleBinding/cloud-controller-manager:azure-cloud-controller-manager",
 			"ValidatingAdmissionPolicy/openshift-cloud-controller-manager-cloud-provider-azure-node-admission",
 			"ValidatingAdmissionPolicyBinding/openshift-cloud-controller-manager-cloud-provider-azure-node-admission",
+			"ValidatingAdmissionPolicyBinding/azure-load-balancer-tcp-idle-timeout-validation-annotation-binding",
+			"ValidatingAdmissionPolicy/azure-load-balancer-tcp-idle-timeout-annotation-validation-policy",
 			"PodDisruptionBudget/azure-cloud-controller-manager",
 		},
 	}, {
@@ -175,6 +181,8 @@ func TestGetResources(t *testing.T) {
 			"ClusterRoleBinding/cloud-controller-manager:azure-cloud-controller-manager",
 			"ValidatingAdmissionPolicy/openshift-cloud-controller-manager-cloud-provider-azure-node-admission",
 			"ValidatingAdmissionPolicyBinding/openshift-cloud-controller-manager-cloud-provider-azure-node-admission",
+			"ValidatingAdmissionPolicyBinding/azure-load-balancer-tcp-idle-timeout-validation-annotation-binding",
+			"ValidatingAdmissionPolicy/azure-load-balancer-tcp-idle-timeout-annotation-validation-policy",
 		},
 	}, {
 		name:                  "Azure Stack resources returned as expected",
--- a/pkg/cloud/cloud_test.go
+++ b/pkg/cloud/cloud_test.go
@@ -153,7 +153,7 @@ func TestGetResources(t *testing.T) {
 	}, {
 		name:                  "Azure resources returned as expected",
 		testPlatform:          platformsMap[string(configv1.AzurePlatformType)],
-		expectedResourceCount: 9,
+		expectedResourceCount: 11,
 		expectedResourcesKindName: []string{
 			"Deployment/azure-cloud-controller-manager",
 			"DaemonSet/azure-cloud-node-manager",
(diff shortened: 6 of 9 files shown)


miyadav:
Validated for gcp , looks good , yet to check on azure cluster .

miyadav:
validated for azure as well.

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,17 @@
...
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["services"]
+  validations:
+    - expression: "(!has(object.metadata.annotations) || !('cloud.google.com/network-tier' in object.metadata.annotations) ||
+                   (object.metadata.annotations['cloud.google.com/network-tier'] == 'Standard' || object.metadata.annotations['cloud.google.com/network-tier'] == 'Premium') ||
+                   (has(oldObject.metadata.annotations) && oldObject.metadata.annotations['cloud.google.com/network-tier'] == object.metadata.annotations['cloud.google.com/network-tier']))"
```
theobarberbany, line 15: Do we want to allow existing bad configurations?

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: network-tier-annotation-validation-policy
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["services"]
+  validations:
+    - expression: "(!has(object.metadata.annotations) || !('cloud.google.com/network-tier' in object.metadata.annotations) ||
```
theobarberbany, line 13: I might be tempted to flip the logic here to make it easier to read, currently I read it like this: 

Allow if, No annotations at all: !has(object.metadata.annotations) or cloud.google.com/network-tier is not set, or. Annotation is one of the valid values (Standard or Premium), or The old and new values match (allow existing invalid configurations)

It could be:

Deny if:  The new annotation is present and invalid, AND (the oldObject either does not exist or that annotation has changed).

```
!(
   // If the annotation is present on the new object:
   has(object.metadata.annotations['cloud.google.com/network-tier'])

   // and it is not in [Standard, Premium]
   && !(object.metadata.annotations['cloud.google.com/network-tier'] in ['Standard','Premium'])

   // and either there was no oldObject OR
   // the old annotation differs from the new annotation:
   && (
     oldObject == null 
     || !has(oldObject.metadata.annotations['cloud.google.com/network-tier']) 
     || oldObject.metadata.annotations['cloud.google.com/network-tier'] 
          != object.metadata.annotations['cloud.google.com/network-tier']
   )
 )
```

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: azure-load-balancer-tcp-idle-timeout-annotation-validation-policy
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["services"]
+  validations:
+    - expression: |
```
theobarberbany, line 13: Same here, could this maybe change to: 

Deny if: The new annotation is present and does not match the regex, AND the value does not lie between 4 and 100.

I think the validation of the range can be done with a regex as well, although that might be getting unreadable. 

I

pkg/cloud/gcp/assets/validating-admission-policy.yaml
miyadav: got you @theobarberbany , let me try that , thanks

pkg/cloud/gcp/assets/validating-admission-policy.yaml
miyadav: We want to allow that , to not break upgrades I think ..

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
miyadav: @theobarberbany , does below looks good , we want to keep the old resources as is 

```
`        !('service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' in object.metadata.annotations) ||
        (
          object.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].matches('^[0-9]+$') &&
          int(object.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout']) >= 4 &&
          int(object.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout']) <= 100
        ) ||
        (
          oldObject != null &&
          oldObject.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'] == object.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout']
        )`
```

theobarberbany: yep, I think it looks good to me :)

miyadav: Thanks Theo , I got installation issues with changes , will share in slack , so I reverted the changes , I retest them again , what do you think  , if reverted look good , we can let them be ?

pkg/cloud/gcp/assets/validating-admission-policy.yaml
theobarberbany: Ok - i made a mistake here around how `has` works: 

I think this ought to work now :) I've tested on the CEL playground :)
```
!(
   // If the annotation is present on the new object:
   has(object.metadata.annotations) &&
   ('cloud.google.com/network-tier' in object.metadata.annotations)

   // and it is not in [Standard, Premium]
   && !(object.metadata.annotations['cloud.google.com/network-tier'] in ['Standard','Premium'])

   // and either there was no oldObject OR
   // the old annotation differs from the new annotation
   && (
     oldObject == null 
     || !has(oldObject.metadata.annotations) && ('cloud.google.com/network-tier' in oldObject.metadata.annotations) 
     || oldObject.metadata.annotations['cloud.google.com/network-tier'] 
          != object.metadata.annotations['cloud.google.com/network-tier']
   )
 )
```

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
JoelSpeed, line 13: This expression is quite long and complex, if we were to break it down using [variable composition](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#variable-composition) perhaps we could make this a little easier to read?

For example, if we break out two variables, one to check the presence of the annotation, and one to extract the value (using CEL optionals)

```
variables:
- name: hasIdleTimeout
  expression: has(object.metadata.annotations) && 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' in object.metadata.annotations
- name: idleTimeoutValue
  expression: object.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
```

We can then combine that into multiple expressions

```
validations:
- expression: !variables.hasIdleTimeout || variables.idleTimeoutValue.matches('^[0-9]+$')
  message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be an integer"
- expression: !variables.hasIdleTimeout || int(variables.idleTimeoutValue) >= 4 && int(variables.idleTimeoutValue) <= 100"
  message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be no less than 4 and no more than 100"
```

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: network-tier-annotation-validation-policy
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["services"]
+  validations:
+    - expression: |
```
JoelSpeed, line 13: As above, probably easier if we split this out into several variables and then compose a series of expressions from those, WDYT?

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+  validations:
+    - expression: | 
+       !(
+         // If the annotation is present on the new object:
+          has(object.metadata.annotations) &&
+          ('cloud.google.com/network-tier' in object.metadata.annotations)
+         // and it is not in [Standard, Premium]
+             && !(object.metadata.annotations['cloud.google.com/network-tier'] in ['Standard','Premium'])
+         // and either there was no oldObject OR
+        // the old annotation differs from the new annotation
```
JoelSpeed, line 21: Ratcheting, nice

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
JoelSpeed, line 13: Do we need to ratchet this validation as you have done on the GCP version?

miyadav: @JoelSpeed Getting `unkown tag <!variables.has..>`  when adding this ?

pkg/cloud/gcp/assets/validating-admission-policy.yaml
miyadav: yes it looks better.

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
JoelSpeed: I haven't tested this myself, you might need to play with the syntax, I was basing this on the documentation I linked.

Maybe `!(variables.hasIdleTimeout)` is required?

miyadav: yes I think so , will update this and validate.

miyadav: acked, will try it out. Thanks

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
```diff
@@ -0,0 +1,35 @@
...
+          variables.hasIdleTimeout
+          // and it is not a valid integer or out of the range 4-100
+          && (!variables.idleTimeoutValue.matches('^[0-9]+$') ||
+              int(variables.idleTimeoutValue) < 4 || 
+              int(variables.idleTimeoutValue) > 100)
+          // and either the old object does not exist, or it has a different value:
+          && (oldObject == null ||
+              !variables.hasOldIdleTimeout || 
+              oldObject.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'] != object.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'])
+        )
```
JoelSpeed, line 34: Is there a way to simplify this, or break this down at all? 

Perhaps we can leverage more for variables

```
variables:
- name: valueHasChanged
  expression: ?oldObject.metadata.annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("") != object.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
...
validations:
- expression: !variables.hasIdleTimeout || !variables.valueHasChanged || variables.idleTimeoutValue.matches('^[0-9]+$')
  message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be an integer"
- expression: !variables.hasIdleTimeout || || !variables.valueHasChanged || int(variables.idleTimeoutValue) >= 4 && int(variables.idleTimeoutValue) <= 100"
  message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be no less than 4 and no more than 100"
```
In these two expressions, we check either the annotation isn't there, or, it is the same as before, or it matches a condition of being acceptable. I think this would be easier to follow than the larger expression we currently have. Relies on the optional filling in when oldSelf is missing which I think we can do

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
theobarberbany: TIL [variable composition](https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#variable-composition), that's pretty cool!

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
miyadav: let me test this , thanks

miyadav: Made the changes , faced some issues in validation (logical) , but finally it worked , updated Jira with results.

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+metadata:
+  name: network-tier-annotation-validation-policy
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["services"]
+  validations:
```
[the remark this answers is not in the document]
miyadav: sure , will try that .. , also @JoelSpeed do we need different versions , just realized I have been using v1beta1 and v1 for   admissionregistration.k8s.io
for azure and gcp respectively , should both be v1 ?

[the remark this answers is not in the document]
JoelSpeed: Should just be `v1` for the API version yes

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,32 @@
...
+        variables.hasOldNetworkTier &&
+        variables.hasNetworkTier &&
+        variables.oldNetworkTierValue != variables.networkTierValue
+  validations:
+    - expression: |
+        !variables.hasNetworkTier || variables.networkTierValue in ['Standard', 'Premium']
+      message: "The annotation 'cloud.google.com/network-tier', if specified, must be either 'Standard' or 'Premium'."
+    - expression: |
+        !variables.hasNetworkTier || !variables.valueHasChanged || variables.networkTierValue in ['Standard', 'Premium']
+      message: "The annotation 'cloud.google.com/network-tier' must not change to an invalid value."
```
JoelSpeed, line 32: These validations won't allow ratcheting correctly.

You need to have only one of these two, the latter, but with the message from the previous

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - name: hasNetworkTier
+      expression: has(object.metadata.annotations) && 'cloud.google.com/network-tier' in object.metadata.annotations
+    - name: networkTierValue
+      expression: object.metadata.?annotations['cloud.google.com/network-tier'].orValue("")
+    - name: hasOldNetworkTier
+      expression: has(oldObject) && has(oldObject.metadata) && has(oldObject.metadata.annotations) && 'cloud.google.com/network-tier' in oldObject.metadata.annotations
+    - name: oldNetworkTierValue
+      expression: oldObject.metadata.?annotations['cloud.google.com/network-tier'].orValue("")
+    - name: valueHasChanged
+      expression: |
```
JoelSpeed, line 22: Hmm, perhaps actually, valueHasChanged should be

```
variables.hasNetworkTier && (!variables.hasOldNetworkTier || variables.oldNetworkTierValue != variables.networkTierValue)
```

Then below for the validations, it can simply be
```
!variables.valueHasChanged || variables.networkTierValue in ['Standard', 'Premium']
```

As the valueHasChanged will show true only when we definitely have a network tier

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - name: hasIdleTimeout
+      expression: has(object.metadata.annotations) && 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' in object.metadata.annotations
+    - name: idleTimeoutValue
+      expression: object.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
+    - name: hasOldIdleTimeout
+      expression: has(oldObject.metadata.annotations) && 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' in oldObject.metadata.annotations
+    - name: oldIdleTimeoutValue
+      expression: oldObject.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
+    - name: valueHasChanged
+      expression: oldObject == null || oldObject.metadata == null || oldObject.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("") != object.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
```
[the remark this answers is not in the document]
miyadav: let me commit them and build and validate .

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,32 @@
...
+        variables.hasOldNetworkTier &&
+        variables.hasNetworkTier &&
+        variables.oldNetworkTierValue != variables.networkTierValue
```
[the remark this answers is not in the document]
miyadav: will check this , thanks

pkg/cloud/gcp/assets/validating-admission-policy.yaml
miyadav: This will result in early exit , shall we prefer this ? 
`!variables.hasOldNetworkTier ||
        variables.hasNetworkTier &&
        variables.oldNetworkTierValue != variables.networkTierValue`

miyadav: Validation I think we can update to what you are suggesting .

JoelSpeed: But I think the early exit is fine. If we don't have the network tier value now, there's no point continuing, so we return false, and the `!variables.hasValueChanged` returns true, exiting the later validation, which is appropriate

pkg/cloud/gcp/assets/validating-admission-policy.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - name: hasNetworkTier
+      expression: has(object.metadata.annotations) && 'cloud.google.com/network-tier' in object.metadata.annotations
+    - name: networkTierValue
+      expression: object.metadata.?annotations['cloud.google.com/network-tier'].orValue("")
+    - name: hasOldNetworkTier
+      expression: has(oldObject) && has(oldObject.metadata) && has(oldObject.metadata.annotations) && 'cloud.google.com/network-tier' in oldObject.metadata.annotations
+    - name: oldNetworkTierValue
+      expression: oldObject.metadata.?annotations['cloud.google.com/network-tier'].orValue("")
+    - name: valueHasChanged
+      expression: variables.hasNetworkTier && (!variables.hasOldNetworkTier || variables.oldNetworkTierValue != variables.networkTierValue)
```
JoelSpeed, line 22: This is really clean now, can we make the azure one similar in terms of how it calculates the valueHasChanged value?

miyadav:
@JoelSpeed updated Jira with tests from latest commit. PTAL when time permits if it is good for labels.

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
```diff
@@ -0,0 +1,28 @@
...
+      expression: oldObject != null && oldObject.metadata.?annotations != null && 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' in oldObject.metadata.annotations
+    - name: oldIdleTimeoutValue
+      expression: oldObject.metadata.?annotations['service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout'].orValue("")
+    - name: valueHasChanged
+      expression: variables.hasIdleTimeout && (!variables.hasOldIdleTimeout || variables.oldIdleTimeoutValue != variables.idleTimeoutValue) 
+  validations:
+    - expression: "!variables.hasIdleTimeout || !variables.valueHasChanged || variables.idleTimeoutValue.matches('^[0-9]+$')"
+      message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be an integer"
+    - expression: "!variables.hasIdleTimeout || !variables.valueHasChanged || (int(variables.idleTimeoutValue) >= 4 && int(variables.idleTimeoutValue) <= 100)"
+      message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be no less than 4 and no more than 100"
```
JoelSpeed, line 27: Nit, because of how `valueHasChanged` is formed, this is also equivalent functionally, happy for you to leave as is if you prefer though
```suggestion
    - expression: "!variables.valueHasChanged || variables.idleTimeoutValue.matches('^[0-9]+$')"
      message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be an integer"
    - expression: "!variables.valueHasChanged || (int(variables.idleTimeoutValue) >= 4 && int(variables.idleTimeoutValue) <= 100)"
      message: "value for annotation 'service.beta.kubernetes.io/azure-load-balancer-tcp-idle-timeout' must be no less than 4 and no more than 100"
```

JoelSpeed (commented):
One nit, up to you if you'd like to change or not

/approve

@theobarberbany are you happy with this?

pkg/cloud/azure/assets/validating-admission-service-annotation-policy.yaml
miyadav: Thanks @JoelSpeed happy to keep as is, will wait for @theobarberbany if any inputs.

theobarberbany: I have no strong opinion here, If the extra verbosity makes it easier to follow then leave it in :)

miyadav:
Thanks @JoelSpeed , @theobarberbany  , I don't think these failures are related to our change, can we ignore them .

JoelSpeed:
@miyadav Can you take a look at the failing units? They look as though they could be legitimate issues

PR #382, opened 2025-01-28
OCPCLOUD-2831: Bump k8s 1.32

theobarberbany:


go.mod
```diff
@@ -1,32 +1,32 @@
...
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.31.2
-	k8s.io/apiextensions-apiserver v0.31.2
-	k8s.io/apimachinery v0.31.2
-	k8s.io/client-go v0.31.2
+	k8s.io/api v0.32.1
+	k8s.io/apiextensions-apiserver v0.32.1
+	k8s.io/apimachinery v0.32.1
+	k8s.io/client-go v0.32.1
 	k8s.io/cloud-provider-vsphere v1.31.1
```
JoelSpeed, line 25: Can this be updated?

go.mod
```diff
@@ -1,32 +1,32 @@
...
-	k8s.io/component-base v0.31.2
-	k8s.io/controller-manager v0.31.2
+	k8s.io/component-base v0.32.1
+	k8s.io/controller-manager v0.32.1
 	k8s.io/klog/v2 v2.130.1
-	k8s.io/utils v0.0.0-20240921022957-49e7df575cb6
+	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
 	sigs.k8s.io/cloud-provider-azure v1.30.3
 	sigs.k8s.io/cloud-provider-azure/pkg/azclient v0.0.25
 	sigs.k8s.io/controller-runtime v0.19.1
```
JoelSpeed, line 32: Should we be updating controller-runtime and other packages like cloud-provider-azure while we are here?

Makefile
```diff
@@ -3,7 +3,7 @@
 IMG ?= controller:latest
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.31.1
+ENVTEST_K8S_VERSION = 1.32.0
```
damdo, line 6: Will need envtests bin to be pushed to the bucket. I'm seeing to that.
I plan to push 1.32.1 so I suggest to switch to that here as well.

JoelSpeed:
Vendor is sad, otherwise LGTM

JoelSpeed:
/label acknowledge-critical-fixes-only 

Only a dependency update to the operator, can see from E2Es that the operands are still being deployed and updated, this is low risk

PR #383, opened 2025-01-30
OCPCLOUD-2843: Add AWS Cloud Config Transformer to set Shared config for AWS LB health probes

JoelSpeed:
Until recently, we were carrying a patch in the CCM that changed the LB behaviour to use the OVN kube shared health check endpoints. A PR has now landed upstream that allows this configuration, if you configure the Shared option for the load balancer health checks.

This PR adds an initial config transformer to inject this configuration.

It does not tackle handling service endpoints, which are currently injected by the cluster-config-operator. We have some todos across the code base to fix these up. We now have both AWS and Azure config transformers, we should really get to that, but it's not required for this specific fix.

I have included a test case to make sure service overrides are kept.


JoelSpeed:
Can see this is working, and fortunately the AWS CCM isn't complaining about unknown values, we should be able to get this merged and then try merging the AWS CCM update again

/retest-required

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -81,43 +81,57 @@ func makeInfraStatus(platform configv1.PlatformType) configv1.InfrastructureStat
...
 	return &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{
 		Name:      infraCloudConfName,
 		Namespace: OpenshiftConfigNamespace,
-	}, Data: map[string]string{infraCloudConfKey: defaultAzureConfig}}
+	}, Data: map[string]string{infraCloudConfKey: defaultConfig}}
 }
 
-func makeManagedCloudConfig() *corev1.ConfigMap {
+func makeManagedCloudConfig(platform configv1.PlatformType) *corev1.ConfigMap {
+	defaultConfig := `[Global]
```
theobarberbany, line 99: nit: Is this changing the meaning of default to be an aws config, rather than azure? 
I might be tempted to switch on platformtype here to make things explicit.

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -81,43 +81,57 @@ func makeInfraStatus(platform configv1.PlatformType) configv1.InfrastructureStat
 	}
 }
 
-func makeInfraCloudConfig() *corev1.ConfigMap {
+func makeInfraCloudConfig(platform configv1.PlatformType) *corev1.ConfigMap {
+	defaultConfig := `[Global]
```
theobarberbany, line 85: nit: Is this changing the meaning of default to be an aws config, rather than azure?
I might be tempted to switch on platformtype here to make things explicit.

theobarberbany:
minor non blocking nit otherwise 
/lgtm

pkg/controllers/cloud_config_sync_controller_test.go
JoelSpeed: Originally, all platforms used INI, most of the CCMs are still INI configured, Azure on the other hand, is the exception.

Some support INI or yaml, not AWS, not Azure 🤦

theobarberbany: TIL, and Oh, god 😞

PR #384, opened 2025-02-10
OCPCLOUD-2843: AWS CCM should be configured with the AWS cloud provider config

JoelSpeed:
I recently update the AWS CCM to set the load balancer, cluster scoped health checks to port 10256 and /healthz. It turns out, that even though the tests looked good, this config change was not being consumed by the CCM. We did not even pass the configmap into the AWS CCM.

I've tested this manually and when configuring the below, I can now see that the configmap options are honoured, and the health checks appear as expected.

### DIFF
--- a/pkg/cloud/aws/assets/deployment.yaml
+++ b/pkg/cloud/aws/assets/deployment.yaml
@@ -34,6 +34,7 @@ spec:
             source /etc/kubernetes/apiserver-url.env
           fi
           exec /bin/aws-cloud-controller-manager \
+          --cloud-config=$(CLOUD_CONFIG) \
           --cloud-provider=aws \
           --use-service-account-credentials=true \
           --configure-cloud-routes=false \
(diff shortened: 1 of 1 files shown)


JoelSpeed:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn 10 
/payload-aggregate periodic-ci-openshift-release-master-ci-4.19-e2e-aws-ovn-upgrade 10

sunzhaohua2:
/label qe-approved
Tested with pr and checked healthcheck port is 10256, path is /healthz

JoelSpeed:
/label acknowledge-critical-fixes-only

Payload testing looks good in https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/384#issuecomment-2648207841. I cannot see any indication of load balancer introduced disruption during the upgrades

PR #393, opened 2025-06-27
NO-JIRA: Remove old deployment which is deleted

lance5890:
kind of post-cleanup of https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/157

reduce unused yaml


manifests/0000_26_cloud-controller-manager-operator_10_deployment_tombstone.yaml
```diff
@@ -1,10 +0,0 @@
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: cluster-cloud-controller-manager
-  namespace: openshift-cloud-controller-manager-operator
-  annotations:
-    capability.openshift.io/name: CloudControllerManager
-    include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/single-node-developer: "true"
-    release.openshift.io/delete: "true"
```
lance5890, line 10: cvo will skip the yaml with this annotation

PR #395, opened 2025-07-09
OCPCLOUD-2978: add deny all network policy

sunzhaohua2:
CCM is using hostNetwork: true, so most NetworkPolicy functionality does not apply to it [OCPSTRAT-176](https://issues.redhat.com/browse/OCPSTRAT-176) only need to add one default deny all policy.

```
$ oc get deploy -n openshift-cloud-controller-manager-operator -o yaml | grep hostNetwork  
        hostNetwork: true
$ oc get deploy -n openshift-cloud-controller-manager -o yaml | grep hostNetwork  
        hostNetwork: true
```


JoelSpeed:
/lgtm
/approve

@sunzhaohua2 Has this been tested already to check it doesn't impact the CCM functionality?

PR #399, opened 2025-07-25
OCPBUGS-53300: Adds service for cloud-controller-manager

theobarberbany:
Adds services to expose port 10258 for each provider cloud controller manager.

### DIFF
--- a/pkg/controllers/clusteroperator_controller_test.go
+++ b/pkg/controllers/clusteroperator_controller_test.go
@@ -354,7 +354,8 @@ var _ = Describe("Apply resources should", func() {
 		updated, err := reconciler.applyResources(context.TODO(), resources)
 		Expect(err).ShouldNot(HaveOccurred())
 		Expect(updated).To(BeTrue())
-		// two resources should report successful update, deployment and pdb
+		// three resources should report successful update, deployment, pdb and service
+		Eventually(recorder.Events).Should(Receive(ContainSubstring("Resource was successfully created")))
 		Eventually(recorder.Events).Should(Receive(ContainSubstring("Resource was successfully created")))
 		Eventually(recorder.Events).Should(Receive(ContainSubstring("Resource was successfully created")))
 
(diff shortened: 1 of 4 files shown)


JoelSpeed:
Have you verified that these are actually getting applied? Based on previous experience, I would have expected the unit tests to be failing here, as we have some tests that check which resources are being extracted from the assets dir. I'm concerned that the operator is silently ignoring these new files. Any ideas?

theobarberbany:
Latest e2e shows the service in place & looking good :D 

```yaml
---
apiVersion: v1
items:
- apiVersion: v1
  kind: Service
  metadata:
    creationTimestamp: "2025-07-29T12:18:56Z"
    labels:
      k8s-app: aws-cloud-controller-manager
    name: aws-cloud-controller-manager
    resourceVersion: "2865"
    uid: 8b3d9d7c-7c92-436e-965f-ef79fd6432ac
  spec:
    clusterIP: 172.30.228.207
    clusterIPs:
    - 172.30.228.207
    internalTrafficPolicy: Cluster
    ipFamilies:
    - IPv4
    ipFamilyPolicy: SingleStack
    ports:
    - name: https
      port: 10258
      protocol: TCP
      targetPort: 10258
    selector:
      infrastructure.openshift.io/cloud-controller-manager: AWS
    sessionAffinity: None
    type: ClusterIP
  status:
    loadBalancer: {}
kind: ServiceList
metadata:
  resourceVersion: "267247"
```

pkg/controllers/clusteroperator_controller_test.go
```diff
@@ -354,7 +354,8 @@ var _ = Describe("Apply resources should", func() {
 		updated, err := reconciler.applyResources(context.TODO(), resources)
 		Expect(err).ShouldNot(HaveOccurred())
 		Expect(updated).To(BeTrue())
-		// two resources should report successful update, deployment and pdb
+		// two resources should report successful update, deployment, pdb and service
```
JoelSpeed, line 357:
```suggestion
		// three resources should report successful update, deployment, pdb and service
```

theobarberbany: 🤦 good catch 😂

PR #404, opened 2025-08-05
OCPBUGS-59251: Default cloud.conf if no configmap is found

nrb:
Prior to 4.14, OCP's installer did not provide any cloud configuration in the `Infrastructure` resource. This was ok because nothing relied on it.

However in 4.19, we introduced a hard requirement for the cloud configuration on AWS.

This change allows us to gracefully handle missing information in `Infrastructure` by providing a minimal default `cloud.conf` on AWS.

### DIFF
--- a/pkg/cloud/aws/aws_config_transformer.go
+++ b/pkg/cloud/aws/aws_config_transformer.go
@@ -13,8 +13,11 @@ import (
 	awsconfig "k8s.io/cloud-provider-aws/pkg/providers/v1/config"
 )
 
+const defaultConfig = `[Global]
+`
+
 // CloudConfigTransformer is used to inject OpenShift configuration defaults into the Cloud Provider config
-// for the AWS Cloud Provider.
+// for the AWS Cloud Provider. If an empty source string is provided, a minimal default configuration will be created.
 func CloudConfigTransformer(source string, infra *configv1.Infrastructure, network *configv1.Network) (string, error) {
 	cfg, err := readAWSConfig(source)
 	if err != nil {
@@ -26,13 +29,19 @@ func CloudConfigTransformer(source string, infra *configv1.Infrastructure, netwo
 	return marshalAWSConfig(cfg)
 }
 
+// readAWSConfig will parse a source string into a proper *awsconfig.CloudConfig.
+// If an empty source string is provided, a default configuration will be used.
 func readAWSConfig(source string) (*awsconfig.CloudConfig, error) {
+	cfg := &awsconfig.CloudConfig{}
+
+	// There are cases in which a cloud config was not installed with a cluster, and this is valid.
+	// We should, however, populate the configuration so that it doesn't fail on later versions that require
+	// a cloud.conf.
 	if len(source) == 0 {
-		return nil, fmt.Errorf("empty INI file")
+		source = defaultConfig
 	}
 
 	// Use the same method the AWS CCM uses to load configuration.
-	cfg := &awsconfig.CloudConfig{}
 	if err := gcfg.FatalOnly(gcfg.ReadStringInto(cfg, source)); err != nil {
 		return nil, fmt.Errorf("failed to parse INI file: %w", err)
 	}
--- a/pkg/cloud/aws/aws_config_transformer.go
+++ b/pkg/cloud/aws/aws_config_transformer.go
@@ -13,6 +13,8 @@ import (
 	awsconfig "k8s.io/cloud-provider-aws/pkg/providers/v1/config"
 )
 
+// defaultConfig is a string holding the absolute bare minimum INI string that the AWS CCM needs to start.
+// The value will be further customized for OCP in the CloudConfigTransformer.
 const defaultConfig = `[Global]
 `
 
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -115,12 +115,13 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			Namespace: OpenshiftConfigNamespace,
 		}
 		if err := r.Get(ctx, openshiftUnmanagedCMKey, sourceCM); errors.IsNotFound(err) {
-			klog.Warningf("unmanaged cloud-config is not found, falling back to default cloud config.")
+			klog.Warningf("managed cloud-config is not found, falling back to default cloud config.")
 		} else if err != nil {
-			klog.Errorf("unable to get cloud-config for sync")
+			klog.Errorf("unable to get cloud-config for sync: %v", err)
 			if err := r.setDegradedCondition(ctx); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
 			}
+			return ctrl.Result{}, err
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -183,27 +177,6 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-		return false, fmt.Errorf("platformStatus is required")
-	}
-	switch platformStatus.Type {
-	case configv1.AzurePlatformType,
-		configv1.GCPPlatformType,
-		configv1.VSpherePlatformType,
-		configv1.IBMCloudPlatformType,
-		configv1.PowerVSPlatformType,
-		configv1.OpenStackPlatformType,
-		configv1.NutanixPlatformType:
```
JoelSpeed, line 197: Does removing this mean that we will now sync a metal platform configmap? Or does the sync controller just not run at all on metal?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -218,10 +191,11 @@ func (r *CloudConfigReconciler) prepareSourceConfigMap(source *corev1.ConfigMap,
...
 	}
-	return nil, fmt.Errorf(
-		"key %s specified in infra resource does not found in source configmap %s",
-		infraConfigKey, client.ObjectKeyFromObject(source),
-	)
+
+	// If there was no relevant data in the configmap, make a default entry for it.
+	cloudConfCm.Data[defaultConfigKey] = ""
+
+	return cloudConfCm, nil
```
JoelSpeed, line 198: Not sure about this one. If the user has specced a custom key and that custom key isn't there, we probably still want to error

I assume our fallback detects an that neither the managed or unmanaged confs exist, so creates a blank entry which exists this function on L185? So this change shouldn't be needed?

pkg/controllers/cloud_config_sync_controller.go
nrb: It does run on metal, and removing this caused their tests to fail. I've added it back and simply moved AWS's entry up. This way we're not changing behavior on platforms that relied on it before.

pkg/cloud/aws/aws_config_transformer.go
```diff
@@ -26,13 +29,19 @@ func CloudConfigTransformer(source string, infra *configv1.Infrastructure, netwo
...
+// If an empty source string is provided, a default configuration will be used.
 func readAWSConfig(source string) (*awsconfig.CloudConfig, error) {
+	cfg := &awsconfig.CloudConfig{}
+
+	// There are cases in which a cloud config was not installed with a cluster, and this is valid.
+	// We should, however, populate the configuration so that it doesn't fail on later versions that require
+	// a cloud.conf.
 	if len(source) == 0 {
-		return nil, fmt.Errorf("empty INI file")
+		source = defaultConfig
```
nrb, line 41: I'm torn on this - should we expand the `CloudConfigTransformer` interface to provide the defaults? Or is it enough for each platform to set their minimum requirements during parsing?

JoelSpeed: I suspect each platform can set their minimum requirements? Can they each handle an empty file? Is that something we need to add testing for?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -188,40 +190,51 @@ func (r *CloudConfigReconciler) isCloudConfigSyncNeeded(platformStatus *configv1
 		return false, fmt.Errorf("platformStatus is required")
 	}
 	switch platformStatus.Type {
-	case configv1.AzurePlatformType,
+	case configv1.AWSPlatformType,
```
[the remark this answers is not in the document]
nrb: Dam's correct - AWS is now unconditionally trying to sync. Any platform not specifically called out in the list, like bare metal, will not sync.

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -188,40 +190,51 @@ func (r *CloudConfigReconciler) isCloudConfigSyncNeeded(platformStatus *configv1
...
 	}
-	return nil, fmt.Errorf(
-		"key %s specified in infra resource does not found in source configmap %s",
-		infraConfigKey, client.ObjectKeyFromObject(source),
-	)
+
+	// If there was no data in the configmap and the user didn't specify their own make a default entry for it.
+	cloudConfCm.Data[defaultConfigKey] = ""
+
+	return cloudConfCm, nil
```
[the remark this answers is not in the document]
damdo: Thanks, the move looks clearer to me. TY

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -108,17 +108,19 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	if !managedConfigFound && infra.Spec.CloudConfig.Name != "" {
 		openshiftUnmanagedCMKey := client.ObjectKey{
 			Name:      infra.Spec.CloudConfig.Name,
 			Namespace: OpenshiftConfigNamespace,
 		}
-		if err := r.Get(ctx, openshiftUnmanagedCMKey, sourceCM); err != nil {
+		if err := r.Get(ctx, openshiftUnmanagedCMKey, sourceCM); errors.IsNotFound(err) {
+			klog.Warningf("unmanaged cloud-config is not found, falling back to default cloud config.")
+		} else if err != nil {
 			klog.Errorf("unable to get cloud-config for sync")
```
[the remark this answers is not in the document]
theobarberbany: Question for my understanding: Do we? Doesn't the `return ctrl.Result{}, err` below return it?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -108,17 +108,19 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 			if err := r.setDegradedCondition(ctx); err != nil {
 				return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
 			}
-			return ctrl.Result{}, err
```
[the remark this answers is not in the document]
theobarberbany: I agree with @damdo, we don't want to be defaulting to an empty config if there's a transient error when getting the unmanaged one. 

> So we should keep retrying to get it (likely it is a transient error e.g. network hiccup or similar)

For my own understanding: the pre existing behaviour is a failure state - so we would retry indefinitely until success? Could people be relying on this behaviour?

damdo:
/hold

Feel free to unhold once happy @nrb

sunzhaohua2:
/label qe-approved

Upgrade cluster from 4.13 to a image built with this pr, upgrade succeed.  aws-cloud-controller-manager  pod in running status, 
```
$ oc get pod -n openshift-cloud-controller-manager                                                                                                
NAME                                            READY   STATUS    RESTARTS   AGE
aws-cloud-controller-manager-6847869f8b-2sq9w   1/1     Running   0          28m
aws-cloud-controller-manager-6847869f8b-phc2p   1/1     Running   0          2m36s

$ oc describe po aws-cloud-controller-manager-6847869f8b-2sq9w -n openshift-cloud-controller-manager
Events:
  Type    Reason     Age    From               Message
  ----    ------     ----   ----               -------
  Normal  Scheduled  9m39s  default-scheduler  Successfully assigned openshift-cloud-controller-manager/aws-cloud-controller-manager-6847869f8b-2sq9w to ip-10-0-177-70.us-east-2.compute.internal
  Normal  Pulling    9m40s  kubelet            Pulling image "registry.build10.ci.openshift.org/ci-ln-qbszx3b/stable@sha256:08f547a14d2cea95e5164bcf8d5882a1ef87153d6dc2c919ed147c80231fc488"
  Normal  Pulled     9m37s  kubelet            Successfully pulled image "registry.build10.ci.openshift.org/ci-ln-qbszx3b/stable@sha256:08f547a14d2cea95e5164bcf8d5882a1ef87153d6dc2c919ed147c80231fc488" in 2.454s (2.454s including waiting). Image size: 499957008 bytes.
  Normal  Created    9m37s  kubelet            Created container: cloud-controller-manager
  Normal  Started    9m37s  kubelet            Started container cloud-controller-manager

$ oc get clusterversion                                                                                                                            
NAME      VERSION   AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.20.0-0-2025-08-27-085941-test-ci-ln-qbszx3b-latest    True        True          51m     4.20.0-0-2025-08-27-085941-test-ci-ln-qbszx3b-latest
```

PR #407, opened 2025-08-07
[release-4.17] OCPBUGS-60211: Add Service using common resource templating

theobarberbany:
Adds a service for the cloud-controller-manager on ports 10258 and 10260 using the common resource templating pattern we already have.

PR #411, opened 2025-08-27
[release-4.16] OCPBUGS-60387: Add Service using common resource templating

theobarberbany:
Adds a service for the cloud-controller-manager on ports 10258 and 10260 using the common resource templating pattern we already have.

cherry-pick of https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/407


pkg/cloud/common/resources.go
```diff
@@ -53,3 +57,42 @@ func getPDB(config config.OperatorConfig) (*policyv1.PodDisruptionBudget, error)
...
+func getService(config config.OperatorConfig) *corev1.Service {
+	matchLabels := map[string]string{
+		CloudControllerManagerProviderLabel: config.GetPlatformNameString(),
+	}
+	name := fmt.Sprintf("%s-cloud-controller-manager", strings.ToLower(config.GetPlatformNameString()))
+
+	return &corev1.Service{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "core/v1",
```
chrischdi, line 72:
```suggestion
			APIVersion: "v1",
```

Not sure what we use to create this object, but setting typemeta should maybe not really be required?!

theobarberbany: Yeah, this was to make existing unit tests happy :( We have to do the same for the PDB above, so it's matching the existing pattern. 

I agree though, It's ugly :(

shirmoran:
@huali9 Can you help with verified label please?

PR #416, opened 2025-09-16
OCPBUGS-60385: Add Microsoft.Network/publicIPAddresses/delete permission

RadekManak:
This permission is required for the DeletePublicIP function in the
Azure Cloud Controller Manager to properly clean up orphaned public IP
addresses when LoadBalancer services are deleted.

Code location:
- File: pkg/provider/azure_publicip_repo.go:81
- Function: DeletePublicIP
- API call: NetworkClientFactory.GetPublicIPAddressClient().Delete()


damdo (commented):
/approve

/hold

PR #436, opened 2026-03-03
OCPBUGS-42837: Do not set Degraded=True on transient errors

nrb:
CloudConfigReconciler: gate transient errors behind a 2-minute window

Three related fixes to stop upgrade-time API blips from immediately setting CloudConfigControllerDegraded=True:

1. Infrastructure NotFound now calls setAvailableCondition (nil return) instead of setDegradedCondition, matching the main controller's existing behaviour.

2. Errors are classified as transient (API blips: all Get/Create/Update calls, feature-gate informer not yet synced) or permanent (config problems that won't self-heal: nil platformStatus, unsupported platform, missing user config key, nil FeatureGateAccess, transform failure).

3. handleTransientError() only sets degraded after consecutiveFailureSince has been set for longer than transientDegradedThreshold (2 min); handleDegradeError() sets degraded immediately and returns nil so controller-runtime does not requeue (existing watches re-trigger when the underlying config changes).  clearFailureWindow() is called at every successful reconcile exit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Introduced a ~2-minute transient-failure window so temporary errors no longer cause immediate degradation; persistent failures still mark operators/controllers degraded. Error-handling now yields clearer, more consistent availability/degraded status signals.

* **Tests**
  * Expanded test coverage for transient vs persistent failures, threshold-driven degradation, and updated status-condition assertions to verify the new signaling behavior.

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 	"sync"
@@ -157,25 +156,7 @@ func (fw *failureWindow) handleTerminal(name string, err error, setDegraded func
 func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, retErr error) {
 	klog.V(1).Infof("Syncing cloud-conf ConfigMap")
 
-	// Deferred dispatcher: classifies the returned error and calls the right handler.
-	// Terminal errors (wrapped with reconcile.TerminalError()) degrade immediately without requeue.
-	// Transient errors enter the failure window and only degrade after the threshold.
-	// All nil-error paths clear the failure window.
-	defer func() {
-		if retErr == nil {
-			r.failures.clear()
-			return
-		}
-		if errors.Is(retErr, reconcile.TerminalError(nil)) {
-			result, retErr = r.failures.handleTerminal("CloudConfigReconciler", retErr, func() error {
-				return r.setDegradedCondition(ctx)
-			})
-		} else {
-			result, retErr = r.failures.handleTransient(r.Clock.Now(), 0, transientDegradedThreshold, "CloudConfigReconciler", retErr, func() error {
-				return r.setDegradedCondition(ctx)
-			})
-		}
-	}()
+	defer finalizeReconcile(&r.failures, r.Clock, 0, transientDegradedThreshold, "CloudConfigReconciler", r.failures.clear, degradedSetter(ctx, r.setDegradedCondition), &result, &retErr)
 
 	infra := &configv1.Infrastructure{}
 	if err := r.Get(ctx, client.ObjectKey{Name: infrastructureResourceName}, infra); apierrors.IsNotFound(err) {
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -151,18 +151,18 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		}
 		return ctrl.Result{}, nil
 	} else if err != nil {
-		return ctrl.Result{}, err // transient
+		return ctrl.Result{}, fmt.Errorf("failed to get Infrastructure: %w", err)
 	}
 
 	network := &configv1.Network{}
 	if err := r.Get(ctx, client.ObjectKey{Name: "cluster"}, network); err != nil {
-		return ctrl.Result{}, err // transient
+		return ctrl.Result{}, fmt.Errorf("failed to get cluster Network: %w", err)
 	}
 
 	syncNeeded, err := r.isCloudConfigSyncNeeded(infra.Status.PlatformStatus, infra.Spec.CloudConfig)
 	if err != nil {
 		// nil platformStatus is a terminal misconfiguration.
-		return ctrl.Result{}, reconcile.TerminalError(err)
+		return ctrl.Result{}, reconcile.TerminalError(fmt.Errorf("failed to check cloud config sync requirements: %w", err))
 	}
 	if !syncNeeded {
 		if err := r.setAvailableCondition(ctx); err != nil {
@@ -194,7 +194,7 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	if err != nil {
 		// Unsupported platform won't change without a cluster reconfigure.
 		klog.Errorf("unable to get cloud config transformer function; unsupported platform")
-		return ctrl.Result{}, reconcile.TerminalError(err)
+		return ctrl.Result{}, reconcile.TerminalError(fmt.Errorf("failed to get cloud config transformer: %w", err))
 	}
 
 	platformType := infra.Status.PlatformStatus.Type
@@ -219,7 +219,7 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		} else if apierrors.IsNotFound(err) {
 			klog.Warningf("managed cloud-config is not found, falling back to infrastructure config")
 		} else {
-			return ctrl.Result{}, err // transient
+			return ctrl.Result{}, fmt.Errorf("failed to get managed cloud config: %w", err)
 		}
 	}
 
@@ -252,12 +252,12 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	sourceCM, err = r.prepareSourceConfigMap(sourceCM, infra)
 	if err != nil {
 		// User-supplied key mismatch: terminal until the ConfigMap or Infrastructure changes.
-		return ctrl.Result{}, reconcile.TerminalError(err)
+		return ctrl.Result{}, reconcile.TerminalError(fmt.Errorf("failed to prepare source cloud config: %w", err))
 	}
 
 	// Apply transformer if needed
 	if r.FeatureGateAccess == nil {
-		// Operator misconfiguration at startup: ermanent.
+		// Operator misconfiguration at startup: Permanent.
 		return ctrl.Result{}, reconcile.TerminalError(fmt.Errorf("FeatureGateAccess is not configured"))
 	}
 
@@ -265,7 +265,7 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	if err != nil {
 		// The feature-gate informer may not have synced yet: transient.
 		klog.Errorf("unable to get feature gates: %v", err)
-		return ctrl.Result{}, err // transient
+		return ctrl.Result{}, fmt.Errorf("failed to get feature gates: %w", err)
 	}
 	if cloudConfigTransformerFn != nil {
 		// We ignore stuff in sourceCM.BinaryData. This isn't allowed to
@@ -274,7 +274,7 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		output, err := cloudConfigTransformerFn(sourceCM.Data[defaultConfigKey], infra, network, features)
 		if err != nil {
 			// Platform-specific transform failed on the current config data: terminal.
-			return ctrl.Result{}, reconcile.TerminalError(err)
+			return ctrl.Result{}, reconcile.TerminalError(fmt.Errorf("failed to transform cloud config: %w", err))
 		}
 		sourceCM.Data[defaultConfigKey] = output
 	}
@@ -294,7 +294,7 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	// Sync the transformed config to the target configmap for CCM consumption
 	if err := r.syncCloudConfigData(ctx, sourceCM); err != nil {
 		klog.Errorf("unable to sync cloud config")
-		return ctrl.Result{}, err // transient
+		return ctrl.Result{}, fmt.Errorf("failed to sync cloud config: %w", err)
 	}
 
 	if err := r.setAvailableCondition(ctx); err != nil {
@@ -326,7 +326,7 @@ func (r *CloudConfigReconciler) handleTransientError(ctx context.Context, err er
 	}
 	klog.Warningf("CloudConfigReconciler: transient failure exceeded threshold (%s), setting degraded: %v", elapsed, err)
 	if setErr := r.setDegradedCondition(ctx); setErr != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to set degraded condition: %v", setErr)
+		return ctrl.Result{}, fmt.Errorf("failed to set degraded condition: %w", setErr)
 	}
 	return ctrl.Result{}, err
 }
@@ -338,7 +338,7 @@ func (r *CloudConfigReconciler) handleTransientError(ctx context.Context, err er
 func (r *CloudConfigReconciler) handleTerminalError(ctx context.Context, err error) (ctrl.Result, error) {
 	klog.Errorf("CloudConfigReconciler: terminal error, setting degraded: %v", err)
 	if setErr := r.setDegradedCondition(ctx); setErr != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to set degraded condition: %v", setErr)
+		return ctrl.Result{}, fmt.Errorf("failed to set degraded condition: %w", setErr)
 	}
 	return ctrl.Result{}, nil
 }
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -2,11 +2,14 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
+	"sync"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -14,6 +17,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/api/features"
@@ -30,6 +34,12 @@ const (
 	// Controller conditions for the Cluster Operator resource
 	cloudConfigControllerAvailableCondition = "CloudConfigControllerAvailable"
 	cloudConfigControllerDegradedCondition  = "CloudConfigControllerDegraded"
+
+	// transientDegradedThreshold is how long transient errors must persist before
+	// the controller sets Degraded=True. This prevents brief
+	// API server blips during upgrades from immediately degrading the operator.
+	// Applies to both CloudConfigController and TrustedCAController.
+	transientDegradedThreshold = 2 * time.Minute
 )
 
 // shouldManageManagedConfigMap returns true if CCCMO should manage the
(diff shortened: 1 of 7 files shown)


nrb:
/retitle OCPBUGS-42837: Do not set Degraded=True on transient errors

RadekManak (commented):
one nit, lgtm.

nrb:
Updated the test assertions so they're not just `To(BeTrue())` and consolidated everything into 1 commit.

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -29,40 +33,93 @@ const (
...
 	}
 
 	syncNeeded, err := r.isCloudConfigSyncNeeded(infra.Status.PlatformStatus, infra.Spec.CloudConfig)
 	if err != nil {
-		if err := r.setDegradedCondition(ctx); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
-		}
-		return ctrl.Result{}, err
+		// nil platformStatus is a terminal misconfiguration.
+		return ctrl.Result{}, reconcile.TerminalError(err)
```
JoelSpeed: We are losing context here, because we are no longer doing the `fmt.Errorf`. We should still do the `fmt` but use `%w` and provide additional context as we did before

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -116,41 +167,29 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	}
 
-	// Check if FeatureGateAccess is configured
 	if r.FeatureGateAccess == nil {
-		klog.Errorf("FeatureGateAccess is not configured")
-		if err := r.setDegradedCondition(ctx); err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to set conditions for cloud config controller: %v", err)
-		}
-		return ctrl.Result{}, fmt.Errorf("FeatureGateAccess is not configured")
+		// Operator misconfiguration at startup: ermanent.
```
JoelSpeed:
```suggestion
		// Operator misconfiguration at startup: Permanent.
```

JoelSpeed (commented):
Can you please go through all of the returns you've changed and re-add the `fmt.Errorf` if it had one before, seems a shame to lose the context that we had in place for why/where the errors were coming from? (Unless there's some good reason I'm missing for their removal)

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -29,40 +33,93 @@ const (
 	// Controller conditions for the Cluster Operator resource
 	cloudConfigControllerAvailableCondition = "CloudConfigControllerAvailable"
 	cloudConfigControllerDegradedCondition  = "CloudConfigControllerDegraded"
+
+	// transientDegradedThreshold is how long transient errors must persist before
+	// the controller sets Degraded=True. This prevents brief
+	// API server blips during upgrades from immediately degrading the operator.
+	// Applies to both CloudConfigController and TrustedCAController.
+	transientDegradedThreshold = 2 * time.Minute
```
damdo: How did we come up with this value? Have we tested it in upgrades and checked if it is sufficient?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -206,6 +236,45 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+// relevant resource will re-trigger reconciliation when the problem is fixed.
+// Called only from the deferred dispatcher in Reconcile.
+func (r *CloudConfigReconciler) handleTerminalError(ctx context.Context, err error) (ctrl.Result, error) {
+	klog.Errorf("CloudConfigReconciler: terminal error, setting degraded: %v", err)
+	if setErr := r.setDegradedCondition(ctx); setErr != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to set degraded condition: %v", setErr)
+	}
+	return ctrl.Result{}, nil
+}
+
```
damdo: These seem quite similar across the two controllers. Is it worth making them generic enough so we can deduplicate them?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -29,40 +33,93 @@ const (
...
+		if retErr == nil {
+			r.clearFailureWindow()
+			return
+		}
+		if errors.Is(retErr, reconcile.TerminalError(nil)) {
+			result, retErr = r.handleTerminalError(ctx, retErr)
+		} else {
+			result, retErr = r.handleTransientError(ctx, retErr)
+		}
+	}()
```
damdo: Is this generic enough common logic that it can be defined in a package?

damdo (commented):
Thanks Nolan, a couple of Qs

pkg/controllers/cloud_config_sync_controller.go
nrb: I think that would be worthwhile; should we maybe defer that to a later PR, or do you want to do it in this one?

damdo: I'd say we can do this here. So it is easier to backport too :)

pkg/controllers/cloud_config_sync_controller.go
nrb: I haven't tested this across upgrades, but I will.

This was taken from a couple other repos doing similar work on their readiness work. Will run a few upgrades and see if the window causes issues and needs to be adjusted

damdo: TY!

damdo: Did we end up trying the upgrades? Is the value solid?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -78,34 +88,117 @@ type CloudConfigReconciler struct {
...
+// handleTerminal degrades immediately and returns nil so controller-runtime does not requeue.
+// An existing watch on the relevant resource will re-trigger reconciliation when fixed.
+// name labels log messages.
+func (fw *failureWindow) handleTerminal(name string, err error, setDegraded func() error) (ctrl.Result, error) {
+	klog.Errorf("%s: terminal error, setting degraded: %v", name, err)
+	if setErr := setDegraded(); setErr != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to set degraded condition: %w", setErr)
+	}
+	return ctrl.Result{}, nil
+}
```
damdo: I think we should all put this in a separate file as it is common code across controllers

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -30,6 +34,12 @@ const (
 	// Controller conditions for the Cluster Operator resource
 	cloudConfigControllerAvailableCondition = "CloudConfigControllerAvailable"
 	cloudConfigControllerDegradedCondition  = "CloudConfigControllerDegraded"
+
+	// transientDegradedThreshold is how long transient errors must persist before
+	// the controller sets Degraded=True. This prevents brief
+	// API server blips during upgrades from immediately degrading the operator.
+	// Applies to both CloudConfigController and TrustedCAController.
```
damdo: I'd avoid mentioning to what it applies, as this can go stale. Also I'd put this in a separate file like the comment below.

pkg/controllers/trusted_ca_bundle_controller.go
```diff
@@ -43,16 +44,44 @@ type TrustedCABundleReconciler struct {
...
+func (r *TrustedCABundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, retErr error) {
 	klog.V(1).Infof("%s emitted event, syncing %s ConfigMap", req, trustedCAConfigMapName)
 
+	// partialRun is set to true on the early-exit path where the event is for
+	// an unrelated ConfigMap. That path returns available=true but should NOT
+	// reset an ongoing transient failure window from a previous full reconcile.
+	partialRun := false
+
+	// Deferred dispatcher: classifies the returned error and calls the right handler.
+	// Permanent errors (wrapped with terminal()) degrade immediately without requeue.
```
damdo:
```suggestion
	// Terminal errors (wrapped with terminal()) degrade immediately without requeue.
```

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -78,34 +88,117 @@ type CloudConfigReconciler struct {
...
+		return ctrl.Result{}, fmt.Errorf("failed to set degraded condition: %w", setErr)
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, retErr error) {
 	klog.V(1).Infof("Syncing cloud-conf ConfigMap")
 
+	// Deferred dispatcher: classifies the returned error and calls the right handler.
+	// Permanent errors (wrapped with reconcile.TerminalError()) degrade immediately without requeue.
```
damdo:
```suggestion
	// Terminal errors (wrapped with reconcile.TerminalError()) degrade immediately without requeue.
```

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -78,34 +88,117 @@ type CloudConfigReconciler struct {
...
+		}
+		if errors.Is(retErr, reconcile.TerminalError(nil)) {
+			result, retErr = r.failures.handleTerminal("CloudConfigReconciler", retErr, func() error {
+				return r.setDegradedCondition(ctx)
+			})
+		} else {
+			result, retErr = r.failures.handleTransient(r.Clock.Now(), 0, transientDegradedThreshold, "CloudConfigReconciler", retErr, func() error {
+				return r.setDegradedCondition(ctx)
+			})
+		}
```
damdo: Could these 3 anonymous defer funcs be a named func in a separate file that take in args like:
- r (so we have access to)
  - `setDegradedCondition()`
  - controller/reconciler name
- a degradedThreshold
- a call back func to clear failures

?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -78,34 +88,117 @@ type CloudConfigReconciler struct {
...
+	elapsed, started := fw.observe(now, staleAfter)
+	if started {
+		klog.V(4).Infof("%s: transient failure started (%v), will degrade after %s", name, err, threshold)
+		return ctrl.Result{}, err
+	}
+	if elapsed < threshold {
+		klog.V(4).Infof("%s: transient failure ongoing for %s (threshold %s): %v", name, elapsed, threshold, err)
+		return ctrl.Result{}, err
+	}
+	klog.Warningf("%s: transient failure exceeded threshold (%s), setting degraded: %v", name, elapsed, err)
```
damdo:
```suggestion
	klog.Warningf("%s: transient failure exceeded threshold (%s), setting degraded condition for controller: %v", name, elapsed, err)
```

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -78,34 +88,117 @@ type CloudConfigReconciler struct {
...
+	defer fw.mu.Unlock()
+	stale := staleAfter > 0 && fw.lastTransientFailureAt != nil && now.Sub(*fw.lastTransientFailureAt) > staleAfter
+	if fw.consecutiveFailureSince == nil || stale {
+		fw.consecutiveFailureSince = &now
+		fw.lastTransientFailureAt = &now
+		return 0, true
+	}
+	fw.lastTransientFailureAt = &now
+	return now.Sub(*fw.consecutiveFailureSince), false
+}
```
damdo: Why do we need stale-window?
Is the stale-window error independent? Or do we clear it if two different errors in the same controller come one after the other?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -48,6 +51,13 @@ const (
 
 	// Condition type for Cloud Controller ownership
 	cloudControllerOwnershipCondition = "CloudControllerOwner"
+
+	// aggregatedTransientDegradedThreshold is how long transient errors must persist before
+	// the controller sets Degraded=True.
+	// This prevents brief API server blips during upgrades from immediately degrading the operator.
+	// Applies to top-level operator, and is longer in order
+	// to accomodate changes in the lower-level operators.
+	aggregatedTransientDegradedThreshold = 2*time.Minute + (30 * time.Second)
```
damdo, line 59: Did we validate this is a solid choice in terms of value chosen for time ?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -66,78 +77,118 @@ type CloudOperatorReconciler struct {
...
+		if retErr == nil {
+			r.clearFailureWindow()
+			return
+		}
+		if errors.Is(retErr, reconcile.TerminalError(nil)) {
+			result, retErr = r.handleDegradeError(ctx, conditionOverrides, retErr)
+		} else {
+			result, retErr = r.handleTransientError(ctx, conditionOverrides, retErr)
+		}
+	}()
```
damdo: Could be a shared func

pkg/controllers/clusteroperator_controller.go
```diff
@@ -66,78 +77,118 @@ type CloudOperatorReconciler struct {
...
+		return ctrl.Result{}, fmt.Errorf("failed to get Infrastructure: %w", err)
 	}
 
+	// Known limitation: when provisioningAllowed internally calls setStatusDegraded
+	// (e.g. a sub-controller has Degraded=True, or IsCloudProviderExternal errors),
+	// it returns a non-nil error. Reconcile passes that error to handleTransientError,
+	// which starts the 2m30s window. After the threshold, handleTransientError calls
+	// setStatusDegraded again — redundant but harmless, since status is already degraded.
+	// This is a consequence of keeping status-setting inside provisioningAllowed rather
+	// than pushing it into Reconcile.
```
damdo:
```suggestion
	// Known limitation: when provisioningAllowed internally calls setStatusDegraded
	// (e.g. a sub-controller has Degraded=True, or IsCloudProviderExternal errors),
	// it returns a non-nil error. Reconcile passes that error to handleTransientError,
	// which starts the transientDegradedThreshold window. After the threshold, handleTransientError calls
	// setStatusDegraded again — redundant but harmless, since status is already degraded.
	// This is a consequence of keeping status-setting inside provisioningAllowed rather
	// than pushing it into Reconcile.
```

pkg/controllers/clusteroperator_controller.go
```diff
@@ -66,78 +77,118 @@ type CloudOperatorReconciler struct {
...
+// handleDegradeError sets OperatorDegraded=True immediately and returns nil so
+// controller-runtime does NOT requeue. Existing watches on Infrastructure,
+// ConfigMaps, and Secrets will re-trigger reconciliation when the problem is fixed.
+// Called only from the deferred dispatcher in Reconcile.
+func (r *CloudOperatorReconciler) handleDegradeError(ctx context.Context, conditionOverrides []configv1.ClusterOperatorStatusCondition, err error) (ctrl.Result, error) {
+	klog.Errorf("CloudOperatorReconciler: persistent error, setting degraded: %v", err)
+	if setErr := r.setStatusDegraded(ctx, err, conditionOverrides); setErr != nil {
+		return ctrl.Result{}, fmt.Errorf("error syncing ClusterOperatorStatus: %w", setErr)
+	}
+	return ctrl.Result{}, nil // do not requeue; a watch event will re-trigger
```
damdo: Is this a common pattern? Why are we calling it degradeError here instead of terminalError as for the others?

pkg/controllers/clusteroperator_controller.go
nrb: This is what we're using on the KAS Operator, too - https://github.com/openshift/cluster-kube-apiserver-operator/pull/2052

The bare metal case uses [2 hours](https://github.com/openshift/library-go/pull/2081/changes#diff-3e6eecdb37386d21a87863b978e3dcf05ef2a0ecce8efd37822a21d02bfb02e4R25)

I think 2 minutes is reasonable for thing to settle down in the cluster.

pkg/controllers/cloud_config_sync_controller.go
nrb: The stale window is because the [errors aren't re-queued](https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/436#discussion_r2895053985)

It is independent of errors, and only clears based on a successful reconcile. If different errors happen consecutively it doesn't get reset.

nrb:
Succesful test runs via clusterbot:

- https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2057115930252546048
- https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2057159377864888320
- https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2057323068899987456
- https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2057397394395369472

pkg/controllers/cloud_config_sync_controller.go
theobarberbany: +1 - follow up here? @nrb

pkg/controllers/clusteroperator_controller.go
theobarberbany: agree here - but non blocking nit

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -78,34 +86,99 @@ type CloudConfigReconciler struct {
 	ClusterOperatorStatusClient
 	Scheme            *runtime.Scheme
 	FeatureGateAccess featuregates.FeatureGateAccess
+	failures          failureWindow
+}
+
+// failureWindow tracks consecutive transient failures. All methods are safe for concurrent use.
+type failureWindow struct {
```
theobarberbany: this is used across all controllers - feels like a library we import may make more sense? or just another file?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -66,78 +77,118 @@ type CloudOperatorReconciler struct {
...
+
+func (r *CloudOperatorReconciler) clearFailureWindow() {
+	r.failures.clear()
+}
+
+// handleTransientError records the start of a failure window and degrades the
+// operator only after aggregatedTransientDegradedThreshold has elapsed. It always returns
+// a non-nil error so controller-runtime requeues with exponential backoff.
+// Called only from the deferred dispatcher in Reconcile.
+func (r *CloudOperatorReconciler) handleTransientError(ctx context.Context, conditionOverrides []configv1.ClusterOperatorStatusCondition, err error) (ctrl.Result, error) {
```
theobarberbany: This is a little confusing for me - as its a duplicate of fw.handleTransient but with a slightly different signature? 

Can we unify these - remove the duplications?

pkg/controllers/clusteroperator_controller.go
```diff
@@ -66,78 +77,118 @@ type CloudOperatorReconciler struct {
...
+	if started {
+		klog.V(4).Infof("CloudOperatorReconciler: transient failure started (%v), will degrade after %s", err, aggregatedTransientDegradedThreshold)
 		return ctrl.Result{}, err
 	}
+	if elapsed < aggregatedTransientDegradedThreshold {
+		klog.V(4).Infof("CloudOperatorReconciler: transient failure ongoing for %s (threshold %s): %v", elapsed, aggregatedTransientDegradedThreshold, err)
+		return ctrl.Result{}, err
+	}
+	klog.Warningf("CloudOperatorReconciler: transient failure exceeded threshold (%s), setting degraded: %v", elapsed, err)
+	if setErr := r.setStatusDegraded(ctx, err, conditionOverrides); setErr != nil {
```
theobarberbany: we could pass a closure to the setDegraded() function  on fw.handleTransient()

and avoid this double duplication?

same below

pkg/controllers/cloud_config_sync_controller.go
theobarberbany: missed one ?

theobarberbany: missed?

theobarberbany: also missed? been a while since ive looked at feature gates? are transient errors possible?

pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -229,17 +303,14 @@ func (r *CloudConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			if err := r.setDegradedCondition(ctx); err != nil {
```
theobarberbany: missed?

pkg/controllers/trusted_ca_bundle_controller.go
```diff
@@ -43,16 +43,28 @@ type TrustedCABundleReconciler struct {
...
-func (r *TrustedCABundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *TrustedCABundleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, retErr error) {
 	klog.V(1).Infof("%s emitted event, syncing %s ConfigMap", req, trustedCAConfigMapName)
 
+	// partialRun is set to true on the early-exit path where the event is for
+	// an unrelated ConfigMap. That path returns available=true but should NOT
+	// reset an ongoing transient failure window from a previous full reconcile.
+	partialRun := false
+
+	defer finalizeReconcile(&r.failures, r.Clock, transientDegradedThreshold, transientDegradedThreshold, "TrustedCABundleReconciler", func() {
```
theobarberbany: we use 0 in the other calls? here we repeat `transientDegradedThreshold` twice?

pkg/controllers/trusted_ca_bundle_controller.go
```diff
@@ -43,16 +43,28 @@ type TrustedCABundleReconciler struct {
...
+		if !partialRun {
```
theobarberbany, line 64: worth testing partialRun functionality?

pkg/controllers/trusted_ca_bundle_controller.go
nrb: Captured the values as constants now so they're less of a mystery at the call sites.

pkg/controllers/trusted_ca_bundle_controller.go
nrb: Implemented

theobarberbany:
/verified by new units and an integration to be added in #467

theobarberbany:
*yay*

PR #442, opened 2026-04-08
SPLAT-2651: Added support to manage kube-cloud-config for vSphere in openshift-config-managed

vr4manta:
[SPLAT-2651](https://redhat.atlassian.net/browse/SPLAT-2651)

### Changes

Enhanced the cluster-cloud-controller-manager-operator (CCCMO) to manage the openshift-config-managed/kube-cloud-config ConfigMap for vSphere platforms,    
  migrating ownership from the Cluster Config Operator (CCO) to CCCMO. This enables CCCMO to:
  1. Convert INI-format cloud configs to YAML                                                                                                                 
  2. Apply transformations to add Infrastructure-derived values (vCenters, labels, networking)                                                                
  3. Maintain a single source of truth for cloud configuration                                
  4. Gate functionality behind the VSphereMultiVCenterDay2 feature gate                                                                                       
             
  Tests covering:                                                                                                                                             
  - vSphere with feature gate enabled → returns true
  - vSphere with feature gate disabled → returns false                                                                                                        
  - vSphere with nil feature gates → returns false    
  - Other platforms (AWS, Azure, GCP) → returns false                                                                        
  - Creates ConfigMap if doesn't exist                                                                                                                        
  - Updates ConfigMap when data differs                                                                                                                       
  - Skips update when data is identical (equality check)
  - Error handling: nil source, missing data, missing required key                                                                                            
                      
  Enhanced RBAC for openshift-config-managed namespace - Lines 284-295                                                                                        
                  
  Added verbs to existing Role:                                                                                                                               
  verbs:          
    - get                                                                                                                                                     
    - list                                                                                                                                                    
    - watch                                                                                                                                                   
    - create    # NEW: For initial ConfigMap creation (migration from CCO)                                                                                    
    - update    # NEW: For ongoing updates                                                                                                                    
    - patch     # NEW: For ongoing updates                                                                                                                    
                                                                                                                                                              
  Purpose: Allow CCCMO service account to manage ConfigMaps in openshift-config-managed namespace                                                             
                                                                                                        
  Feature Gate Integration                                                                                                                                    
                                                                                                                                                              
  Feature Gate: VSphereMultiVCenterDay2                                                                                                                       
                                                                                                                                                              
  - Type: configv1.FeatureGateName (typed constant)                                                                                                           
  - Constant: features.FeatureGateVSphereMultiVCenterDay2
  - Purpose: Gates the managed ConfigMap functionality                                                                                                        
  - Pattern: Uses nil-safe checking via isFeatureGateEnabled()                                                                                                
                      
  Before This Enhancement:                                                                                                                                    
                  
  - vSphere cloud config existed in openshift-config (user-provided)                                                                                          
  - CCO might manage openshift-config-managed/kube-cloud-config
  - CCCMO only synced to openshift-cloud-controller-manager/cloud-conf                                                                                        
  - No automatic INI→YAML conversion                                                                                                                          
  - No Infrastructure-derived value injection                                                                                                                 
                                                                                                                                                              
  After This Enhancement:                                                                                                                                     
                                                                                                                                                              
  - CCCMO now manages openshift-config-managed/kube-cloud-config for vSphere (when feature gate enabled)                                                      
  - Automatic INI→YAML conversion using existing ReadConfig/MarshalConfig
  - CloudConfigTransformer adds Infrastructure-derived values                                                                                                 
  - Creates managed ConfigMap if it doesn't exist (migration scenario)                                                                                        
  - Updates managed ConfigMap only when content changes (equality check)                                                                                      
  - Target ConfigMap always updated (no equality check for CCM consumption)                                                                                   
                                                                                                                                                              
  Migration Path from CCO to CCCMO                                                                                                                            
                                                                                                                                                              
  1. Feature gate disabled: CCO manages openshift-config-managed/kube-cloud-config                                                                            
  2. Feature gate enabled: CCCMO takes over:
    - Creates ConfigMap if CCO never created it                                                                                                               
    - Updates ConfigMap with transformed content                                                                                                              
    - Becomes single source of truth                                                                                                                          
  3. Future platforms: Design allows easy addition of AWS/Azure/etc. by updating shouldManageManagedConfigMap()                                               
                        

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

## Release Notes

* **Bug Fixes & Improvements**
  * Cloud controller manager operator now has expanded permissions to manage ConfigMaps in the `openshift-config-managed` namespace, including create, update, and patch operations.
  * Cloud config synchronization enhanced with improved handling for managed platform configurations.

* **Dependencies**
  * Go module dependencies updated for vSphere cloud provider support.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


pkg/controllers/cloud_config_sync_controller.go
```diff
@@ -31,6 +32,36 @@ const (
...
 
+// isFeatureGateEnabled checks if a feature gate is enabled.
+// Returns false if the feature gate is nil.
+// This provides a nil-safe way to check feature gates.
+func isFeatureGateEnabled(featureGates featuregates.FeatureGate, featureName configv1.FeatureGateName) bool {
+	if featureGates == nil {
+		return false
+	}
+	return featureGates.Enabled(featureName)
+}
```
JoelSpeed: Do we really need a nil safe way to check? Shouldn't the `featuregates.FeatureGate` always be initialised? Could this lead to a programatic error with bad behaviour because we missed populating the gate somewhere?

pkg/controllers/cloud_config_sync_controller_test.go
```diff
@@ -596,3 +632,206 @@ var _ = Describe("Cloud config sync reconciler", func() {
...
+
+		allCMs := &corev1.ConfigMapList{}
+		Expect(cl.List(ctx, allCMs)).To(Succeed())
+		for _, cm := range allCMs.Items {
+			Expect(cl.Delete(ctx, cm.DeepCopy(), deleteOptions)).To(Succeed())
+			Eventually(func() error {
+				return cl.Get(ctx, client.ObjectKeyFromObject(cm.DeepCopy()), &corev1.ConfigMap{})
+			}).Should(MatchError(apierrors.IsNotFound, "IsNotFound"))
+		}
+	})
```
JoelSpeed, line 652: Pretty sure we have some test utils for this that we use in CAPI, I wonder if they could also be used here 🤔 

Something like https://github.com/openshift/cluster-api-actuator-pkg/blob/master/testutils/cleanup.go

pkg/controllers/cloud_config_sync_controller.go
vr4manta: I can remove this.  Its been encouraged by our AI friends to do this.  I'll remove and make sure code rabbit is happy.

pkg/controllers/cloud_config_sync_controller_test.go
vr4manta: nice, i can update changes to use that.

go.mod
```diff
@@ -6,13 +6,16 @@ go 1.25.0
 // Upstream release 1.35.0 vendors k8s 1.36-alpha. This is a patched release to vendor k8s 1.35.1
 replace k8s.io/cloud-provider-vsphere => github.com/openshift-cloud-team/cloud-provider-vsphere v1.19.1-0.20260317135518-758abc9d59a5
 
+// Temporary while testing: go get github.com/vr4manta/library-go@SPLAT-2651
+replace github.com/openshift/library-go => github.com/vr4manta/library-go v0.0.0-20260423150140-79a3161b14ce
+
```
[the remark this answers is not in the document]
vr4manta: Yep, this is marked temporary in the comment above it.  We do this while waiting for changes to be merged into target project.

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -7,7 +7,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"k8s.io/utils/net"
 
-	ccmConfig "github.com/openshift/cluster-cloud-controller-manager-operator/pkg/cloud/vsphere/vsphere_cloud_config"
+	ccmConfig "github.com/openshift/library-go/pkg/cloudprovider/vsphere"
```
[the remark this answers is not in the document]
vr4manta: Correct, this is in the replacement PR.  It is now merged and I'll be updating shortly

pkg/controllers/cloud_config_sync_controller_test.go
vr4manta: I tried to but it seems to cause some errors.  I then got claude to take a look and I was told:

```
● Based on my analysis, I can see that in the "vSphere managed config sync" Describe block, the tests only create ConfigMaps in the                           
  OpenshiftManagedConfigNamespace namespace (lines 689, 700, 733).
                                                                                                                                                              
  However, there's an issue with using testutils.CleanupResources() directly: it will try to delete the namespace after cleanup (if the namespace parameter is
   not empty), which you don't want because OpenshiftManagedConfigNamespace is a shared namespace created in BeforeSuite.                                     
                                                                                                                                                              
  Looking at the CleanupResources function signature:                                                                                                         
  func CleanupResources(g gomega.Gomega, ctx context.Context, cfg *rest.Config, k8sClient client.Client, namespace string, objs ...client.Object)
                                                                                                                                                              
  After cleaning up resources, it calls removeNamespace() if namespace is not empty, which would delete the OpenshiftManagedConfigNamespace.                  
                                                                                                                                                              
  Recommended Solution                                                                                                                                        
                                                                                                                                                              
  The best approach is to keep the for loop but add a namespace filter to only clean up ConfigMaps in the specific namespace where this test creates them
```

pkg/cloud/vsphere/vsphere_config_transformer.go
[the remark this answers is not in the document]
vr4manta: can you verify this is still an issue?

go.mod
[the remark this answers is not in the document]
vr4manta: can you verify this is still an issue?

pkg/cloud/vsphere/vsphere_config_transformer.go
```diff
@@ -62,13 +66,34 @@ func CloudConfigTransformer(source string, infra *configv1.Infrastructure, netwo
...
+	// Only initialize Nodes if there's at least one non-empty field to set
+	// This prevents empty "nodes: {}" from appearing in the marshaled YAML
+	if nodeNetworking.External.Network == "" &&
+		externalNetworkSubnetCIDR == "" &&
+		excludeExternalNetworkSubnetCIDR == "" &&
+		nodeNetworking.Internal.Network == "" &&
+		internalNetworkSubnetCIDR == "" &&
+		excludeInternalNetworkSubnetCIDR == "" {
+		return
+	}
```
JoelSpeed: Would adding `omitzero` to the struct mean we don't have to do this?

JoelSpeed (commented):
Just one comment, otherwise LGTM

pkg/cloud/vsphere/vsphere_config_transformer.go
vr4manta: hmm, I'll try that.  It currently has omitempty.  That will require a change to the library-go, but shouldn't be bad.

vr4manta: i tried omit zero but didn't seem to work. 

According to Claude:
```
● Looking at your code, I can see the issue. The omitzero tag you added in line 143 of yaml.go isn't actually supported by the sigs.k8s.io/yaml package you're
   using. That package delegates to Go's standard encoding/json, which only supports omitempty, not omitzero.
```

This seems to be accurate: https://github.com/openshift/cluster-cloud-controller-manager-operator/blob/0776133b8bff2ac9c8a7ab4f9d38bb13c358f785/vendor/sigs.k8s.io/yaml/fields.go#L69-L79

Whats interesting is they do have omitzero in the json package.  /shrug

JoelSpeed:
> Looking at your code, I can see the issue. The omitzero tag you added in line 143 of yaml.go isn't actually supported by the sigs.k8s.io/yaml package you're
   using. That package delegates to Go's standard encoding/json, which only supports omitempty, not omitzero.

So claude is clearly wrong here. `omitzero` is definitely supported by the standard json library, and has been since Go 1.24.

JoelSpeed:
@vr4manta I've created a [go playground](https://go.dev/play/p/SP9qXd3dFvf) which demonstrates that it is supported

vr4manta:
@JoelSpeed I made a PR in library-go to change the behavior of the vsphere cloud config struct.  Now with the omitzero it should work.  Key was to change the field to no longer be a pointer.  That was my issue.  Thanks for the suggestion.

PR #458, opened 2026-04-30
TRT-2652: Allow info and list tests commands to work without KUBECONFIG

petr-muller:
## Summary

Split `initFrameworkForTests()` into two functions:

- `initFrameworkDefaults()` — sets cluster-independent framework defaults (provider, flags, OS distro). Runs at startup, no KUBECONFIG needed.
- `initFrameworkCluster()` — loads kubeconfig, sets host, calls `AfterReadingAllFlags`. Deferred to `AddBeforeAll`, only runs when tests are actually executed.

This enables workflows that need to discover or list tests from ccm-aws-tests without a live cluster connection, such as the `openshift-tests list all-tests` command being added in openshift/origin#31105.

## Related

- openshift/origin#31105 — adds `openshift-tests list all-tests` command, which calls `info` and `list tests` on all payload extension binaries without cluster access

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Test framework init split into cluster-independent startup and deferred cluster-dependent setup.
  * Test discovery and non-cluster commands (e.g., info/list) run without cluster credentials, reducing startup noise.
  * Startup redirects framework output to stderr to keep stdout clean for command output.
  * Temporary placeholder host used to avoid warnings when no kubeconfig is present.
  * AWS region from environment variables continues to be applied.

### DIFF
--- a/openshift-tests/ccm-aws-tests/main.go
+++ b/openshift-tests/ccm-aws-tests/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	kclientset "k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/kubernetes/test/e2e/framework"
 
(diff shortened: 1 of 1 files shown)


petr-muller:
/close

End result too fiddly to be worth it, closing at least for now. [TRT-2652](https://redhat.atlassian.net/browse/TRT-2652) may be picked up in the future if needed.

PR #485, opened 2026-07-08
OCPBUGS-94183, OCPBUGS-94184: [release-4.22] Add transient error handling across all three controllers, fix progressing=true

theobarberbany:
Summary

- All three reconcilers (CloudConfigReconciler, TrustedCABundleReconciler, CloudOperatorReconciler) now distinguish transient errors (API blips) from permanent errors (corrupt config, unsupported platform). Transient errors are silently requeued and only set Degraded=True after a threshold (2m for sub-controllers, 2m30s for the main controller). Permanent errors degrade immediately without requeue.
- Failure handling is consolidated into a shared finalizeReconcile() deferred function in reconcile_dispatch.go, using controller-runtime's TerminalError to classify errors.
- Fix Progressing=True being immediately overwritten by Available — sync() now returns (bool, error) so Reconcile skips setStatusAvailable when resources were just updated.

This depends on https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/484 to avoid the race that required the revert previously.

### DIFF
--- a/pkg/controllers/cloud_config_sync_controller.go
+++ b/pkg/controllers/cloud_config_sync_controller.go
@@ -6,7 +6,7 @@ import (
 	"reflect"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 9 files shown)


nrb (approved):
/approve

PR #497, opened 2026-07-29
OCPBUGS-100052: Created new job to update vSphere nodes to have vsphere label

vr4manta:
OCPBUGS-100052

### Changes
- Created new job that runs during install to update existing nodes to have new label `node.openshift.io/platform-type=vsphere`

### Notes
Idea is to have this only be attempted to run once and only on first upgrade with this change.  Following upgrades should not rerun this job.

## Summary by CodeRabbit

* **New Features**
  * Added a one-time job that automatically applies the vSphere platform label to eligible cluster nodes.
  * The job runs only for vSphere environments with mixed-node support enabled.
  * Added deployment configuration with appropriate security, resource, timeout, and restart settings.

* **Bug Fixes**
  * Ensures eligible vSphere nodes missing the platform label are identified and updated without modifying other node types or existing labels.

* **Tests**
  * Added coverage for supported, unsupported, disabled-feature, and already-labeled scenarios.

### DIFF
--- a/cmd/node-label-sync-job/main.go
+++ b/cmd/node-label-sync-job/main.go
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Command vsphere-node-label-sync-job runs to completion once, backfilling the
+// Command node-label-sync-job runs to completion once, backfilling the
 // node.openshift.io/platform-type=vsphere label onto vSphere nodes that are missing it. It is
 // installed as a Kubernetes Job by the CVO (see
-// manifests/0000_26_cloud-controller-manager-operator_45_job-vsphere-node-label-sync.yaml) rather
+// manifests/0000_90_cloud-controller-manager-operator_00_job.yaml) rather
 // than run as an ongoing in-process controller.
 package main
 
@@ -56,7 +56,7 @@ var (
 )
 
 const (
-	recorderName           = "cloud-controller-manager-operator-vsphere-node-label-sync-job"
+	recorderName           = "cloud-controller-manager-operator-node-label-sync-job"
 	missingVersion         = "0.0.1-snapshot"
 	managedNamespace       = controllers.DefaultManagedNamespace
 	featureGateWaitTimeout = 1 * time.Minute
--- a/cmd/node-label-sync-job/main.go
+++ b/cmd/node-label-sync-job/main.go
@@ -62,7 +62,7 @@ const (
 	managedNamespace       = controllers.DefaultManagedNamespace
 	featureGateWaitTimeout = 1 * time.Minute
 	// jobTimeout bounds the process's overall runtime safely below the Job manifest's
-	// activeDeadlineSeconds (300s), so it can log a clear error and exit cleanly instead of
+	// activeDeadlineSeconds (5m), so it can log a clear error and exit cleanly instead of
 	// being killed by the kubelet once the deadline is hit.
 	jobTimeout = 4 * time.Minute
 )
--- a/cmd/node-label-sync-job/main.go
+++ b/cmd/node-label-sync-job/main.go
@@ -42,6 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
@@ -125,7 +126,17 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := controllers.SyncVSphereNodeLabels(ctx, k8sClient, featureGateAccessor); err != nil {
+	currentFeatureGates, err := featureGateAccessor.CurrentFeatureGates()
+	if err != nil {
+		setupLog.Error(err, "failed to get current feature gates")
+		os.Exit(1)
+	}
+	if !currentFeatureGates.Enabled(features.FeatureGateVSphereMixedNodeEnv) {
+		setupLog.Info("VSphereMixedNodeEnv feature gate is disabled, nothing to do")
+		return
+	}
+
+	if err := controllers.SyncVSphereNodeLabels(ctx, k8sClient); err != nil {
 		setupLog.Error(err, "failed to sync vSphere node labels")
 		os.Exit(1)
 	}
--- a/cmd/vsphere-node-label-sync-job/main.go
+++ b/cmd/vsphere-node-label-sync-job/main.go
@@ -1,0 +1,134 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Command vsphere-node-label-sync-job runs to completion once, backfilling the
+// node.openshift.io/platform-type=vsphere label onto vSphere nodes that are missing it. It is
+// installed as a Kubernetes Job by the CVO (see
+// manifests/0000_26_cloud-controller-manager-operator_45_job-vsphere-node-label-sync.yaml) rather
+// than run as an ongoing in-process controller.
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"os"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	// to ensure that exec-entrypoint and run can make use of them.
+	"k8s.io/client-go/kubernetes"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	klog "k8s.io/klog/v2"
+	"k8s.io/utils/clock"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	configv1 "github.com/openshift/api/config/v1"
+	configv1client "github.com/openshift/client-go/config/clientset/versioned"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"github.com/openshift/library-go/pkg/operator/events"
+
+	"github.com/openshift/cluster-cloud-controller-manager-operator/pkg/controllers"
+)
+
+var (
+	scheme   = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+const (
+	recorderName           = "cloud-controller-manager-operator-vsphere-node-label-sync-job"
+	missingVersion         = "0.0.1-snapshot"
+	managedNamespace       = controllers.DefaultManagedNamespace
+	featureGateWaitTimeout = 1 * time.Minute
+	// jobTimeout bounds the process's overall runtime safely below the Job manifest's
+	// activeDeadlineSeconds (300s), so it can log a clear error and exit cleanly instead of
+	// being killed by the kubelet once the deadline is hit.
+	jobTimeout = 4 * time.Minute
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(configv1.AddToScheme(scheme))
+}
+
+func main() {
+	klog.InitFlags(flag.CommandLine)
+	flag.Parse()
+
+	ctrl.SetLogger(klog.NewKlogr().WithName("VSphereNodeLabelSyncJob"))
+
+	ctx, cancel := context.WithTimeout(ctrl.SetupSignalHandler(), jobTimeout)
+	defer cancel()
+
+	restConfig := ctrl.GetConfigOrDie()
+
+	k8sClient, err := client.New(restConfig, client.Options{Scheme: scheme})
+	if err != nil {
+		setupLog.Error(err, "unable to create client")
+		os.Exit(1)
+	}
+
+	configClient, err := configv1client.NewForConfig(restConfig)
+	if err != nil {
+		setupLog.Error(err, "unable to create config client")
+		os.Exit(1)
+	}
+
+	kubeClient, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		setupLog.Error(err, "unable to create kube client")
+		os.Exit(1)
+	}
+
+	configInformers := configinformers.NewSharedInformerFactory(configClient, 10*time.Minute)
+	controllerRef, err := events.GetControllerReferenceForCurrentPod(ctx, kubeClient, managedNamespace, nil)
+	if err != nil {
+		klog.Warningf("unable to get owner reference (falling back to namespace): %v", err)
+	}
+
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		controllers.GetReleaseVersion(), missingVersion,
+		configInformers.Config().V1().ClusterVersions(), configInformers.Config().V1().FeatureGates(),
+		events.NewKubeRecorder(kubeClient.CoreV1().Events(managedNamespace), recorderName, controllerRef, clock.RealClock{}),
+	)
+	featureGateAccessor.SetChangeHandler(func(featuregates.FeatureChange) {})
+	go featureGateAccessor.Run(ctx)
+	go configInformers.Start(ctx.Done())
+
+	select {
+	case <-featureGateAccessor.InitialFeatureGatesObserved():
+	case <-ctx.Done():
+		setupLog.Error(ctx.Err(), "context canceled while waiting for FeatureGate detection")
+		os.Exit(1)
+	case <-time.After(featureGateWaitTimeout):
+		setupLog.Error(errors.New("timed out waiting for FeatureGate detection"), "unable to determine feature gate state")
+		os.Exit(1)
+	}
+
+	if err := controllers.SyncVSphereNodeLabels(ctx, k8sClient, featureGateAccessor); err != nil {
+		setupLog.Error(err, "failed to sync vSphere node labels")
+		os.Exit(1)
+	}
+
+	setupLog.Info("vSphere node label sync completed successfully")
+}
--- a/manifests/0000_26_cloud-controller-manager-operator_45_job-vsphere-node-label-sync.yaml
+++ b/manifests/0000_26_cloud-controller-manager-operator_45_job-vsphere-node-label-sync.yaml
@@ -1,0 +1,118 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: vsphere-node-label-sync
+  namespace: openshift-cloud-controller-manager-operator
+  annotations:
+    capability.openshift.io/name: CloudControllerManager
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+    # Only create this Job on clusters where the VSphereMixedNodeEnv feature gate is enabled,
+    # matching the same gate the vSphere CCM checks before applying the node.openshift.io/platform-type
+    # label at node-init time (see pkg/cloud/vsphere/vsphere.go). The CVO re-evaluates this against the
+    # live FeatureGate object on every sync, so the Job is created as soon as the gate turns on.
+    release.openshift.io/feature-gate: "VSphereMixedNodeEnv"
+    # This Job must only ever be created once, never updated. Its pod template's container image
+    # changes on nearly every release, but a Job's pod template is immutable once created, so if
+    # the CVO ever tried to reconcile this Job in place (instead of only creating it) the update
+    # would fail on every subsequent upgrade after the first. create-only makes the CVO create it
+    # once and leave it alone from then on, matching the intended "backfill once" semantics.
+    release.openshift.io/create-only: "true"
+  labels:
+    k8s-app: vsphere-node-label-sync
+spec:
+  backoffLimit: 6
+  activeDeadlineSeconds: 300
+  template:
+    metadata:
+      labels:
+        k8s-app: vsphere-node-label-sync
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: hostaccess
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: cluster-cloud-controller-manager
+      restartPolicy: OnFailure
+      # This Job may run before pod networking (CNI/OVN service routing) is functional on a given
+      # node -- e.g. right after a control-plane node reboots during an upgrade. hostNetwork plus
+      # sourcing /etc/kubernetes/apiserver-url.env (mirroring the operator Deployment in
+      # 0000_26_cloud-controller-manager-operator_50_deployment.yaml) lets the container reach the
+      # API server directly instead of through the "kubernetes" Service ClusterIP, which requires
+      # kube-proxy/OVN to already be wired up. That file is only populated on control-plane nodes,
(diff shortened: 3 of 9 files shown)


manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,102 @@
...
+      serviceAccountName: cluster-cloud-controller-manager
+      restartPolicy: OnFailure
+      # This Job may run before pod networking (CNI/OVN service routing) is functional on a given
+      # node -- e.g. right after a control-plane node reboots during an upgrade. hostNetwork plus
+      # sourcing /etc/kubernetes/apiserver-url.env (mirroring the operator Deployment in
+      # 0000_26_cloud-controller-manager-operator_50_deployment.yaml) lets the container reach the
+      # API server directly instead of through the "kubernetes" Service ClusterIP, which requires
+      # kube-proxy/OVN to already be wired up. That file is only populated on control-plane nodes,
+      # hence the master nodeSelector/tolerations.
+      hostNetwork: true
```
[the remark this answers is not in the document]
vr4manta: Hi coderabbit, i asked Claude to make these changes and it responded with the following:
```
  Reasoning for skipping:

  This is the exact setting added to resolve the production failure Neil reported earlier in this conversation: oc logs showed dial tcp 
  172.30.0.1:443: i/o timeout when the Job pod tried to reach the "kubernetes" Service ClusterIP right after a control-plane node reboot during an
  upgrade — because pod-network service routing (kube-proxy/OVN) wasn't wired up on that node yet. The fix relies on hostNetwork: true together
  with sourcing /etc/kubernetes/apiserver-url.env (line 79-80) to reach the local control-plane apiserver directly, bypassing the Service ClusterIP
  entirely — mirroring the operator's own Deployment (0000_26_cloud-controller-manager-operator_50_deployment.yaml), which uses the identical
  pattern for the identical reason.

  There's also a technical reason removing hostNetwork would break this even independent of the ClusterIP-timing issue: apiserver-url.env points at
  a host-local address reachable only from the host's network namespace. Without hostNetwork: true, the pod has its own network namespace, so that
  address wouldn't even be reachable — "ordinary Kubernetes Service connectivity" is precisely the mode that failed in practice.

  I'm leaving the manifest unchanged.
```

What are your thoughts?

manifests/0000_26_cloud-controller-manager-operator_45_job-vsphere-node-label-sync.yaml
```diff
@@ -0,0 +1,102 @@
...
+      # This Job may run before pod networking (CNI/OVN service routing) is functional on a given
+      # node -- e.g. right after a control-plane node reboots during an upgrade. hostNetwork plus
+      # sourcing /etc/kubernetes/apiserver-url.env (mirroring the operator Deployment in
+      # 0000_26_cloud-controller-manager-operator_50_deployment.yaml) lets the container reach the
+      # API server directly instead of through the "kubernetes" Service ClusterIP, which requires
+      # kube-proxy/OVN to already be wired up. That file is only populated on control-plane nodes,
+      # hence the master nodeSelector/tolerations.
+      hostNetwork: true
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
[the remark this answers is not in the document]
vr4manta: Hi coderabbit, i asked Claude to make these changes and it responded with the following:
```
  Reasoning for skipping:

  1. The HyperShift premise doesn't apply here. CVO only includes a manifest for a given cluster profile (self-managed-HA, single-node, HyperShift,
  IBM-cloud-managed) if it carries that profile's include.release.openshift.io/<profile> annotation. This Job — like every manifest in this repo,
  including the operator's own Deployment it mirrors — only declares self-managed-high-availability and single-node-developer. None of this repo's
  payload is ever applied to a HyperShift guest cluster, regardless of nodeSelector content, so there's no HyperShift compatibility gap to close.
  2. Removing it would break the fix from two reviews ago. The node-role.kubernetes.io/master selector isn't a stray "legacy" label choice — it's
  required by the hostNetwork + /etc/kubernetes/apiserver-url.env fix (lines 37-44) that resolved Neil's actual production failure. That file is
  only populated on control-plane nodes. Without the selector, the pod could land on a worker node where the file doesn't exist, the wrapper script
  would fall through to the (non-functional, per earlier verification) URL_ONLY_KUBECONFIG fallback, and the pod would be back to relying on
  ordinary Service networking — reintroducing the exact dial tcp 172.30.0.1:443: i/o timeout failure this design was built to avoid. This exactly
  mirrors why the operator's own Deployment uses the identical node-role.kubernetes.io/master selector for the identical reason.
```

What are your thoughts?

pkg/controllers/vsphere_node_label_sync.go
```diff
@@ -0,0 +1,91 @@
...
+	if node.Labels == nil {
+		node.Labels = map[string]string{}
+	}
+	node.Labels[vsphere.NodePlatformTypeLabelKey] = vsphere.NodePlatformTypeLabelValueVSphere
+
+	if err := c.Patch(ctx, node, patch); err != nil {
+		return fmt.Errorf("failed to patch label: %w", err)
+	}
+
+	klog.Infof("added %s=%s label to node %s", vsphere.NodePlatformTypeLabelKey, vsphere.NodePlatformTypeLabelValueVSphere, node.Name)
```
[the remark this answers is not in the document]
vr4manta: Hi coderabbit, i asked Claude to make these changes and it responded with the following:
```
● Verified against current code — line 89 matches exactly: klog.Infof("added %s=%s label to node %s", ..., node.Name).

  Skipping this finding — not valid. Kubernetes Node names aren't sensitive/PII in this ecosystem: they're readable via kubectl get nodes by any
  principal with basic Node RBAC, and logged at Info level constantly by core Kubernetes/OpenShift components (kubelet, the node lifecycle
  controller, machine-api-operator, cluster-autoscaler, etc.) as completely standard practice — there's no compliance convention here treating them
  as requiring redaction.

  This specific line is also a direct carry-over from the original VSphereNodeLabelSyncer controller (klog.Infof("VSphereNodeLabelSyncer: added 
  %s=%s label to node %s", ...)), not something newly introduced by the Job refactor — it was already reviewed once. For a one-time migration
  backfill, logging exactly which node got relabeled is genuinely useful audit/debugging information (confirming which legacy vSphere nodes were
  affected by the pre-feature-gate gap); collapsing it to a bare count would lose that without addressing any real exposure.
```

What are your thoughts?

vr4manta:
I noticed the azure job had the following in it:
```
event happened 22 times, something is wrong: namespace/openshift-cloud-controller-manager-operator node/ci-op-f3bpi1wx-1a9f6-wqlfv-master-0 pod/vsphere-node-label-sync-mr9f4 hmsg/dbcbf706fd - reason/Pulled Container image "registry.build13.ci.openshift.org/ci-op-f3bpi1wx/stable@sha256:70533cbc11986458b5137dd40ffdec8370b4ccb8c037a3292ff98ddba7cc01f1" already present on machine and can be accessed by the pod (18:14:45Z) result=reject 
event happened 23 times, something is wrong: namespace/openshift-cloud-controller-manager-operator node/ci-op-f3bpi1wx-1a9f6-wqlfv-master-0 pod/vsphere-node-label-sync-mr9f4 hmsg/1f4f0000c2 - reason/Failed Error: container has runAsNonRoot and image will run as root (pod: "vsphere-node-label-sync-mr9f4_openshift-cloud-controller-manager-operator(55ee35eb-5b85-4e15-aa0f-7b9a829056f2)", container: vsphere-node-label-sync) (18:14:45Z) result=reject }
```
I am thinking the change coderabbit suggested for changing the run definition of the job's container may be causing this.  Need to check vsphere log and retest.

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,102 @@
...
+          value: "0.0.1-snapshot"
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+        terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - mountPath: /etc/kubernetes
+          name: host-etc-kube
+          readOnly: true
```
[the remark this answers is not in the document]
vr4manta: Hi coderabbitai, When i made this change, I am seeing the following in some of the presubmits:
```
event happened 22 times, something is wrong: namespace/openshift-cloud-controller-manager-operator node/ci-op-f3bpi1wx-1a9f6-wqlfv-master-0 pod/vsphere-node-label-sync-mr9f4 hmsg/dbcbf706fd - reason/Pulled Container image "registry.build13.ci.openshift.org/ci-op-f3bpi1wx/stable@sha256:70533cbc11986458b5137dd40ffdec8370b4ccb8c037a3292ff98ddba7cc01f1" already present on machine and can be accessed by the pod (18:14:45Z) result=reject 
event happened 23 times, something is wrong: namespace/openshift-cloud-controller-manager-operator node/ci-op-f3bpi1wx-1a9f6-wqlfv-master-0 pod/vsphere-node-label-sync-mr9f4 hmsg/1f4f0000c2 - reason/Failed Error: container has runAsNonRoot and image will run as root (pod: "vsphere-node-label-sync-mr9f4_openshift-cloud-controller-manager-operator(55ee35eb-5b85-4e15-aa0f-7b9a829056f2)", container: vsphere-node-label-sync) (18:14:45Z) result=reject }
```
Here is the job link: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-cloud-controller-manager-operator/497/pull-ci-openshift-cluster-cloud-controller-manager-operator-main-e2e-azure-ovn-upgrade/2082509487373029376

I believe this suggestion caused it.  can you check to confirm.

vr4manta:
/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-vsphere-ovn-upgrade

vr4manta:
Installs are now working; however, both ovn test runs hit the following: 
```
[sig-arch] Managed cluster should set requests but not limits [Early] [Suite:openshift/conformance/parallel] expand_less	2s
{  fail [github.com/openshift/origin/test/extended/operators/resources.go:247]: Pods in platform namespaces are not following resource request/limit rules or do not have an exception granted:
  batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync defines a limit on cpu of 100m which is not allowed (rule: "batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync/limit[cpu]")
  batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync defines a limit on memory of 100Mi which is not allowed (rule: "batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync/limit[memory]")}
```
This was recommended by coderabbitai.  Need to see what we should do for this.

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,102 @@
...
+          fi
+          exec /vsphere-node-label-sync-job
+        env:
+        - name: RELEASE_VERSION
+          value: "0.0.1-snapshot"
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+        terminationMessagePolicy: FallbackToLogsOnError
```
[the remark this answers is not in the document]
vr4manta: Hi coderabbitai, it seems setting the limits on this job is causing an e2e failure.  Here is the error:
```
: [sig-arch] Managed cluster should set requests but not limits [Early] [Suite:openshift/conformance/parallel] expand_less	2s
{  fail [github.com/openshift/origin/test/extended/operators/resources.go:247]: Pods in platform namespaces are not following resource request/limit rules or do not have an exception granted:
  batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync defines a limit on cpu of 100m which is not allowed (rule: "batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync/limit[cpu]")
  batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync defines a limit on memory of 100Mi which is not allowed (rule: "batch/v1/Job/openshift-cloud-controller-manager-operator/<batch_job>/container/node-label-sync/limit[memory]")}
```
You can see it in the following two tests:
- https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-cloud-controller-manager-operator/497/pull-ci-openshift-cluster-cloud-controller-manager-operator-main-e2e-aws-ovn/2083277347062026240
- https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-cloud-controller-manager-operator/497/pull-ci-openshift-cluster-cloud-controller-manager-operator-main-e2e-vsphere-ovn/2083277438430744576

vr4manta:
/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-vsphere-ovn-upgrade

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,93 @@
...
+      serviceAccountName: cluster-cloud-controller-manager
+      restartPolicy: OnFailure
+      # This Job may run before pod networking (CNI/OVN service routing) is functional on a given
+      # node -- e.g. right after a control-plane node reboots during an upgrade. hostNetwork plus
+      # sourcing /etc/kubernetes/apiserver-url.env (mirroring the operator Deployment in
+      # 0000_26_cloud-controller-manager-operator_50_deployment.yaml) lets the container reach the
+      # API server directly instead of through the "kubernetes" Service ClusterIP, which requires
+      # kube-proxy/OVN to already be wired up. That file is only populated on control-plane nodes,
+      # hence the master nodeSelector/tolerations.
+      hostNetwork: true
```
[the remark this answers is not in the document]
vr4manta: ● This finding is already resolved in the current file. Lines 45–69 already contain the node-role.kubernetes.io/master: "" nodeSelector and the full 6-entry
  toleration set, matching the operator Deployment. This was the fix applied at the end of the previous session. No changes needed.

vr4manta:
All install and upgrade jobs have passed.  The payload-job test for upgrade is also passing which is what reported the initial bug.

cmd/vsphere-node-label-sync-job/main.go
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2021.
```
theobarberbany: 2026?

vr4manta: just copied from sync controllers. I can update to newer year.

vr4manta:
/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-vsphere-ovn-upgrade

pkg/controllers/vsphere_node_label_sync.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	currentFeatureGates, err := featureGateAccess.CurrentFeatureGates()
+	if err != nil {
+		return fmt.Errorf("failed to get current feature gates: %w", err)
+	}
+	if !currentFeatureGates.Enabled(features.FeatureGateVSphereMixedNodeEnv) {
+		klog.V(2).Infof("%s feature gate is disabled, skipping node label sync", features.FeatureGateVSphereMixedNodeEnv)
+		return nil
+	}
```
damdo: Optimization nit, we could move this above the infra get to avoid doing the get in the first place if the `featureGateAccess.CurrentFeatureGates()` (checking the feature gate first is cheaper then getting the infra (in-memory lookup))

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,115 @@
...
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      tolerations:
+      - key: node.cloudprovider.kubernetes.io/uninitialized
+        effect: NoSchedule
+        value: "true"
```
damdo: Is this a duplicate of the one at line 62?

cmd/node-label-sync-job/main.go
```diff
@@ -0,0 +1,134 @@
...
+	// jobTimeout bounds the process's overall runtime safely below the Job manifest's
+	// activeDeadlineSeconds (300s), so it can log a clear error and exit cleanly instead of
+	// being killed by the kubelet once the deadline is hit.
+	jobTimeout = 4 * time.Minute
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(configv1.AddToScheme(scheme))
+}
```
damdo, line 73: We are not registering these for the fake client in the tests

damdo: We could do this in a TestMain or at the top of the test, matching what the real main.go does in its init()

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,115 @@
...
+        tolerationSeconds: 120
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoExecute"
+        tolerationSeconds: 120
+      - key: "node.cloudprovider.kubernetes.io/uninitialized"
+        operator: "Exists"
+        effect: "NoSchedule"
+        # CNI relies on CCM to fill in IP information on Node objects.
+        # Therefore we must schedule before the CNI can mark the Node as ready.
```
damdo: Indentation of the comment is wrong here if it refers to the key below

cmd/node-label-sync-job/main.go
```diff
@@ -0,0 +1,134 @@
...
+		os.Exit(1)
+	}
+
+	if err := controllers.SyncVSphereNodeLabels(ctx, k8sClient, featureGateAccessor); err != nil {
+		setupLog.Error(err, "failed to sync vSphere node labels")
+		os.Exit(1)
+	}
+
+	setupLog.Info("vSphere node label sync completed successfully")
+}
```
damdo, line 145: The job creates three K8s clients and a full informer framework to check one feature gate before doing a straightforward list+patch and for a run-once Job that feels a bit heavy.

So I looked at whether we could replace the informer with a `Get()` on the `FeatureGate` resource + `NewHardcodedFeatureGateAccessFromFeatureGate`, but that won't work cleanly as `NewFeatureGateAccess` has version logic.

That said, you could still simplify things by moving the feature gate check into main.go.
Do the informer dance there, resolve the gate, then pass a plain bool (or just don't call SyncVSphereNodeLabels at all if the gate is off). That way the function signature drops from `(ctx, client, featureGateAccess)` to just `(ctx, client)`

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,115 @@
...
+      serviceAccountName: cluster-cloud-controller-manager
+      restartPolicy: OnFailure
+      # This Job may run before pod networking (CNI/OVN service routing) is functional on a given
+      # node -- e.g. right after a control-plane node reboots during an upgrade. hostNetwork plus
+      # sourcing /etc/kubernetes/apiserver-url.env (mirroring the operator Deployment in
+      # 0000_26_cloud-controller-manager-operator_50_deployment.yaml) lets the container reach the
+      # API server directly instead of through the "kubernetes" Service ClusterIP, which requires
+      # kube-proxy/OVN to already be wired up. That file is only populated on control-plane nodes,
+      # hence the master nodeSelector/tolerations.
+      hostNetwork: true
```
damdo, line 44: Could we make it run strictly only after CNI/OVN is up so we drop the hostNet?

pkg/controllers/vsphere_node_label_sync.go
vr4manta: Ok, I can make that change.

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
vr4manta: IMO no.  This is the same as what you'll find in the deployment yaml in this same manifest directory.  These were added based on recommendations from CodeRabbitAI above.

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
vr4manta: ok, i'll fix.

cmd/node-label-sync-job/main.go
vr4manta: ok, i'll look into this.

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
vr4manta: Well not that it is run much later in the deployment, I can look into adjust this.

vr4manta:
@damdo I have pushed changes addressing your feedback.  Please review when you have time.

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
vr4manta: Made some adjustments.  Hope this is better.

damdo (commented):
/approve

/assign @theobarberbany 

for LGTM

@vr4manta Feel free to add verified when all the manual and CI payload testing has been done you are happy with. Thanks!

vr4manta:
/payload-job periodic-ci-openshift-release-main-ci-5.0-e2e-vsphere-ovn-upgrade

vr4manta:
/verified by @vr4manta manual tests covering install and upgrade in addition to payload-job above and vsphere ci test.

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
```diff
@@ -0,0 +1,112 @@
...
+    # changes on nearly every release, but a Job's pod template is immutable once created, so if
+    # the CVO ever tried to reconcile this Job in place (instead of only creating it) the update
+    # would fail on every subsequent upgrade after the first. create-only makes the CVO create it
+    # once and leave it alone from then on, matching the intended "backfill once" semantics.
+    release.openshift.io/create-only: "true"
+  labels:
+    k8s-app: node-label-sync
+spec:
+  backoffLimit: 6
+  activeDeadlineSeconds: 300
```
theobarberbany, line 25: Do we want a [`spec.ttlSecondsAfterFinished`](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) to clean these up?

pkg/controllers/vsphere_node_label_sync.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return errors.Join(errs...)
+}
+
+func syncVSphereNodeLabel(ctx context.Context, c client.Client, node *corev1.Node) error {
+	if !strings.HasPrefix(node.Spec.ProviderID, vsphere.NodeProviderIDPrefix) {
+		return nil
+	}
+
+	if _, ok := node.Labels[vsphere.NodePlatformTypeLabelKey]; ok {
```
theobarberbany, line 64: If we have `vsphere.NodePlatformTypeLabelKey` set to an incorrect value, e.g `foo` - does that matter?

cmd/node-label-sync-job/main.go
```diff
@@ -0,0 +1,145 @@
...
+
+const (
+	recorderName           = "cloud-controller-manager-operator-node-label-sync-job"
+	missingVersion         = "0.0.1-snapshot"
+	managedNamespace       = controllers.DefaultManagedNamespace
+	featureGateWaitTimeout = 1 * time.Minute
+	// jobTimeout bounds the process's overall runtime safely below the Job manifest's
+	// activeDeadlineSeconds (300s), so it can log a clear error and exit cleanly instead of
+	// being killed by the kubelet once the deadline is hit.
+	jobTimeout = 4 * time.Minute
```
theobarberbany: nit: This confused me first read, 300s==5 mins, id stick to one unit, seconds or mins :)

pkg/controllers/vsphere_node_label_sync.go
vr4manta: in the eyes of deployments, it would assume not a vsphere node.  I think in future we can leverage this for other platforms.  We also, here, do not want to override it if it already exists.  No need to make an unnecessary update if already vsphere.

cmd/node-label-sync-job/main.go
vr4manta: good catch.  That is weird.  I'll fix this.

theobarberbany:
/lgtm 

/hold

in case you want to address any nits

manifests/0000_90_cloud-controller-manager-operator_00_job.yaml
vr4manta: I think plan was to leave it around in case we needed to check it.  The job should still exist since we do not want each update in future to rerun this check since all CCMs pods should now be updating new nodes.

vr4manta:
/verified by @vr4manta manual tests covering install and upgrade in addition to payload-job above and vsphere ci test.

PR #502, opened 2026-08-12
CORS-4551: GCP: mount bound sa token volume

patrickdillon:
On GCP, we should mount the bound sa token volume, similarly to how it is done on Azure. Typically the volume is mounted by the CCO webhook, but that pattern is problematic if the CCM needs the projected credentials to schedule the CCO to run it's webhook, in which case we hit a deadlock. We can avoid this entirely by simply always mounting.

This doesn't occur in most installs because the CCM is authenticated by a VM attached to the service account, but GCD uses a projected credential and we want to move that direction for all installs because you can dynamically upgrade permissions.

## Summary by CodeRabbit

* **New Features**
  * Added an OpenShift-audience service-account token to the cloud controller manager deployment.
  * The token is securely mounted for use at `/var/run/secrets/openshift/serviceaccount`.

### DIFF
--- a/pkg/cloud/gcp/assets/cloud-controller-manager.yaml
+++ b/pkg/cloud/gcp/assets/cloud-controller-manager.yaml
@@ -117,6 +117,9 @@ spec:
             - name: trusted-ca
               mountPath: /etc/pki/ca-trust/extracted/pem
               readOnly: true
+            - name: bound-sa-token
+              mountPath: /var/run/secrets/openshift/serviceaccount
+              readOnly: true
       volumes:
         - name: config-gccm
           configMap:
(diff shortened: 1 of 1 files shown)


tthvo:
/payload-job periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-upgrade

patrickdillon:
Tested with https://github.com/openshift/cloud-credential-operator/pull/1073 

/verified by @padillon (see comment in other PR)

e2es show no regressions

damdo (commented):
/approve

/assign @theobarberbany @nrb

